PR #1, opened 2022-02-04
Strip reader

sameeul:
Added strip Tiff file reading capability. 
Now, ImageLoader class will check whether a file is tiled or not and based on that instantiate either GrayScaleTileTiffLoader or GrayScaleStripTiffLoader.

### DIFF
--- a/src/nyx/image_loader.cpp
+++ b/src/nyx/image_loader.cpp
@@ -15,15 +15,9 @@ bool ImageLoader::open(const std::string& int_fpath, const std::string& seg_fpat
 		} 
 		else 
 		{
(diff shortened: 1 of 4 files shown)


friskluft:
I suggest changing "uint32 row, layer" in C:\WORK\AXLE\FastLoader\FastLoader-main__Github\FastLoader-main\fast_loader\specialised_tile_loader\grayscale_tiff_strip_loader.h / loadTileFromFile() to "uint32_t row, layer" or "unsigned int row, layer", otherwise this code won't build with Visual C++.

friskluft:
My tests pass OK with this pr.

PR #4, opened 2022-02-08
Improved CLI and Python library output

friskluft:


hsidky:
@friskluft Any reason we need all the `#ifdef WITH_PYTHON_H` conditionals? Why not just throw an error and catch it in main or somewhere and print out to cerr? Then the python code will operate naturally and you don't need all the redundant code.

friskluft:
Top reasons:
- preferring to keep the stable and tested workflow logic based on boolean success/error returned results 
- following a rule to avoid cross-module signaling by means of exceptions

PR #16, opened 2022-03-23
update build script

sameeul:


hsidky:
@sameeul Let's unit test this and I'll merge.

PR #35, opened 2022-07-01
Unit tests - intensity features

friskluft:


tests/CMakeLists.txt
```diff
@@ -6,12 +6,80 @@ set(CMAKE_EXE_LINKER_FLAGS "-v")
...
+	../src/nyx/phase3.cpp
+	../src/nyx/reduce_by_feature.cpp
+	../src/nyx/reduce_trivial_rois.cpp
+	../src/nyx/roi_cache.cpp
+	../src/nyx/roi_cache_basic.cpp
+	../src/nyx/scan_fastloader_way.cpp
+	../src/nyx/pixel_feed.cpp
+)
 
 add_executable(runAllTests ${TEST_SRC})
```
sameeul, line 76: In some platforms (CentOS, RedHat etc.)  the build fails due to not finding std::filesystem. We should link against `stdc++fs` to avoid this. I would also propose to only add Tiff libraries and header under ` if(TIFF_FOUND) ` logic and keep the initial lines 15 -17 separate. Something like the following:

`target_include_directories (runAllTests PUBLIC ${GTEST_INCLUDE_DIRS})
target_link_directories (runAllTests PUBLIC ${GTEST_LIBRARY_PATH})
target_link_libraries (runAllTests gtest ${runAllTests_LIBRARIES} stdc++fs)

find_package(TIFF REQUIRED)
if (TIFF_FOUND)
	target_include_directories (runAllTests PUBLIC ${TIFF_CXX_FLAGS})
	target_link_libraries (runAllTests ${TIFF_LIBRARIES} stdc++fs)
else ()
	message(FATAL_ERROR "libtiff library has not been found, please install it.")
	SET(FastLoader_FOUND OFF)
endif (TIFF_FOUND)`
`

sameeul: Here is an example CMakeLists.txt that I tested: https://github.com/sameeul/nyxus/blob/pr_35_merge/tests/CMakeLists.txt

src/nyx/features/image_matrix.cpp
```diff
@@ -120,6 +102,68 @@ void ImageMatrix::print (const std::string& head, const std::string& tail, std::
 	std::cout << tail;
 }
 
+void ImageMatrix::print (std::ofstream& f, const std::string& head, const std::string& tail, std::vector<PrintablePoint> special_points)
```
sameeul, line 105: Should we pass the vector as a reference?

src/nyx/features/pixel.h
```diff
@@ -52,7 +52,8 @@ struct Pixel2 : public Point2i
 	Pixel2 (StatsInt x_, StatsInt y_, PixIntens i_) : Point2(x_, y_), inten(i_) {}
 	Pixel2 (int x_, int y_, PixIntens i_) : Point2(x_, y_), inten(i_) {}
 	Pixel2 (float x_, float y_, PixIntens i_) : Point2((StatsInt)x_, (StatsInt)y_), inten(i_) {}
-	Pixel2 (size_t x_, size_t y_, double i_) : Point2(x_, y_), inten(i_) {}
+	Pixel2 (size_t x_, size_t y_, double i_) : Point2((StatsInt)x_, (StatsInt)y_), inten((PixIntens)i_) {}
```
sameeul, line 55: Guessing the loss of precision from `i_` to `inten` is intentional.

friskluft: Yes, coordinates are 'unsigned int' anyway.

src/nyx/features/image_matrix.cpp
friskluft: Here - no, otherwise it'd ruin the initializer list based- default parameter logic:
void print(const std::string& head = "", const std::string& tail = "", std::vector<PrintablePoint> special_points = {});
(See image_matrix.h)

tests/CMakeLists.txt
friskluft: Thanks. Updating my CMakeLists.txt

sameeul (approved):

JesseMckinzie (approved):

PR #36, opened 2022-07-01
GPU enabled Gabor features

JesseMckinzie:
- Gabor features have the option to be computed on the GPU on a machine with a CUDA enabled GPU
- C++ and Python unit tests to compare the results of the CPU Gabor features to the GPU results
- CMake is updated to find if the machine has a CUDA enabled GPU and find the compute capability of the device
- GPU functions added to the environment to select whether to use the gpu, get the gpu properties, and check if a gpu is available 
- Functions added to Python interface to check for GPU, get GPU properties, and set whether to use GPU. Constructor now takes in device number to choose the GPU


src/nyx/features/gabor.cpp
```diff
@@ -552,7 +913,18 @@ void GaborFeature::reduce (size_t start, size_t end, std::vector<int>* ptrLabels
         LR& r = (*ptrLabelData)[lab];
 
         GaborFeature gf;
-        gf.calculate (r);
+
+
+        #ifdef USE_GPU
```
friskluft, line 918: This method of selective calculating a feature on CPU- or GPGPU-side implemented in Gaborfeature::reduce() that I originally suggested you has become old. There is a better one. The problem with the old method is that if Gaborfeature::reduce() is called from multiple CPU-side threads (command line argument --reduceThreads=n, n>1), function GaborFeature::GaborEnergyGPU() will be hitting the GPU resource in parallel creating shortage of thread blocks. The better method to call Gabor feature calculation mentioned is to put the #ifdef USE_GPU and theEnvironment.using_gpu() logic in the reduction function and call the GPU code in the main thread keeping the CPU code callable in multiple threads. You can check file https://github.com/PolusAI/nyxus/blob/main/src/nyx/reduce_trivial_rois.cpp, lines 220-241, for an example of implementing this logic for 2D moments. Feel free to message me for more details.

CMakeLists.txt
```diff
@@ -1,22 +1,48 @@
...
+	SET(CMAKE_CUDA_ARCHITECTURES ${CUDA_ARCH_LIST})
+	set_property(GLOBAL PROPERTY CUDA_ARCHITECTURES "${CUDA_ARCH_LIST}")
+	message("Found GPU with compute capability ${CUDA_ARCH_LIST}.")
 endif()
 
 #==== Compiler Options
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
-set(CMAKE_CUDA_STANDARD 17)
-set(CMAKE_CUDA_STANDARD_REQUIRED True)
```
sameeul, line 19: Are we dropping these lines intentionally?

CMakeLists.txt
```diff
@@ -159,11 +175,25 @@ pybind11_add_module(backend
...
+		${CUDA_TOOLKIT_INCLUDE})
+	target_link_libraries(backend PRIVATE ${CUDA_LIBRARIES} ${CUDA_CUFFT_LIBRARIES})
+	set_target_properties(backend PROPERTIES CUDA_ARCHITECTURES "35;37;50;72")
+
 	target_link_libraries(backend PRIVATE nyxus_gpu)
-       	set_target_properties(backend PROPERTIES CUDA_ARCHITECTURES "35;37;50;72")
 	if(BUILD_CLI)
 		target_link_libraries(nyxus PRIVATE nyxus_gpu)
-        	set_target_properties(nyxus PROPERTIES CUDA_ARCHITECTURES "35;37;50;72")
+		#target_link_libraries(nyxus PRIVATE nyxus_gpu)
```
sameeul, line 194: are lines 193 and 194 same? May be we can remove the commented one?

src/nyx/environment.cpp
```diff
@@ -1003,6 +1032,32 @@ bool Environment::using_gpu()
...
+
+    for(int i = 0; i < n_devices; ++i){
+        cudaDeviceProp prop;
+        cudaGetDeviceProperties(&prop, i);
+
+        std::map<std::string, std::string> temp;
+
+        temp["Device number"] =  std::to_string(i);
+        temp["Device name"] = prop.name;
+        temp["Memory"] = std::to_string(prop.totalGlobalMem/pow(10,9)) + " GB";
```
sameeul, line 1050: Just curious if we should we pow(2,xxx) to get the correct number?

src/nyx/features/gabor.cpp
```diff
@@ -89,6 +91,216 @@ void GaborFeature::calculate (LR& r)
...
+        writeablePixels e2_pix_plane = e2img.WriteablePixels();
+        GaborEnergyGPU (Im0, e2_pix_plane.data(), auxC.data(), auxG.data(), f0[ii], sig2lam, gamma, theta, n);
+
+        //
+        //Moments2 local_stats2;
+        //e2img.GetStats(local_stats2);
+        //double max_val2 = local_stats2.max__();
+        //
+        //e2_pix_plane.array() = (e2_pix_plane.array() / max_val2).unaryExpr(Moments2func(e2img.stats));
+        //
```
sameeul, line 168: Same comments as Andiy. IMO, removing dead and commented out codes increase readability. I am also guilty of that at times :-(

src/nyx/gpu/gabor.cu
```diff
@@ -0,0 +1,418 @@
...
+
+        // Wait for device to finish all operation
+        cudaDeviceSynchronize();
+
+        // Check if kernel execution generated and error
+        //getLastCudaError("Kernel execution failed [ solvePoisson ]");
+        cudaError_t err = cudaGetLastError();   
+        if ( err != cudaSuccess ){
+                //fprintf(stderr, "Kernel execution failed [ solvePoisson ]\n");
+                printf("CUDA Error: %s\n", cudaGetErrorString(err));
```
sameeul, line 39: I am not very familiar with CUDA best practices. So, just curious where `std::cout` can be used instead of `printf` here.

src/nyx/gpu/gabor.cu
```diff
@@ -0,0 +1,418 @@
...
+        int size = row_size * col_size;
+
+        if ((2 * size * batch_size) >= CUFFT_MAX_SIZE) {
+            throw invalid_argument("Batch of images is too large. The maximumum number of values in cuFFT is 2^27.");
+        }
+
+        // allocate space for linear indexed arrays
+        Complex* linear_image = (Complex*)malloc(size * batch_size * sizeof(Complex));
+        Complex* result = (Complex*)malloc(size * batch_size * sizeof(Complex));
+        Complex* linear_kernel = (Complex*)malloc(size * batch_size * sizeof(Complex));
```
sameeul, line 102: We should add check for allocation failure. Also, can STL containers (vector?)  be used here?

src/nyx/gpu/gabor.cu
```diff
@@ -0,0 +1,418 @@
...
```
sameeul, line 263: Needs check for allocation failure.

src/nyx/gpu/gabor.cu
```diff
@@ -0,0 +1,418 @@
...
+        cmat_mult(d_image, row_size, col_size, d_kernel, d_result, batch_size);
+
+        // transform out of fourier space
+        if (cufftExecZ2Z(plan, d_result, d_result, CUFFT_INVERSE) != CUFFT_SUCCESS){
+            fprintf(stderr, "CUFFT Error: Unable to execute plan\n");
+            //return;		
+        }
+
+        // copy results from device to host
+        cudaMemcpy(result, d_result, batch_size*size*sizeof(CuComplex), cudaMemcpyDeviceToHost);
```
sameeul, line 391: Some of the cudaMemcpy calls have error checking after the call but some are missing. I am not sure what is the performance cost of these checks but I am in favor of having them.

tests/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+cmake_minimum_required(VERSION 3.12)
```
sameeul, line 1: This file needs to be rebased and merged

tests/test_gabor.cc
```diff
@@ -0,0 +1,67 @@
...
+    while(true) {
+        getline(gpu_in, gpu);
+        getline(cpu_in, cpu);
+        
+        if(gpu == "" || cpu == "") break;
+
+        if(gpu != cpu){
+            cout << "gpu: " << gpu << endl;
+            cout << "cpu: " << cpu << endl;
+            continue;
```
sameeul, line 27: Should we compare with some sort of tolerance?

sameeul (commented):
I am not very familiar with the internals of the algorithm. So I only tried to look at the coding style and made some comments on that. This is a really large PR (great job!)  and hopefully my comments are helpful :-)

CMakeLists.txt
JesseMckinzie: Those lines were duplicates. The same lines appear on lines 16-17 and 18-19 in the current CMake so I removed the duplicate. Let me know if you want them added back.

src/nyx/gpu/gabor.cu
JesseMckinzie: std::cout cannot be used in kernel functions. std is general not available for use in CUDA functions

src/nyx/features/gabor.cpp
```diff
@@ -543,6 +755,155 @@ void GaborFeature::GaborEnergy (
...
+    double gamma, 
+    double theta, 
+    int n,
+    int num_filters) 
+{
+    int n_gab = n;
+
+    //cout << "g filters" << endl;
+    double* g_filters;
+    g_filters = (double*)malloc(2 * n * n * num_filters * sizeof(double));
```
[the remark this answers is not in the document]
JesseMckinzie: I changed g_filters to be an std::vector.

src/nyx/features/gabor.cpp
```diff
@@ -543,6 +755,155 @@ void GaborFeature::GaborEnergy (
...
+        f0 = f[i];
+        Gabor (Gexp, f0, sig2lam, gamma, theta, fi, n_gab);
+        for(int i = 0; i < 2*n*n; ++i) {
+            g_filters[idx+i] = Gexp[i];
+        }
+
+        idx += 2*n*n;
+    }
+
+    //for(int i = 0; i < 2 * n *n; ++i){
```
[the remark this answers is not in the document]
JesseMckinzie: Code has been cleaned up.

src/nyx/features/gabor.cpp
JesseMckinzie: Code has been cleaned up.

tests/test_gabor.cc
JesseMckinzie: These lines are no longer needed. I will remove them. If we change from double to float in GPU code, we will need to compare with a tolerance.

src/nyx/gpu/gabor.cu
JesseMckinzie: Changed pointers to std::vector

sameeul (approved):

PR #50, opened 2022-08-10
ImageMatrix Initialization Tests

JesseMckinzie:
Added tests to check if the ImageMatrix has been initialized for all features calculations.


tests/test_main_nyxus.h
```diff
@@ -0,0 +1,78 @@
...
+            roidata.raw_pixels.push_back(Pixel2(i%w, i/w, px));
+            ++i;
+        }
+
+        if (allocate_IM)
+            roidata.aux_image_matrix = ImageMatrix(roidata.raw_pixels);
+
+        //std::cout << "printing" << std::endl;
+        //roidata.aux_image_matrix.print();
+        //std::cout << "end printing" << std::endl;
```
sameeul, line 75: we can delete these also :-)

tests/test_initialization.h
```diff
@@ -0,0 +1,38 @@
...
+
+    for(int i = 0; i < pending.size(); ++i){
+        LR& r = roiData[i];
+
+        const ImageMatrix& im = r.aux_image_matrix;
+
+        ASSERT_TRUE(im.height > 0);
+        ASSERT_TRUE(im.width > 0);
+    }
+
```
sameeul, line 36: We will also probably need `freeTrivialRoisBuffers (pending);` just to be on the safe side of memory leak

tests/test_download_data.cc
```diff
@@ -4,11 +4,15 @@ using namespace std;
 
 int get(const string& url, const string& filename){
```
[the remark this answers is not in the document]
JesseMckinzie: Not anymore. I removed these files

tests/test_main_nyxus.h
JesseMckinzie: Removed

tests/test_main_nyxus.h
```diff
@@ -0,0 +1,78 @@
...
+#include <src/nyx/parallel.h>
+#include "test_dsb2018_data.h"
+#include "test_data.h"
+
+namespace Nyxus
+{
+    extern void init_label_record_2(LR& lr, const std::string& segFile, const std::string& intFile, int x, int y, int label, PixIntens intensity, unsigned int tile_index);
+    extern void update_label_record_2(LR& lr, int x, int y, int label, PixIntens intensity, unsigned int tile_index);
+    extern void allocateTrivialRoisBuffers(const std::vector<int>& Pending);
+
```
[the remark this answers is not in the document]
JesseMckinzie: Added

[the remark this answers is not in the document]
sameeul: Hi Jesse, what I meant was to add those two declarations to `globals.h` and then add `globals.h` as an include in this file.  The functions `init_label_record_2` and `update_label_record_2` are already declared in `globals.h`. Sorry for the lack of clarity in my previous comment. I think this prevents from extra declaration of these functions in the test code and use the declarations from the source.

tests/test_initialization.h
```diff
@@ -11,6 +11,9 @@
 #include "../src/nyx/environment.h"
 #include "../src/nyx/globals.h"
 
+using Nyxus::allocateTrivialRoisBuffers;
+using Nyxus::freeTrivialRoisBuffers;
```
sameeul, line 15: I don't think we need these. I removed them and compiled. The only think I had to add was `#include "../src/nyx/globals.h" ` in `test_main_nyxus.h`

sameeul (approved):

PR #60, opened 2022-10-11
Updated GLCM features (additional features matching NIST Feature2DJava, graytones to levels conversion ...

friskluft:
… like in F2DJ and Matlab IPT)


JesseMckinzie (commented):
Approved once checks are passed

PR #62, opened 2022-10-18
Added custom intensity-mask mapping in Py interface

friskluft:


src/nyx/python/new_bindings_py.cpp
```diff
@@ -118,6 +118,72 @@ py::tuple process_data(
...
+    if (intensFiles.size() != labelFiles.size())
+        throw std::runtime_error("Imbalanced intensity and segmentation file lists");
+    for (auto i = 0; i < intensFiles.size(); i++)
+    {
+        const std::string& i_fname = intensFiles[i];
+        const std::string& s_fname = labelFiles[i];
+
+        if (!directoryExists(i_fname))
+        {
+            std::cout << i_fname << "\n";
```
hsidky, line 149: Filename should be part of the run-time error string thrown or else the user won't see it in a jupyter notebook. Change text to "File does not exist: /path/to/file" 

Also, the function being called is `directoryExists`. If it works on both directories or files then it should be renamed to something more generic like `exists` or `existsOnFilesystem` or something else of your choosing.

src/nyx/python/new_bindings_py.cpp
```diff
@@ -118,6 +118,72 @@ py::tuple process_data(
...
+            throw std::runtime_error("Non-existing file");
+        }
+        if (!directoryExists(s_fname))
```
hsidky, line 152: Same comment as above.

src/nyx/python/nyxus/nyxus.py
```diff
@@ -153,6 +151,48 @@ def featurize(
...
     def using_gpu(self, gpu_on: bool):
         use_gpu(gpu_on)
+
+    def featurize (
+        self,
+        int_fnames: list,
+        seg_fnames: list):
+        """Extract features from image file pairs passed as lists
+
+        Extract stoy
```
hsidky, line 161: Copy the long description from `featurize_directory` here and modify accordingly.

hsidky:
@friskluft Thanks for addressing the comments. One final remark: Can we rename the arguments from `int_fnames` to `intensity_files` and same with masks? More Pythonic that way. Make sure to update the docstrings and documentation where necessary. Thanks!

PR #66, opened 2022-11-07
Added RLE fast pass to phase1/2

Nicholas-Schaub:
This PR implements a piece of code to quickly loop over structured (ROI label) data. While `pytest` does run, it should be considered WIP. Also, to run the `pytest` code, the pytest-benchmark plugin should be installed before running.

Most of the modifications are designed to operate with the use of a `use_fastloop` parameter, but function overloading was used so that no breaking changes were made.

The performance improvements on phase 1 come primarily from implementation of a fast run length encoding (RLE) algorithm that reduces the initial pass over the pixels by a factor of 5-10x. Overall, the reduction of processing time using the pytest benchmarks averages about 20%, but this is only for calculating all intensity features.

Since the RLE algorithm is fast, it might be beneficial to use RLE for calculating some of the features in Nyxus. For example, RLE effectively creates a pseudo-hull, drastically reducing the number of calculations for things like convex hull. This would be future work since some algorithms like convex hull could use RLE as a drop in replacement while other features might require retooling to take full advantage of RLE.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,7 +69,7 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
-set(CMAKE_CXX_FLAGS_RELEASE "-O2")
+set(CMAKE_CXX_FLAGS_RELEASE "-O2 -march=native")
 
 #== GTest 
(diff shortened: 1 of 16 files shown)


sameeul:
Closing as no plan/resource to incorporate this at the moment

Issue #69, opened 2022-11-12
Perimeter feature is incorrect


Nicholas-Schaub:
When getting area and perimeter features for the demo image `0acd2c223d300ea55d0546797713851e818e5c697d073b7f4091b96ce0f3d2fe.tif`, the following results are given.

```
                                          mask_image                                    intensity_image  label  AREA_PIXELS_COUNT  PERIMETER
0  0acd2c223d300ea55d0546797713851e818e5c697d073b...  0acd2c223d300ea55d0546797713851e818e5c697d073b...      1              134.0      169.0
1  0acd2c223d300ea55d0546797713851e818e5c697d073b...  0acd2c223d300ea55d0546797713851e818e5c697d073b...      2              266.0      324.0
2  0acd2c223d300ea55d0546797713851e818e5c697d073b...  0acd2c223d300ea55d0546797713851e818e5c697d073b...      3              192.0      272.0
3  0acd2c223d300ea55d0546797713851e818e5c697d073b...  0acd2c223d300ea55d0546797713851e818e5c697d073b...      4              256.0      330.0
4  0acd2c223d300ea55d0546797713851e818e5c697d073b...  0acd2c223d300ea55d0546797713851e818e5c697d073b...      5              279.0      360.0
5  0acd2c223d300ea55d0546797713851e818e5c697d073b...  0acd2c223d300ea55d0546797713851e818e5c697d073b...      6              392.0      506.0
6  0acd2c223d300ea55d0546797713851e818e5c697d073b...  0acd2c223d300ea55d0546797713851e818e5c697d073b...      7              267.0      324.0
7  0acd2c223d300ea55d0546797713851e818e5c697d073b...  0acd2c223d300ea55d0546797713851e818e5c697d073b...      8               21.0       25.0
8  0acd2c223d300ea55d0546797713851e818e5c697d073b...  0acd2c223d300ea55d0546797713851e818e5c697d073b...      9               95.0      117.0
```

At most, I would expect the perimeter to have an upper bound on area_pixel_count, but it is always higher.

It appears as though the error is caused by commented code here:
https://github.com/PolusAI/nyxus/blob/b066182b157616b3b295fdf0da7e33642e97cb8f/src/nyx/features/contour.cpp#L173

It looks like all pixels in the bounding box of the contour are being stored in the contour. I suspect this means the rest of the contour features are also incorrect.

Nicholas-Schaub:
[...]

PR #70, opened 2022-11-12
Use RLE to calculate morphological features

Nicholas-Schaub:
This PR is a proof of concept approach that demonstrates the performance improvements that may be gained by using RLE to calculate metrics. It is an extension of what was implemented in #66. In particular, modifications were made to the way the convex hull, basic morphology, and contour calculations were made. Each of these modifications improves the performance of Nyxus by reducing run time, and in the case of calculating the contour, also reducing memory overhead (since the aux image and padding doesn't need to be calculated).

As reported in issue #69, It appears as though the contour is incorrectly calculated. Assuming the calculation is correct, the performance improvements of `*ALL_MORPHOLOGIES*` improves by a factor of 3-4x overall when using RLE. If the assumed bug is fixed from #69, then the improvements are 40%-50% when using RLE. The reason why fixing the presumed bug makes the non-RLE version run more quickly is likely due to the number of neighbors calculation.

This PR is intended to only be a proof of concept, as with #66. It is possible there are bugs or additional considerations may need to be made for larger images and accounting for tile boundaries.


sameeul:
Closing as no plan/resource to incorporate this at the moment

PR #92, opened 2023-02-16
Handling zero-background, skipping ROIs, and improved timing output

friskluft:


src/nyx/python/nyxus/nyxus.py
```diff
@@ -195,7 +195,65 @@ def featurize (
             df["label"] = df.label.astype(np.uint32)
 
         return df
-		
+
+
+    def blacklist_roi(self, blacklist:str):
```
JesseMckinzie, line 200: I think the blacklist parameter could be a little confusing or error prone. Would it be possible to use a dict mapping the filename to a list of ROIs to blacklist or a similar approach that wouldn't require the user to learn a new notation? Let me know if you want to meet to chat about this

JesseMckinzie (commented):
I suggest adding unit tests for the blacklist functionality

src/nyx/python/nyxus/nyxus.py
friskluft: Top reasons for passing a blacklist as a string rather than something pythonic (a dictionary):
-- providing 100% compatibility with CLI. A CLI user is supposed to copy-paste a blacklist expression from --skiproi=<blacklist expression> and pass it to a Python API method and vice versa. 
-- giving a user a functionality at minimum development cost. ROI blacklisting is a kind of emergency functionality because an image having a mislabeled ROI is an exception, so I opted for reusing the CLI blacklisting code rather than spending time on creating a piece of versatile pythonic API serving an emergency use case, intranslatable to CLI directly.

src/nyx/roi_blacklist.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <string>
+#include <vector>
+
+//using pairType = std::pair<std::string, std::vector<int>>;
+class pairType : public std::pair<std::string, std::vector<int>> {};
```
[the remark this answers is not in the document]
friskluft: Reason: iterator dereferencing ("pairType& p = *foundFname;") needs a class, not an alias.

src/nyx/python/nyxus/nyxus.py
friskluft: A unit test is generally a good idea (when time permits).

JesseMckinzie: I see. I agree that its a good format for the CLI and that makes sense to keep it the same for exceptional functionality

JesseMckinzie (approved):

PR #95, opened 2023-03-09
Python in-memory API documentation

JesseMckinzie:
- Add documentation for `featurize` (in-memory api) to the ReadtheDocs
- Fix casting for 2D case in in-memory API
- Add unit test for 2D case of featurize

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ seg = [
 ]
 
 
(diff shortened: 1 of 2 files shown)


hsidky:
@JesseMckinzie Can you resolve the conflict here please?

PR #100, opened 2023-04-06
POL2943 gabor perf tuneup 3

friskluft:
Changing Gabor feature defaults via Python API

### DIFF
--- a/src/nyx/python/nyxus/nyxus.py
+++ b/src/nyx/python/nyxus/nyxus.py
@@ -45,6 +45,20 @@ class Nyxus:
         Id of the gpu to use. To find available gpus along with ids, using nyxus.get_gpu_properties().
         The default value of -1 uses cpu calculations. Note that the gpu features only support a single 
         thread for feature calculation. 
+    gabor_kersize: str (optional, default "16")
+        GABOR feature: dimension of the 2D Gabor filter kernel
+    gabor_gamma: str (optional, default "0.1")
+        GABOR feature: aspect ratio of Gabor filter's Gaussian factor
+    sig2lam: str (optional, default "0.8")
+        GABOR feature: spatial frequency bandwidth of Gabor filter
+    gabor_f0: str (optional, default "0.1")
+        GABOR feature: frequency of the baseline lowpass filter as denominator of `\pi`
+    gabor_theta: str (optional, default "45")
+        GABOR feature: orientation of the Gaussian in degrees 0-180
+    gabor_thold: str (optional, default "0.025")
+        GABOR feature: lower threshold of the filtered image to baseline ratio
+    gabor_freqs: str (optional, default "1,2,4,8,16,32,64")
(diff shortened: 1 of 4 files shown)


sameeul (approved):

src/nyx/python/nyxus/nyxus.py
```diff
@@ -347,6 +356,32 @@ def roi_blacklist_get_summary (self):
...
+        """Sets parameters of feature GABOR
+
+        Keyword args:
+        * kersize (int): size of filter kernel's side. Example: customize_gabor_feature(kersize=16)
+        * gamma (float): aspect ratio of the Gaussian factor. Example: customize_gabor_feature(gamma=0.1)
+        * sig2lam (float): spatial frequency bandwidth. Example: customize_gabor_feature(sig2lam=0.8)
+        * f0 (float): frequency of the baseline lowpass filter as denominator of `\pi`. Example: customize_gabor_feature(f0=0.1)
+        * theta (float): orientation of the Gaussian in degrees 0-180. Example: customize_gabor_feature(theta=1.5708)
+        * thold (float): lower threshold of the filtered image to baseline ratio. Example: customize_gabor_feature(thold=0.025)
+        * freqs (str): comma-separated denominators of `\pi` as frequencies of Gabor filter's harmonic factor. Example: customize_gabor_feature(freqs="1,2,4,8,16,32,64")
```
sameeul, line 369: According to the feedback from Hythem, this need to be a List of numbers.

src/nyx/python/nyxus/nyxus.py
```diff
@@ -58,8 +69,7 @@ def __init__(
         n_feature_calc_threads: int = 4,
         n_loader_threads: int = 1,
         using_gpu: int = -1,
-        ibsi: bool = False
-    ):
```
sameeul, line 62: my suggestion will be to add something like **gabor_kwargs here.

src/nyx/python/nyxus/nyxus.py
friskluft: I have a different view of utility of these string configurations, for example, an ability to add support of directional harmonics freqs="30:1,2,3;60:8,16;120:8,16;150:1,2,3" but OK, I'll rework this into a number list.

src/nyx/python/nyxus/nyxus.py
```diff
@@ -347,6 +356,32 @@ def roi_blacklist_get_summary (self):
             return None
         return s
 
+    def customize_gabor_feature (self, **kwargs):
```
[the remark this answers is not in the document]
friskluft: Customizing feature parameters via the class constructor or single point of initialization is (a) textually awkward when multiple features need to be preset in practical feature extraction cases and (b) incorrect when we need to fit some feature's parameters, for example, doing something like \arg\max_{\gamma,\lambda,\theta} \sum\frac{between}{within} for the Gabor feature - this way we'd need to construct a Nyxus instance in each iteration. 
I suggest to route this design via the product owner (Hythem) and I'll quietly implement the PO's design.

PR #109, opened 2023-05-03
POL2510 NGLDM features

friskluft:


src/nyx/features/ngldm.cpp
```diff
@@ -0,0 +1,346 @@
...
+		f_DCNUN =
+		f_GLCM =
+		f_GLV =
+		f_DCM =
+		f_DCV =
+		f_DCE =
+		f_DCENE = 0;
+}
+
+template <class PixelCloud> void NGLDMfeature::gather_unique_intensities (std::vector<PixIntens> & V, PixelCloud & C, PixIntens max_inten)
```
sameeul, line 50: should we use `const PixelCloud & C`?

src/nyx/features/ngldm.cpp
```diff
@@ -0,0 +1,346 @@
...
+		PixIntens inten_ = Nyxus::to_grayscale(p.inten, 0, range, nGrays, Environment::ibsi_compliance);
+		U.insert(inten_);
+	}
+
+	// -- Set to vector to be able to know each intensity's index
+	V.insert (V.end(), U.begin(), U.end()); //std::vector<PixIntens> V (U.begin(), U.end());
+	std::sort (V.begin(), V.end());
+}
+
+template <class Imgmatrix> void NGLDMfeature::calc_ngldm (SimpleMatrix<unsigned int> & NGLDM, Imgmatrix & I, std::vector<PixIntens> & V, PixIntens max_inten)
```
sameeul, line 66: should we add `const` qualifier to things that we take by reference but do not change in the function?

src/nyx/features/ngldm.cpp
```diff
@@ -0,0 +1,346 @@
...
+				if (!match)
+					continue;
+
+				unsigned int& binCount = NGLDM.yx(row, i);
+				binCount += (int)match;
+			}
+		}
+}
+
+void NGLDMfeature::calculate (LR& r)
```
sameeul, line 124: similar observation here. It is not clear to me whether `r` changes inside the function. If it does not change, I would prefer a `const`.

src/nyx/features/ngldm.cpp
```diff
@@ -0,0 +1,346 @@
...
+	for (int r = 0; r < Nr; r++)
+	{
+		double sumG = 0;
+		for (int gray_i = 0; gray_i < Ng; gray_i++)
+			sumG += NGLDM.yx (gray_i, r);
+		Sr[r] = sumG;
+	}
+}
+
+void NGLDMfeature::calc_features (const std::vector<double>& Sg, const std::vector<double>& Sr, SimpleMatrix<unsigned int>& M, unsigned int roi_area)
```
sameeul, line 185: Question: Does `M` get modified here?

src/nyx/features/ngldm.cpp
```diff
@@ -0,0 +1,346 @@
...
+		// Calculate feature of this ROI
+		NGLDMfeature f;
+		f.calculate(r);
+		f.save_value(r.fvals);
+	}
+}
+
+void NGLDMfeature::osized_add_online_pixel(size_t x, size_t y, uint32_t intensity) {}
+
+void NGLDMfeature::osized_calculate(LR& r, ImageLoader&)
```
sameeul, line 316: Is `ImageLoader&` used here?

src/nyx/features/ngldm.cpp
friskluft: We should but we can't at the moment - a trivial pixel cloud argument would work as const while its oversized variant wouldn't. I need to make multiple changes in the latter that is a subject of a separate PR.

src/nyx/features/ngldm.cpp
friskluft: "r" may be changed, for example, a feature class X can create a per-ROI object and leave it in "r" for another feature class Y that depends on X. With this purpose in mind, method "calculate(.)" is declared up in abstract class "FeatureMethod", parent of "NGLDMfeature". Examples of calculate(LR & r) that modify r's state is the contour feature leaving contour pixels container for the radial distribution features calculated after the contour, the convex hull feature leaving the convex hull pixels container for caliper features calculated after the contour, etc.

src/nyx/features/ngldm.cpp
friskluft: No, therefore its parameter is unnamed. May be modified in other children of class "FeatureMethod".

sameeul (approved):

Issue #116, opened 2023-05-20
Feature group names are case-sensitive


haesleinhuepf:
I was just trying to retrive all features of a given group from Python. The feature group names given [here](https://github.com/PolusAI/nyxus#feature-groups) lead to an error:
```
nyx = Nyxus(['*all_morphology*'])
```
Error:
```
File ~\mambaforge\envs\bio39\lib\site-packages\nyxus\nyxus.py:86, in Nyxus.__init__(self, features, neighbor_distance, pixels_per_micron, coarse_gray_depth, n_feature_calc_threads, n_loader_threads, using_gpu, ibsi)
     83     print("No gpu available.")
     84     using_gpu = -1
---> 86 initialize_environment(
     87     features,
     88     neighbor_distance,
     89     pixels_per_micron,
     90     coarse_gray_depth, 
     91     n_feature_calc_threads,
     92     n_loader_threads,
     93     using_gpu,
     94     ibsi
     95 )

ValueError: Error: '*all_morphology*' is not a valid feature name. 
```

When calling the code like this, it works:
```
nyx = Nyxus(['*ALL_MORPHOLOGY*'])
```

Would it be possible to make the names case-insensitive for user convenience?

PR #117, opened 2023-05-23
POL3115 case insensitive feature names

friskluft:
Addressing issue #116

### DIFF
--- a/src/nyx/featureset.h
+++ b/src/nyx/featureset.h
@@ -254,20 +254,20 @@ namespace Nyxus
 		GLSZM_LAHGLE,	// Large Area High Gray Level Emphasis
 
 		// GLDM:
-		GLDM_SDE,	// Small Dependence Emphasis(SDE)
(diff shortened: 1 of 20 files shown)


src/nyx/featureset.h
```diff
@@ -450,7 +450,7 @@ class FeatureSet
 			m_enabledFeatures[STAT_FERET_DIAM_MODE] = true;
 	}
 	bool isEnabled(int fc) const { return fc < AvailableFeatures::_COUNT_ ? m_enabledFeatures[fc] : false; }
-	bool anyEnabled(std::initializer_list<int> F) const
+	bool anyEnabled(const std::initializer_list<AvailableFeatures> F) const
```
sameeul, line 453: should we pass this by reference?

friskluft: It's not a typo, just a habit from the early times of std::initializer_list when passing a constant list by reference caused a CL's hiccup. Fixed

sameeul (approved):

PR #133, opened 2023-08-11
Update pyarrow import for build without Arrow

JesseMckinzie:
- Add check to see if Nyxus was built with Arrow before importing pyarrow
- Add exception to get_arrow_memory_map() on Python side to prevent crashing


sameeul:
I think a good test will be to remove `pyarrow` from `CIBW_TEST_REQUIRES:` in `build_wheels.yml` and `publish_pypi.yml`.

src/nyx/python/nyxus/nyxus_arrow.py
```diff
@@ -0,0 +1,28 @@
+import os
+import sys
+
+def link_arrow_lib():
+    
+    if arrow_headers_found():
+        import pyarrow as pa
+    
+    if os.sys.platform == "win32":
```
sameeul, line 9: If `arrow_headers_found()` returns `False`, this will result an error. So, we should execute it only for `True` case

sameeul (approved):

PR #142, opened 2023-09-05
POL3416 fixed no-values in ellipse and polygonality features

friskluft:


src/nyx/features/circle.cpp
```diff
@@ -142,16 +137,16 @@ void EnclosingInscribingCircumscribingCircleFeature::findSecondPoint (const std:
     }
 }
 
-void EnclosingInscribingCircumscribingCircleFeature::findMinEnclosingCircle (const std::vector<Pixel2>& contour, Point2f& center, float& radius)
+void EnclosingInscribingCircumscribingCircleFeature::findMinEnclosingCircle (std::vector<Pixel2>& contour, Point2f& center, float& radius)
```
sameeul, line 140: not clear why we are dropping the `const`.

src/nyx/features/hexagonality_polygonality.h
```diff
@@ -23,5 +23,6 @@ class HexagonalityPolygonalityFeature: public FeatureMethod
 
 private:
 	double polyAve = 0, hexAve = 0, hexSd = 0;
+	static const int novalue = -1;	// feature falue made intentionally blank and unsuitable for model training
```
sameeul, line 26: We need to document it somewhere, I think.

friskluft: Later, when I do similar NAN intercept for other feature classes. I may have 3 more ones today

src/nyx/features/circle.cpp
friskluft: Accidentally. Reverting....

JesseMckinzie (approved):

sameeul (approved):

PR #144, opened 2023-09-08
drop support for python 3.7

jfennick:


sameeul:
Can you also drop 3.7 from `publish_pypi.yml`?

sameeul (approved):

PR #146, opened 2023-09-18
POL3417 real-valued intensities for weighted moments (CPU)

friskluft:


src/nyx/features/image_moments.cpp
```diff
@@ -224,13 +207,33 @@ double ImageMomentsFeature::moment (const pixcloud& cloud, int p, int q)
...
+
 void ImageMomentsFeature::calcOrigins (const std::vector<Pixel2>& cloud)
 {
     double m00 = moment (cloud, 0, 0);
     originOfX = moment (cloud, 1, 0) / m00;
     originOfY = moment (cloud, 0, 1) / m00;
 }
 
+void ImageMomentsFeature::calcOrigins (const std::vector<Pixel2>& cloud, const std::vector<float>& real_valued_intensities)
+{
```
sameeul, line 231: Not seeing where `real_valued_intensities` is used.

src/nyx/features/image_moments.cpp
```diff
@@ -445,3 +534,35 @@ void ImageMomentsFeature::gpu_process_all_rois (const std::vector<int> & Labels,
...
+    {
+        dst.reserve (src.size());
+        for (auto pxl : src)
+            dst.push_back (float(pxl.inten));
+    }
+
+    /// @brief Applies to distance-to-contour weighting to intensities of pixel cloud. Saves the result in 'realintens' 
+    void apply_dist2contour_weighting(
+        // input & output
+        std::vector<float>& realintens,
```
sameeul, line 549: maybe just say `intcloud` to stay consistent with the header?

src/nyx/features/image_moments.cpp
friskluft: Fixed in the update

src/nyx/features/image_moments.cpp
friskluft: Fixed in the update, check it out

sameeul (approved):

PR #147, opened 2023-09-20
Add Apache Arrow stream writers

JesseMckinzie:
- Update Apache Arrow writers for Arrow IPC and Parquet formats to write in streams to make Arrow writing fully scalable
- Update Python API and CLI to use new writers
- Add unit tests for updated get_arrow_table method in the Python API


src/nyx/main_nyxus.cpp
```diff
@@ -3,7 +3,8 @@
 #include "dirs_and_files.h"
 #include "environment.h"
 #include "globals.h"
-
+#include "arrow_output_stream.h"
+#include <glog/logging.h>
```
sameeul, line 7: This is actually causing the build failure. Do we need this?

src/nyx/scan_fastloader_way.cpp
```diff
@@ -26,11 +26,19 @@ namespace py = pybind11;
...
+#include "arrow_output_stream.h"
+#include "output_writers.h"
+#endif
+
 // Sanity
 #ifdef _WIN32
 #include<windows.h>
 #endif
 
+#include <chrono>
```
sameeul, line 39: Do we need this?

src/nyx/scan_fastloader_way.cpp
```diff
@@ -252,15 +273,35 @@ namespace Nyxus
...
+					ok = save_features_2_csv(ifp, lfp, outputDir);
+				} else {
+					ok = save_features_2_buffer(theResultsCache);
+				}
+
+				if (ok == false)
+				{
+					std::cout << "save_features_2_csv() returned an error code" << std::endl;
+					return 2;
+				}
```
sameeul, line 304: This need to be updated to refer to proper function call. May be something like 
```
if (!ok){
  if(save2csv){...}
  else {...}
}
```

src/nyx/scan_fastloader_way.cpp
```diff
@@ -252,15 +273,35 @@ namespace Nyxus
...
+			if (arrow_output) {
+
+				auto features = Nyxus::get_feature_values();
+
+				auto status = theEnvironment.arrow_writer->write(Nyxus::get_feature_values());
+				
+				if (!status.ok()) {
+                    // Handle read error
+                    auto err = status.ToString();
+                    throw std::runtime_error("Error writing Arrow file: " + err);
```
sameeul, line 286: Should we `throw`? This can be performance degrading if we are in a situation where we keep throwing. Are we catching this in the caller code? If not, may be just ignore/log and move on? 
Since this function already return error codes if anything goes wrong, we should just use the same error reporting mechanism instead of `throw`.

src/nyx/main_nyxus.cpp
```diff
@@ -68,7 +75,8 @@ int main (int argc, char** argv)
 		theEnvironment.n_pixel_scan_threads, 
 		theEnvironment.n_reduce_threads,
 		min_online_roi_size,
-		theEnvironment.useCsv, // 'true' to save to csv
+		use_arrow, // 'true' to save to csv
+		theEnvironment.useCsv,
```
sameeul, line 79: Update the arg comment :-)

src/nyx/output_writers.h
```diff
@@ -16,9 +20,20 @@
 #include <memory>
 
 #include "helpers/helpers.h"
+#include "globals.h"
```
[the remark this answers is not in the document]
JesseMckinzie: This is no longer needed

src/nyx/output_2_buffer.cpp
```diff
@@ -27,6 +27,7 @@
 
 namespace Nyxus
 {
+
```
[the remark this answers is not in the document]
JesseMckinzie: This was left behind after removing a function I implemented

src/nyx/scan_fastloader_way.cpp
```diff
@@ -252,15 +271,35 @@ namespace Nyxus
 				return 1;
 			}
 
+		#ifdef USE_ARROW
+			if (arrow_output) {
+
+				auto features = Nyxus::get_feature_values();
```
sameeul, line 277: I guess this is an extra call that we don't need?

src/nyx/arrow_output_stream.cpp
```diff
@@ -0,0 +1,49 @@
...
+                                                         const std::string& arrow_file_path,
+                                                         const std::vector<std::string>& header) {
+    
+    std::string arrow_file_type_upper = Nyxus::toupper(arrow_file_type);
+
+    if(arrow_file_path != "" && !fs::is_directory(arrow_file_path) && !(Nyxus::ends_with_substr(arrow_file_path, ".arrow") || Nyxus::ends_with_substr(arrow_file_path, ".feather") || Nyxus::ends_with_substr(arrow_file_path, ".parquet"))) {
+        throw std::invalid_argument("The arrow file path must end in \".arrow\"");
+    }
+
+    if (!(arrow_file_type_upper == "ARROW" || arrow_file_type_upper == "ARROWIPC" || arrow_file_type_upper == "PARQUET")) {
```
friskluft, line 15: How a user can specify via arrow_file_type_upper that the desired format is ".feather" ?

src/nyx/globals.h
```diff
@@ -32,7 +32,7 @@ namespace Nyxus
 
 	bool scanFilePairParallel(const std::string& intens_fpath, const std::string& label_fpath, int num_fastloader_threads, int num_sensemaker_threads, int filepair_index, int tot_num_filepairs);
 	std::string getPureFname(const std::string& fpath);
-	int processDataset(const std::vector<std::string>& intensFiles, const std::vector<std::string>& labelFiles, int numFastloaderThreads, int numSensemakerThreads, int numReduceThreads, int min_online_roi_size, bool save2csv, const std::string& csvOutputDir);
+	int processDataset(const std::vector<std::string>& intensFiles, const std::vector<std::string>& labelFiles, int numFastloaderThreads, int numSensemakerThreads, int numReduceThreads, int min_online_roi_size, bool save2csv, bool arrow_output, const std::string& csvOutputDir);
```
friskluft, line 35: Wouldn't having overloaded function processDataset() for the separate cases of CSV and Apache output types be less confusing than having a single overparametered processDataset() ?

src/nyx/main_nyxus.cpp
```diff
@@ -59,6 +59,13 @@ int main (int argc, char** argv)
 	auto startTS = getTimeStr();
 	VERBOSLVL1(std::cout << "\n>>> STARTING >>> " << startTS << "\n";)
 
+
+	bool use_arrow = false;
+
+#ifdef USE_ARROW
+	use_arrow = theEnvironment.arrow_output_type == "ARROW" || theEnvironment.arrow_output_type == "PARQUET";
+#endif
+
```
friskluft, line 68: This piece of logic is OK but is just somewhat not elegant to put such a low-level stuff in main(). Again, if we overload processDataset() as mentioned earlier, we would be able to call it separately with Apache-related and CSV parameters.

src/nyx/scan_fastloader_way.cpp
```diff
@@ -367,6 +452,19 @@ namespace Nyxus
...
+			// close arrow file after use
+			auto status = theEnvironment.arrow_writer->close();
+			
+			if (!status.ok()) {
+				// Handle read error
+				error_message = "Error closing Arrow file: " + status.ToString();
+				return 2;
+			}
+		}
+	#endif
```
friskluft, line 467: File scan_fastloader_way.cpp is getting cluttered with #ifdef USE_ARROW code. Which is OK for this PR but I would immediately fix in a subsequent PR to keep the image scan logic free of low-level file format stuff. (Principle of separation of concerns.)

friskluft (approved):

sameeul (approved):

PR #150, opened 2023-10-03
updated 2d-moments

friskluft:


src/nyx/gpu/geomoments_central.cu
```diff
@@ -0,0 +1,234 @@
...
+
+    //=== device-reduce:
+    // Determine temporary device storage requirements
+    double* d_out = nullptr;
+    CHECKERR(cudaMalloc(&d_out, sizeof(double)));
+    void* d_temp_storage = NULL;
+    size_t   temp_storage_bytes = 0;
+    cub::DeviceReduce::Sum(d_temp_storage, temp_storage_bytes, Nyxus::devPrereduce/*d_in*/, d_out, cloudlen/*num_items*/);
+    // Allocate temporary storage
+    cudaMalloc(&d_temp_storage, temp_storage_bytes);
```
JesseMckinzie, line 94: Do we need to check the error status of cudaMalloc (and other allocations missing error check)? i.e. 

ok = cudaMalloc(&d_temp_storage, temp_storage_bytes);
CHECKERR(ok);

src/nyx/gpu/geomoments_raw.cu
```diff
@@ -0,0 +1,224 @@
...
+    const Pixel2 *d_roicloud,
+    size_t cloudlen,
+    StatsInt base_x,
+    StatsInt base_y)
+{
+    int nblo = whole_chunks2(cloudlen, blockSize);
+    kerRawMoment <<< nblo, blockSize >>> (Nyxus::devPrereduce, d_roicloud, cloudlen, base_x, base_y, p, q);
+
+    CHECKERR (cudaPeekAtLastError());
+    CHECKERR (cudaGetLastError());
```
JesseMckinzie, line 84: Do we need both cudaPeekAtLastError and cudaGetLastError?

src/nyx/gpu/geomoments_central.cu
friskluft: We have 2 error-checking options here: (1) not bother of checking cudaMalloc() relying on internal error handling in Sum() since those functions are coupled (2) do error checking defensively forgetting that cudaMalloc() allocates a tiny chunk of memory and Sum() is the only point of failure. Opted for (2).

src/nyx/gpu/geomoments_raw.cu
friskluft: Having cudaPeekAtLastError() is the maximum defensive way to run kernels because unlike the pair cudaDeviceSynchronize()-cudaGetLastError() checking the error status of all the finished threads, the former checks the thread blocks launch phase, for example, if a block of requested size is available (someone may tweak #define blockSize in the header) or if requested amount of shared memory can be allocated. I'd leave cudaPeekAtLastError() until we fully optimize and freeze the launch parameters (the block size and kernel).

PR #153, opened 2023-10-17
Arrow writers refactor

JesseMckinzie:
- Refactor process dataset functions to be overloaded to separate csv/buffer functionality from Arrow functionality
- Remove passing status to get_arrow_table and check for nullptr instead
- Remove low level functionality from nyxus main to process functions and the environment


src/nyx/environment.h
```diff
@@ -117,12 +117,12 @@ class Environment: public BasicEnvironment
 
 	bool singleROI = false; // is set to 'true' parse_cmdline() if labels_dir==intensity_dir
 
-#ifdef USE_ARROW
 
+	bool use_arrow = false;
```
friskluft, line 121: "use_arrow" will be defined as a class member no matter if we have Apache functionality enabled or disabled. Can we have no reference to Arrow if USE_ARROW is not defined ?

src/nyx/globals.h
```diff
@@ -32,7 +32,8 @@ namespace Nyxus
 
 	bool scanFilePairParallel(const std::string& intens_fpath, const std::string& label_fpath, int num_fastloader_threads, int num_sensemaker_threads, int filepair_index, int tot_num_filepairs);
 	std::string getPureFname(const std::string& fpath);
-	int processDataset(const std::vector<std::string>& intensFiles, const std::vector<std::string>& labelFiles, int numFastloaderThreads, int numSensemakerThreads, int numReduceThreads, int min_online_roi_size, bool save2csv, bool arrow_output, const std::string& csvOutputDir);
+	int processDataset(const std::vector<std::string>& intensFiles, const std::vector<std::string>& labelFiles, int numFastloaderThreads, int numSensemakerThreads, int numReduceThreads, int min_online_roi_size, bool save2csv, const std::string& csvOutputDir);
+	int processDataset(const std::vector<std::string>& intensFiles, const std::vector<std::string>& labelFiles, int numFastloaderThreads, int numSensemakerThreads, int numReduceThreads, int min_online_roi_size, const std::string& outputDir);
```
friskluft, line 36: Hard to guess what function fits a CSV, memory buffer, or Apache output usage. Could we have argument "save2csv" of enum type (values "save2csv", "save2buffer", "save2arrow", "save2parquet", etc) rather than bool or at least document these functions with a comment block ?

src/nyx/main_nyxus.cpp
```diff
@@ -58,26 +58,34 @@ int main (int argc, char** argv)
...
+			intensFiles, 
+			labelFiles, 
+			theEnvironment.n_loader_threads, 
+			theEnvironment.n_pixel_scan_threads, 
+			theEnvironment.n_reduce_threads,
+			min_online_roi_size,
+			theEnvironment.useCsv,
+			theEnvironment.output_dir);
+
+	}
```
friskluft, line 88: Awkward code, to my taste. Could we have a single call and use a destination type parameter or a more elegant solution? (We are in no rush with this refactor.)

src/nyx/main_nyxus.cpp
```diff
@@ -93,6 +101,8 @@ int main (int argc, char** argv)
 		case 3:		// Memory error
 			std::cout << std::endl << "Memory error" << std::endl;
 			break;
+		case 4:
+			std::cout << std::endl << "Arrow not enabled" << std::endl;
```
friskluft, line 105: Maybe it's worth adding a couple of words how to enable, reinstall, or reset something? If Arrow libraries aren't in place, the executable/library wouldn't even run, so a user may be clueless why Arrow may still not be enabled.

src/nyx/python/new_bindings_py.cpp
```diff
@@ -173,22 +173,36 @@ py::tuple featurize_directory_imp (
     // We're good to extract features. Reset the feature results cache
     theResultsCache.clear();
 
-    auto arrow_output = !pandas_output;
+    theEnvironment.use_arrow = !pandas_output;
```
friskluft, line 176: My traditional question here: are we supporting other Apache formats - "feather" and "parquet"? If yes, then "theEnvironment.use_arrow" should be called "theEnvironment.use_apache" ...

src/nyx/globals.h
JesseMckinzie: Updated to use an enum to choose save type and added in empty classes to replace ifdef statements, as discussed with Samee. Let me know if you agree with this approach

sameeul (approved):

friskluft (approved):

PR #154, opened 2023-10-20
POL3595 unified column name | customizable output file name

friskluft:


src/nyx/environment.cpp
```diff
@@ -826,6 +830,13 @@ bool Environment::parse_cmdline(int argc, char **argv)
 		rawFeatures = FEA_NICK_ALL;
 	}
 
+	//==== Parse optional result file name
+	if (nyxus_result_fname == "")
```
sameeul, line 834: is this going to be a mandatory argument?

friskluft: No, optional. Default ('NyxusFeatures' .csv, .arrow, etc) remains the same

PR #155, opened 2023-10-23
fixed reading TIFF real valued intensities in range 0-1 as integer 0

friskluft:


src/nyx/image_loader1x.cpp
```diff
@@ -40,7 +41,10 @@ bool ImageLoader1x::open(const std::string& fpath)
 		else 
 		{
 			if (Nyxus::check_tile_status(fpath))
-				FL = std::make_unique<NyxusGrayscaleTiffTileLoader<uint32_t>> (n_threads, fpath); 
+				FL = std::make_unique<NyxusGrayscaleTiffTileLoader<uint32_t>> (n_threads, fpath,
```
sameeul, line 44: Why are we passing in globals this way instead of modifying the ImageLoader1x constructor?

src/nyx/grayscale_tiff.h
```diff
@@ -165,10 +174,10 @@ class NyxusGrayscaleTiffTileLoader : public AbstractTileLoader<DataType>
             case 8:
             case 16:
             case 32:
-                loadTile<float>(tiffTile, tileDataVec);
+                loadTile_real_intens <float> (tiffTile, tileDataVec);
```
sameeul, line 177: can we do something like 
```
if (image_options_are_present) {
  loadTile_real_intens(...)
} else {
  loadTile(...)
}
```
That way, if we are not dealing with already scaled image (which is not the common case, we don't spend tile inspecting each pixel value while copying.

```

File: src/nyx/image_loader.cpp [hunk 3]
```diff
@@ -41,7 +44,12 @@ bool ImageLoader::open(const std::string& int_fpath, const std::string& seg_fpat
 		{
 			if (Nyxus::check_tile_status(int_fpath))
 			{
-				intFL = new NyxusGrayscaleTiffTileLoader<uint32_t>(n_threads, int_fpath);
+				intFL = new NyxusGrayscaleTiffTileLoader<uint32_t> (
```
sameeul, line 47: Instead of using globals here, can we modify the open() to pass the `fimageOptions`?

friskluft: My motive to access the global settings as close to the point of use as possible, as just one control flow branch of many passes through the floating point pixels/voxels manipulation. Otherwise we would need to prepare those global settings to pass as parameters early amid the high-level code of phase 2 (trivials) and 3 (non-trivials) image processing logic where we are not concerned about byte layouts in pixels, and do it twice - in function processDataset() and processNontrivialRois(). Your opinion?

src/nyx/grayscale_tiff.h
friskluft: Floating point image options (in the form of expected minimum, maximum, and equivalent integer range) are always present, aren't they? And doing your suggested way, we'd need to do a separate pre-scan through an image to figure out its intensity range, that is also a time expense.

sameeul (approved):

PR #170, opened 2023-11-15
Python API update to match CLI

JesseMckinzie:
- Add custom file names for Nyxus output
- Add floating point image parameters to Python API
- Remove pyarrow dependency 
- Update Arrow unit tests for custom file names
- Update unit tests for getting and setting new parameters


src/nyx/arrow_output_stream.cpp
```diff
@@ -12,95 +12,83 @@
...
-                } 
+        if (path != "" && !fs::exists(path)) {
+            try {        
+                fs::create_directory(path);
+            } catch (std::exception& e) {
+                return {false, e.what()};
             }
-        return false;
-    }(); 
+        }
```
sameeul, line 32: We probably do not this anymore, since the `output_path` should now contain a fully qualified path. It will fail if this is not honored. You have already implemented the logic to get a correct FQP in the helper function. 
If `output_path` is empty, we should return `{false, "No file path provided"}` otherwise keep moving.

src/nyx/helpers/helpers.h
```diff
@@ -324,5 +336,37 @@ namespace Nyxus
...
+        auto current_ext = fs::path(path).extension().u8string();
+
+        if (current_ext == "") {
+            arrow_file_path += file_extension;
+        } else {
+            if (current_ext != file_extension) {
+                std::cerr << "Incorrect file extension \"" + current_ext + "\". Using correct extension \"" + file_extension + "\"." << std::endl;
+
+                auto fs_path = fs::path(arrow_file_path);
+                arrow_file_path = fs_path.remove_filename().u8string() + fs_path.stem().u8string() + file_extension; // correct the extension
```
sameeul, line 365: this can be simplified by `fs::path::replace_extension(...)` .

src/nyx/helpers/helpers.h
```diff
@@ -324,5 +336,37 @@ namespace Nyxus
 		else
 			return x;
 	}
+
+	inline std::string get_arrow_path (const std::string& path, const std::string& file_name, const SaveOption& arrow_file_type) {
+
```
sameeul, line 341: May be create a `fs::path` object here from `path` and then do all the check and manipulation using native `fs` library instead of string manipulation. And then at the very end return string() from that path object.

sameeul: Can do something like the following 
```
std::string get_arrow_filename(const std::string& output_path, const std::string& default_filename, const SaveOption& arrow_file_type){

    std::string valid_ext = [&arrow_file_type](){
        if (arrow_file_type == Nyxus::SaveOption::saveArrowIPC) {
            return ".arrow";
        } else if (arrow_file_type == Nyxus::SaveOption::saveParquet) {
            return ".parquet";
        } else {return "";}
    }();
  
  if (output_path != ""){
    auto arrow_path = std::filesystem::path(output_path);
    // case 1: output_path is a directory, verify by checking if the directory exists 
    if (std::filesystem::is_directory(arrow_path)){
      // append default filename with correct extension
      arrow_path = arrow_path/default_filename;
      arrow_path.replace_extension(valid_ext);
    } 
    // case 2: output_path is of directory format, but it does not yet exists
    else if(!arrow_path.has_extension()) {
      // create directory
        std::filesystem::create_directory(arrow_path);
        // append default filename with correct extension
        arrow_path = arrow_path/default_filename;
        arrow_path.replace_extension(valid_ext);
    } 
    // case 3 : output_path appears a valid FQP, has filename and/or ext
    else {
      if (arrow_path.extension().string() != valid_ext){
        // update extension to match output option
        arrow_path.replace_extension(valid_ext);
      }
      // check if the output dir exists
      if (!std::filesystem::is_directory(arrow_path.parent_path())){
        // if not, create it
        std::filesystem::create_directories(arrow_path.parent_path());
      }
    }
    return arrow_path.string();
  } else {
    // case 4: output_path is empty, return default filename with valid ext
    return default_filename + valid_ext;
  }  
}
```

sameeul (approved):

PR #190, opened 2023-12-22
simplified Feret calculation

friskluft:


src/nyx/helpers/helpers.h
```diff
@@ -324,4 +324,32 @@ namespace Nyxus
...
+
+		return idx;
+	}
+
+	inline size_t get_max_idx(const std::vector<double>& vec)
+	{
+		auto max_ = vec[0];
+		size_t idx = 0;
+		for (size_t i = 0; i < vec.size(); ++i)
+			if ((max_ - vec[i] < GET_MINMAX_IDX_EPSILON_DIFF) && (fabs(vec[i] - max_) > GET_MINMAX_IDX_EPSILON_DIFF))
```
sameeul, line 348: just curious, is this check is correct?

friskluft: I'll simplify those comparison

src/nyx/features/caliper_feret.cpp
```diff
@@ -25,47 +24,28 @@ void CaliperFeretFeature::calculate(LR& r)
...
-	{
-		// Degenerate case
-		minFeretDiameter =
-		maxFeretDiameter =
-		minFeretAngle =
-		maxFeretAngle = 0.0;
-	}
+	// Angles
+	size_t min_i = Nyxus::get_min_idx (ferets);
+	size_t max_i = Nyxus::get_max_idx (ferets);
```
sameeul, line 42: Another question: Now that you have reworked the calculation, can we just use `std::min_element` and `std::max_element` here or it might produce different result based on OS/HW types?

src/nyx/features/caliper_feret.cpp
```diff
@@ -67,83 +53,27 @@ void CaliperFeretFeature::save_value(std::vector<std::vector<double>>& fvals)
 	fvals[(int)Feature2D::STAT_FERET_DIAM_MODE][0] = _mode;
 }
 
-void CaliperFeretFeature::calculate_imp(const std::vector<Pixel2>& convex_hull, std::vector<double>& all_D)
+//void CaliperFeretFeature::calculate_imp (const std::vector<Pixel2>& convex_hull, std::vector<double>& D_min, std::vector<double>& D_max)
```
sameeul, line 56: Can we remove this line of dead code for clarity?

sameeul (approved):

PR #191, opened 2024-01-05
Add pytests for featurize_directory and featurize_files

JesseMckinzie:
- Add pytests for featurize_directory and featurize_files
- Add tissuenet data for tests


tests/python/test_tissuenet_data.py
```diff
@@ -0,0 +1,21 @@
...
+    
+with BioReader(path + '/data/seg/p0_y1_r1_c0.ome.tif') as br:
+    seg_p0_y1_r1_c0 = br.read()
+    
+with BioReader(path + '/data/seg/p0_y1_r1_c1.ome.tif') as br:
+    seg_p0_y1_r1_c1 = br.read()
+
+tissuenet_int = np.array([int_p0_y1_r1_c0, int_p0_y1_r1_c1])
+tissuenet_seg = np.array([seg_p0_y1_r1_c0, seg_p0_y1_r1_c1])
+print(tissuenet_int)
```
sameeul, line 21: remove the `print` please.

tests/python/test_nyxus.py
```diff
@@ -34,8 +37,79 @@ def teardown_class(cls):
...
+            for col in directory_columns:
+                directory_list = directory_features[col].tolist()
+                files_list = files_features[col].tolist()
+                
+                for directory_val, files_val in zip(directory_list, files_list):
+                    if not directory_val == pytest.approx(files_val, rel=1e-5, abs=1e-5):
+                        files_not_equal.append(col)
+                        break
+                    
+            print(files_not_equal)
```
sameeul, line 76: I think we can remove `print`. If necessary, we can modify the `assert` to print a custom message when it fails, something like this:
`assert len(files_not_equal) == 0, f"Files that are not equal:\n {files_not_equal}"`

sameeul (approved):

PR #196, opened 2024-01-25
Add CUDA wheels build and PyPI publishing to Github actions

JesseMckinzie:
- Add builds for CUDA 11 and CUDA 12 wheels in Github actions
- Add publishing of CUDA 11 and CUDA 12 wheels in Github actions
- Update name of wheels for GPU wheels in setup.py


sameeul (approved):
I pushed some minor changes and also removed the test that you added. Since we now FATAL out for `USEGPU=ON` and GPU not being available during the build phase, that test is redundant.

PR #216, opened 2024-04-23
Add image quality features

JesseMckinzie:
- Add features for calculation image quality from https://github.com/hamshkhawar/image-tools/tree/quality/polus-image-quality-plugin


CMakeLists.txt
```diff
@@ -258,7 +263,7 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 	find_package(Threads QUIET)
 	if (Threads_FOUND)
 		if (CMAKE_USE_PTHREADS_INIT)
-			SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
+			SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -g")
```
sameeul, line 266: Why are we adding debug flag here?

src/nyx/helpers/helpers.h
```diff
@@ -355,4 +357,181 @@ namespace Nyxus
...
+
+			padded.insert(padded.end(), row.begin(), row.end());
+		}
+		
+		
+		return padded;
+	}
+
+
+	inline std::vector<int> arrange(int start, int end, int step=1) {
```
sameeul, line 419: This should be replaced by `std::copy`

src/nyx/helpers/helpers.h
```diff
@@ -355,4 +357,181 @@ namespace Nyxus
...
+
+		for (std::map<int, double>::iterator it = sum_values.begin(); it != sum_values.end(); ++it) {
+			out.push_back(it->second);
+		}
+
+		return out;
+	}
+
+
+	inline std::vector<std::vector<int>> flipud (std::vector<std::vector<int>> vec) {
```
sameeul, line 455: use `std::reverse`?

docs/source/imagequality.rst
```diff
@@ -0,0 +1,29 @@
+Image Quality features
```
friskluft, line 1: We need to add the new features' math description, too. (Math/focus_score.rst, Math/psd_slope.rst, etc)

src/nyx/featureset.h
```diff
@@ -457,6 +457,15 @@ namespace Nyxus
 		WEIGHTED_HU_M6,
 		WEIGHTED_HU_M7,
 
+		// Image Quality features
+		FOCUS_SCORE,
```
friskluft, line 461: I suggest naming these and future image quality features in a distinctive way, for example, IMQ_FOCUS_SCORE, IMQ_SHARPNESS, etc in order to emphasize their applicability to the whole slide, not to a ROI.

src/nyx/features/brisque.cpp
```diff
@@ -0,0 +1,128 @@
...
+    brisque(Im0);
+}
+
+void BrisqueFeature::brisque(const ImageMatrix& Im) {
+
+    readOnlyPixels image = Im.ReadablePixels();
+    int rows = Im.height;
+    int cols = Im.width;
+
+    int kernel_size = 7;
```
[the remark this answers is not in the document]
friskluft: Or at least mention those hardcoded constants in feature's math description

friskluft (commented):
Image quality features are supposed to be applicable to the whole image. 
Are we letting them be calculated on the per-ROI basis? (This depends on Hamda's workflow and how she envisions their calculation in the segmented case. I guess she is going to calculate them in the "image quality evaluation and adjustment" step prior to the "feature extraction step".)

src/nyx/features/brisque.cpp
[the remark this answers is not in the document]
JesseMckinzie: These are constant values. I'll mention the constants in the math description. Some of the other features such as local focus score will require values to be set via the APIs. I will update the APIs to take in these values

src/nyx/features/power_spectrum.cpp
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    return 0;
+}
+
+std::vector<double> PowerSpectrumFeature::invariant(std::vector<unsigned int> image) {
+
+    
+    double mean_value = std::accumulate(image.begin(), image.end(), 0.0)/image.size();
```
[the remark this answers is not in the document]
JesseMckinzie: How will this always be an integer? std::accumulate returns a double because the initial value is a double and then this double is divided by an integer, which can end up as a double with a decimal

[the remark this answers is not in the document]
friskluft: You are right, return type comes from the 3rd argument. 
My habit to assume it to come from the container elements is outdated.

src/nyx/features/brisque.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    for (int y = 0; y < kernel_size; ++y) {
+        for (int x = 0; x < kernel_size; ++x) {
+            double x_dist = x - center;
+            double y_dist = y - center;
+            kernel[y * kernel_size + x] = 1 / (2 * M_PI * sigma * sigma) * exp(-(x_dist * x_dist + y_dist * y_dist) / (2 * sigma * sigma));
+        }
+    }
+
+    return normalize(kernel);
```
sameeul, line 110: I would suggest to do the normalizing math operations here instead of calling this function since this function creates another extra vector.

src/nyx/features/brisque_gaussian.h
```diff
@@ -0,0 +1,366 @@
...
+
+        double sum = 0.0;
+        for (double val : _x_vec) {
+            sum += val * val;
+        }
+
+        return std::sqrt(sum / _x_vec.size());
+    }
+
+    double mean_squares(std::vector<double> x_vec) {
```
sameeul, line 258: need to pass const ref

JesseMckinzie: This is a part of the Brisque feature that we are removing for now. I will remove these files now

src/nyx/roi_cache.cpp
```diff
@@ -91,7 +91,7 @@ std::vector<StatsReal> LR::get_fvals (int fcode)
 
 void LR::initialize_fvals()
 {
-	fvals.resize ((int)Nyxus::Feature3D::_COUNT_);
+	fvals.resize ((int)Nyxus::FeatureIMQ::_COUNT_);
 	for (auto& valVec : fvals)
 		valVec.push_back(0.0);
 }
```
sameeul, line 97: I am not sure if this is correct. At a given run, we other either computing 2D or 3D or IMQ features. In that case, `Nyxus::Feature2D::_COUNT_` should not be sufficient (at the moment), since that's max number of features we are going to have at a given run?

@friskluft : Your thoughts?

tests/test_all.cc
```diff
@@ -19,20 +19,10 @@
...
-	test_file_naming();
-}
-
-TEST(TEST_NYXUS, TEST_ARROW) {
-	test_arrow();
-}
-
-TEST(TEST_NYXUS, TEST_PARQUET) {
-	test_parquet();
-}
```
sameeul, line 26: Need to enable these back.

tests/test_feature_calculation.h
```diff
@@ -0,0 +1,52 @@
...
+
+
+#include "../src/nyx/roi_cache.h"
+#include "../src/nyx/parallel.h"
+#include "../src/nyx/features/pixel.h"
+#include "../src/nyx/environment.h"
+#include "test_data.h"
+#include "test_main_nyxus.h"
+
+// check values of multiple features
```
sameeul, line 11: These should go inside namespace `Nyxus`

tests/test_image_quality.h
```diff
@@ -0,0 +1,67 @@
...
+#include "test_data.h"
+#include "test_main_nyxus.h"
+#include "test_feature_calculation.h"
+
+#include "../src/nyx/features/focus_score.h"
+#include "../src/nyx/features/power_spectrum.h"
+#include "../src/nyx/features/saturation.h"
+#include "../src/nyx/features/sharpness.h"
+
+/* GLCM dissimilarity and correlation for image quality are handled in GLCM tests */
```
sameeul, line 16: Also, put them under `Nyxus` namespace

test.py
```diff
@@ -0,0 +1,8 @@
+from nyxus import Nyxus 
+
+
+nyx = Nyxus(["FOCUS_SCORE"])
+
+features = nyx.featurize_directory("/Users/mckinziejr/Documents/GitHub/nyxus/data/int", "/Users/mckinziejr/Documents/GitHub/nyxus/data/int")
+
+print(features)
```
sameeul, line 8: This is probably added by mistake

src/nyx/features/sharpness.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
+
+#include <vector>
+#include "../helpers/helpers.h"
+#include "../roi_cache.h"
+#include "image_matrix.h"
+#include "../feature_method.h"
+#include "../environment.h"
+
+#define EPSILON 1e-8
```
sameeul, line 10: There is already an `EPSILON` variable in Nyxus somwhere. Use a `constexpr` instead of macro

tests/test_feature_calculation.h
```diff
@@ -0,0 +1,42 @@
...
+#include "../src/nyx/roi_cache.h"
+#include "../src/nyx/parallel.h"
+#include "../src/nyx/features/pixel.h"
+#include "../src/nyx/environment.h"
+#include "test_data.h"
+#include "test_main_nyxus.h"
+namespace Nyxus
+{
+    // check value of single feature
+    void test_truth(const std::vector<double> values, double truth_value, double frac_tolerance) {
```
sameeul, line 13: not performance critical but this function  should be like the following: 
```
void test_truth(const  double& value, double truth_value, double frac_tolerance){
     ASSERT_TRUE(agrees_gt(value, truth_value, frac_tolerance));
    }
```

PR #217, opened 2024-05-16
Pol4815 mocking alt libraries (1)

friskluft:


tests/test_all.cc
```diff
@@ -19,6 +19,7 @@
 #include "test_glszm.h"
 #include "test_ngtdm.h"
 #include "test_roi_blacklist.h"
+#if 0
```
sameeul, line 22: any particular reason for disabling these tests?

src/nyx/scan_fastloader_way.cpp
```diff
@@ -91,6 +91,23 @@ namespace Nyxus
 		}  
 	}
 
+	std::string group_thousands (unsigned int x)
```
sameeul, line 94: Seems like this is not used anywhere in the code.

tests/test_all.cc
friskluft: Good find! I overlooked reenabling that fragment.
During development, I have to disable stuff referencing Apache on my development machine every time because it fails my builds and ALLEXTRAS/NOEXTRAS does not fully isolate me from Apache dependencies. 
By the way, where we are at plans to separate Nyxus I/O into a library binary in order not to compile the infrastructure and platform specific code every time Nyxus' algorithmic part is compiled?

src/nyx/scan_fastloader_way.cpp
friskluft: Removed from PR

sameeul (approved):

tests/test_glcm.h
```diff
@@ -13,47 +13,53 @@
 #include <unordered_map> 
 
 // Digital phantom values for intensity based features
-static std::unordered_map<std::string, float> glcm_values {
```
JesseMckinzie, line 16: Why is the IBSI data being changed?

JesseMckinzie (approved):

PR #244, opened 2024-11-22
POL5526 automatic dynamic range for floating-point slides

friskluft:


src/nyx/raw_format.h
```diff
@@ -0,0 +1,67 @@
...
+
+class RawFormatLoader
+{
+private:
+    std::string name_, filePath_;
+    size_t numberThreads_;
+
+public:
+
+    RawFormatLoader (std::string_view const& name, std::string filePath)
```
JesseMckinzie, line 14: Can filePath be a reference or const reference to avoid copying?

PR #264, opened 2025-04-04
POL5719 anisotropy

friskluft:


CMakeLists.txt
sameeul, line 1: Revert Line 1, the Cmake issue is not coming from here. It is already at 3.20 (which is >3.5). Same comment for the tests/CMakeLists.txt. #265 will fix the Cmake issue

PR #283, opened 2025-07-07
chore: replace deprecated ci container

sameeul:


sameeul:
CUDA 11.x wheels for Windows are not produced anymore since Github runner does not provide a valid combination of CUDA 11.x and MSVS C++ compiler in visual studio 2022.

PR #328, opened 2026-03-11
Add configurable binning origin for texture features

walkerbdev:
Add binning_origin parameter to control where the intensity binning range starts for texture features: "zero" (default, bins from [0, max], Nyxus/MATLAB behavior) or "min" (bins from [min, max], PyRadiomics-compatible)

### DIFF
--- a/src/nyx/environment.cpp
+++ b/src/nyx/environment.cpp
@@ -191,6 +191,8 @@ void Environment::show_cmdline_help()
 		<< "\t\t\tDefault: 5 \n"
 		<< "\t\t" << OPT << clo_COARSEGRAYDEPTH << "=<custom number of grayscale levels> \n"
 		<< "\t\t\tDefault: 64 \n"
+		<< "\t\t" << OPT << clo_BINNINGORIGIN << "=<zero|min> Origin of intensity binning range \n"
+		<< "\t\t\t'zero' bins from [0, max] (default), 'min' bins from [min, max] (PyRadiomics-compatible) \n"
 		<< "\t\t" << OPT << clo_GLCMANGLES << "=<one or more comma separated rotation angles from set {0, 45, 90, and 135}> \n"
 		<< "\t\t\tDefault: 0,45,90,135 \n"
 		<< "\t\t" << OPT << clo_VERBOSITY << "=<levels of verbosity 0 (silence), 1 (minimum output), 2 (1 + timing), 3 (2 + roi metrics + more timing), 4 (3 + diagnostic information)> \n"
@@ -441,6 +443,7 @@ bool Environment::parse_cmdline(int argc, char** argv)
 			find_string_argument(i, clo_GLCMOFFSET, glcmOptions.rawOffs) ||
 			find_string_argument(i, clo_PXLDIST, pixel_distance) ||
 			find_string_argument(i, clo_COARSEGRAYDEPTH, raw_coarse_grayscale_depth) ||
+			find_string_argument(i, clo_BINNINGORIGIN, raw_binning_origin) ||
 			find_string_argument(i, clo_VERBOSITY, rawVerbosity) ||
 			find_string_argument(i, clo_IBSICOMPLIANCE, raw_ibsi_compliance) ||
 			find_string_argument(i, clo_RAMLIMIT, rawRamLimit) ||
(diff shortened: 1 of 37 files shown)


src/nyx/features/texture_feature.h
```diff
@@ -20,88 +21,76 @@ class TextureFeature
...
 			auto n = I.size();
 			for (size_t i = 0; i < n; i++)
-				S[i] = I[i];
+				S[i] = to_grayscale_radiomix (I[i], min_I_inten, max_I_inten, n_levels);
+			return;
 		}
+		// matlab binning (BinningOrigin::zero)
+		auto n = I.size();
+		prep_bin_array_matlab (max_I_inten, n_levels);
+		for (size_t i = 0; i < n; i++)
```
vjaganat90: Use for_each or range-based for if order evaluation is not important.

src/nyx/features/texture_feature.h
```diff
@@ -20,88 +21,76 @@ class TextureFeature
...
-			}
-			else
-			{
-				// no binning (IBSI)
-				S.assign (I.begin(), I.end());
-			}
+		// matlab binning (BinningOrigin::zero)
+		auto n = I.size();
+		prep_bin_array_matlab(max_I_inten, n_levels);
+		for (size_t i = 0; i < n; i++)
```
vjaganat90: same as above.

src/nyx/python/new_bindings_py.cpp
```diff
@@ -147,14 +149,15 @@ void set_environment_params_imp (
     const std::vector<std::string> &features = {},
     int neighbor_distance = -1,
     float pixels_per_micron = -1,
-    uint32_t coarse_gray_depth = 0, 
+    int coarse_gray_depth = 0,
```
vjaganat90: why change from unsigned to signed int?

src/nyx/python/new_bindings_py.cpp
```diff
@@ -147,14 +149,15 @@ void set_environment_params_imp (
...
     uint32_t n_reduce_threads = 0,
     int using_gpu = -2,
     float dynamic_range = -1,
     float min_intensity = -1,
     float max_intensity = -1,
     int ram_limit_mb = -1,
-    int verb_level = 0)
+    int verb_level = 0,
+    const std::string& binning_origin = "")
```
vjaganat90, line 160: If you are using default values, I recommend using by copy semantics, const std::string. Move semantics make copy-ing zero-cost.

If it has to be a reference I suggest universal reference const std::string&&, but try simple value semantics.

tests/test_binning_origin.h
```diff
@@ -0,0 +1,149 @@
...
+		int n_bins = 10;
+		PixIntens min_I = 50, max_I = 200, x = 100;
+
+		// slope = 10 / 200 = 0.05, intercept = 1 - 0.05*0 = 1
+		// bin = floor(0.05 * 100 + 1) = floor(6) = 6
+		auto result = TextureFeature::bin_pixel(x, min_I, max_I, n_bins, BinningOrigin::zero);
+		ASSERT_EQ(result, 6);
+	}
+
+	// "min" origin, PyRadiomics-style:
```
vjaganat90, line 27: some of these redundant and noisy comments should be cleaned up.

tests/test_compat_3d_fo_radiomics.h
```diff
@@ -111,7 +111,8 @@ void test_compat_radiomics_3fo_feature (const Nyxus::Feature3D &expected_fcode,
     s[(int)NyxSetting::SOFTNAN].rval = 0.0;
     s[(int)NyxSetting::TINY].rval = 0.0;
     s[(int)NyxSetting::SINGLEROI].bval = false;
-    s[(int)NyxSetting::GREYDEPTH].ival = -20;   // intentionally negative to activate radiomics binCount-based grey-binning
+    s[(int)NyxSetting::GREYDEPTH].ival = 20;
+    s[(int)NyxSetting::BINNING_ORIGIN].ival = static_cast<int>(BinningOrigin::min_based);
```
vjaganat90, line 115: we should use consistent casting, either use C-style (int) or C++-safe static_cast<int> at the same place, not mix-match, I would suggest, convert all these (even where your changes don't touch) to consistent static_cast<int>

src/nyx/environment.cpp
```diff
@@ -915,11 +936,21 @@ int Environment::get_coarse_gray_depth()
 	return coarse_grayscale_depth;
 }
 
-void Environment::set_coarse_gray_depth(unsigned int new_depth)
+void Environment::set_coarse_gray_depth(int new_depth)
```
vjaganat90: again why are we changing the type of depth here? are these unsigned -> signed changes justified/documented?

vjaganat90 (commented):
A few fixes and then it is good.

PR #341, opened 2026-05-07
fix: migrate build_and_test from macos-15-intel to macos-14

vjaganat90:


vjaganat90:
Migrating `build_and_test` from `macos-15-intel` to `macos-14` surfaced a latent 3D shape bug that had been hidden by platform-dependent behavior in the convex-hull geometry path. The immediate issue was a simplex indexing bug that corrupted hull mesh construction, but once that was fixed it became clear that `3AREA` and the related shape metrics were still being derived from convex-hull geometry rather than the ROI’s actual exposed surface. That is why the failure pattern changed across runners: the hull bug was the trigger, but the real semantics bug was in how surface area was defined and computed, and the final fix corrects both layers with an independent SciPy cross-check confirming that convex-hull area and true ROI surface area are materially different quantities for this phantom. CI is now green again, with all 46 checks passing across platforms.

sameeul (approved):

PR #350, opened 2026-06-23
2d audit feature fix 

vjaganat90:
- Fix 2D morphology regressions for equal-area diameter, weighted centroid/mass displacement, compactness distance handling, and Euler number neighborhood counting/padding. 
 
- Refresh the corresponding shape fixture values and independent oracle outputs so `DIAMETER_EQUAL_AREA`, `MASS_DISPLACEMENT`, and `EULER_NUMBER` now reconcile with MATLAB/scikit-image/cp-measure where comparable.


src/nyx/features/ngldm.cpp
```diff
@@ -264,22 +264,30 @@ void NGLDMfeature::calc_features (const std::vector<double>& Sg, const std::vect
...
 		}
 
-	// Calculate features
-	for (int i = 0; i < Ng; ++i)
-	{
-		double sj = 0;
-		for (int j=1; j<Nr; ++j)
-		{
+	// Calculate features
+	for (int i = 0; i < Ng; ++i)
```
darkclad: Why do we have two loops now ?

darkclad (approved):

src/nyx/features/ngldm.cpp
vjaganat90: GLV and DCV need the finalized grey-level and dependence-count means (f_GLCM, f_DCM) before their variance terms can be computed. The first loop accumulates the direct IBSI feature terms and means. The second loop uses those means to compute the variance rows without stale/partial values.

src/nyx/features/basic_morphology.cpp
```diff
@@ -27,28 +27,32 @@ void BasicMorphologyFeatures::calculate (LR& r, const Fsettings& sett)
...
-	val_CENTROID_Y = cen_y;
+	for (auto& px : r.raw_pixels)
+	{
+		cen_x += px.x;
+		cen_y += px.y;
+	}
+
+	cen_x /= n;
+	cen_y /= n;
+	val_CENTROID_X = cen_x;
```
sameeul: `val_CENTROID_X/Y` is computed here (line 44/45) by a local loop, used for `COMPACTNESS`, then reset to 0 and recomputed from scratch by the "Basic morphology :: Centroids" block 20 lines below (the second block also writes `val_CENTROID_X/Y`). The first assignment and its pixel loop are dead — the output is immediately overwritten.

The fix is to drop the first `cen_x/cen_y` loop + assignment and alias from the second block instead:

```cpp
// --COMPACTNESS (must come after the canonical centroid block below)
val_CENTROID_X = val_CENTROID_Y = 0;
for (auto& px : r.raw_pixels) { val_CENTROID_X += px.x; val_CENTROID_Y += px.y; }
val_CENTROID_X /= n;
val_CENTROID_Y /= n;
double cen_x = val_CENTROID_X, cen_y = val_CENTROID_Y;

Moments2 mom2;
for (auto& px : r.raw_pixels) { ... }
val_COMPACTNESS = mom2.std() / n;
```

Same issue exists in `osized_calculate` (two separate centroid passes). Neither is a correctness bug since `cen_x/cen_y` locals are what COMPACTNESS reads, but it wastes two O(n) scans on every ROI.

src/nyx/features/ngldm.cpp
```diff
@@ -264,76 +278,76 @@ void NGLDMfeature::calc_features (const std::vector<double>& Sg, const std::vect
...
-			// Dependence Count Variance
-			//	F_{\mathit{ngl.dc.var}}= \sum_{i=1}^{N_g} \sum_{j=1}^{N_n} (j-\mu)^2 p_{ij}
-			//		where
-			//	\mu = \sum_{i=1}^{N_g} \sum_{j=1}^{N_n} j\,p_{ij}
-			f_DCV += (k - f_DCM) * (k - f_DCM) * pij;	
-		}
-	}
+	// Calculate features. NGLDM column 0 means no matching neighbors, which
+	// corresponds to IBSI dependence count 1 (the center pixel itself).
+	std::vector<double> grey_level_sums(Ng, 0.0);
```
sameeul: `grey_level_sums` and `dependence_count_sums` here are just the marginal sums of the NGLDM — exactly what `calc_rowwise_and_columnwise_totals` already computed as `Sg` and `Sr` (passed in via the function signature). `calc_features` never reads those parameters and recomputes the same values from scratch instead.

This means every call site runs `calc_rowwise_and_columnwise_totals` for nothing. Either use `Sg`/`Sr` directly in place of `grey_level_sums`/`dependence_count_sums`, or drop `Sg`/`Sr` from the signature and remove the now-dead `calc_rowwise_and_columnwise_totals` calls in `calculate()` (line 199) and `osized_calculate()` (line 437).

vjaganat90: Fixed

src/nyx/features/basic_morphology.cpp
vjaganat90: Fixed. No redundant calculations.

PR #351, opened 2026-06-23
feat: add IBSI Intensity Histogram features, per-bin HISTOGRAM output…

darkclad:
…, and --mergerois

Add the 46-feature IBSI Intensity Histogram family (IH_*), computed from a single per-ROI N-bin histogram (N = coarse gray depth): 20 "...Value" features over discretised intensities, 19 "...Index" features over 1-based bin indices, 4 histogram-gradient features, plus Number of Bins and Bin Size.

- New IntensityHistogramFeatures class derives all 46 from one shared histogram; for floating-point images the histogram is built in the original float intensity domain (inverse of the load-time uint quantization), mirroring image_loader.cpp's [fpimgmin,fpimgmax]/dynamic-range choice.
- Gated on IBSI mode: enabled only when --ibsi=true; otherwise stripped (with a warning) at feature-group expansion. New *ALL_IH* group (FG2_IH); selectable individually by IH_* code.
- Expose fp-image quantization params to features via new FPIMG_TARGET_DR / FPIMG_MIN / FPIMG_MAX / FPIMG_ACTIVE settings; resolve IBSI compliance before expand_featuregroups() so the gate sees the final value.

Expose the per-bin intensity histogram as the multi-valued HISTOGRAM feature (HISTOGRAM_BIN_* CSV columns; bin count = coarse gray depth). Opt-in: not in *ALL_INTENSITY*, included by *ALL*.

Add --mergerois (Environment::mergeLabels): collapse all nonzero mask labels into one whole-foreground ROI (background excluded) instead of one ROI per label. Applied at the 2D gather/extract and in-memory label-read sites; distinct from singleROI/wholeslide. Exposed in the Python API; fix initialize_environment to set fp-image options before compile_feature_settings() so the IH float domain uses the requested dynamic_range.

Docs: document the IH family + *ALL_IH* in featurelist.rst and --ibsi / --mergerois in cmdline_and_examples.rst.

Tests: C++ unit tests (tests/test_intensity_histogram.h, 5 cases vs hand-computed ground truth incl. integer/float domains and the IBSI gate) and Python tests (tests/python/test_intensity_histogram.py). Full suite 469/469.


src/nyx/features/histogram.h
```diff
@@ -261,6 +261,30 @@ class TrivialHistogram
...
+
+	// Per-bin frequencies of the custom-resolution histogram, trimmed to exactly
+	// 'n_cust_bins' bins (the internal vector carries one extra, folded, slot).
+	std::vector<double> get_cust_frequencies (int n_cust_bins) const
+	{
+		int n = std::abs(n_cust_bins);
+		std::vector<double> v;
+		v.reserve(n);
+		for (int k = 0; k < n; k++)
+			v.push_back (k < (int)bins_cust_.size() ? double(bins_cust_[k]) : 0.0);
```
vjaganat90: use explicit casts static_cast<int> instead of C-style casts.

src/nyx/features/intensity_histogram.h
```diff
@@ -0,0 +1,115 @@
...
+//
+// Availability is gated on IBSI mode: the family is only enabled when IBSI
+// compliance is on (see Environment::expand_featuregroups()). A defensive
+// compute-time guard fills NaN if invoked with IBSI off.
+
+class IntensityHistogramFeatures : public FeatureMethod
+{
+public:
+
+	const constexpr static std::initializer_list<Nyxus::Feature2D> featureset =
```
vjaganat90, line 27: const is redundant here. Also why not std::array if everything is known at compile time?

src/nyx/env_features.cpp
```diff
@@ -501,6 +509,23 @@ void Environment::expand_featuregroups()
 			continue;
 		}
 	}
+
+	// IBSI gate: the Intensity-Histogram family (46 features) is available ONLY when
+	// IBSI mode is on. If IBSI is off, strip any IH features the user requested
+	// (individually or via *ALL_IH* / *ALL*).
+	if (dim() == 2 && !ibsi_compliance)
+	{
+		bool anyIH = false;
```
vjaganat90: This could be more clear and obvious.

// C++20 version, Nyxus supports
bool anyIH = std::ranges::any_of(IntensityHistogramFeatures::featureset, 
                                 [&theFeatureSet](auto fc) { 
                                     return theFeatureSet.isEnabled(fc); 
                                 });

src/nyx/env_features.cpp
```diff
@@ -699,6 +724,10 @@ void Environment::compile_feature_settings()
 			s[(int)NyxSetting::USEGPU].bval = using_gpu();
 			s[(int)NyxSetting::VERBOSLVL].ival = get_verbosity_level();
 			s[(int)NyxSetting::IBSI].bval = ibsi_compliance;
+			s[(int)NyxSetting::FPIMG_TARGET_DR].rval = fpimageOptions.target_dyn_range();
```
vjaganat90, line 726: I think we should move away from C-style casts everywhere we touch incrementally.

vjaganat90 (changes requested):

src/nyx/features/intensity_histogram.h
darkclad: That is how it is done everywhere in the code, I agree that we can change, but it will standout

src/nyx/env_features.cpp
darkclad: Matches current code style, if restyle is on the table it shall be done in separate PR

src/nyx/features/histogram.h
darkclad: Fixed

src/nyx/env_features.cpp
darkclad: Fixed

vjaganat90 (approved):
LGTM

PR #352, opened 2026-06-26
fix: categorize and name feature unit tests and expected values

vjaganat90:
- This is a major naming and structure refactor that resolves source of feature V&V at the unit test level if it exists. Clearly re-names internal calculation unit tests.

- Added a zero-dependency Python reproduction script in `test_data.h` that regenerates the embedded C++ IBSI phantom image/mask arrays directly from the pinned IBSI standard NIfTI files.  This fixes the prior one-voxel mismatch so the embedded fixture now matches the IBSI standard data exactly.  

- The corrected voxel is outside the active mask/filter used by the current tests, so this does not change existing test results.  The change improves fixture provenance and makes the IBSI reference data easy to audit and reproduce.

- Reconciled IBSI-facing fixtures and comments so the standard phantom source, pinned data repo commit/URL, and regeneration path are explicit and reproducible.

- Updated NGLDM to match IBSI matrix/feature conventions where IBSI defines the feature, while leaving only non-IBSI rows clearly marked as unvetted local regression.

- Addressed review clarity issues by passing review prefixes through shared helpers and documenting why remaining local invariants are not third-party oracle checks.


sameeul (commented):
Code review findings from Sameeul — 4 inline comments covering the issues flagged in the review.

tests/test_ibsi_ngldm.h
```diff
@@ -175,8 +175,8 @@ void test_ibsi_ngldm_feature (const Feature2D& feature_, const std::string& feat
 
     int feature = int(feature_);
 
-    // Check if ground truth is available for the feature
-    ASSERT_TRUE(ngldm_regression_gtruth.count(feature_name) > 0);
+    SCOPED_TRACE(std::string("UNVETTED_NO_DIRECT_ORACLE__") + feature_name);
+    ASSERT_TRUE(unvetted_nyxus_regression_ngldm_feature_golden_values.count(feature_name) > 0);
```
sameeul: **Issue (Important): Incorrect SCOPED_TRACE prefix in IBSI-verified NGLDM tests**

`SCOPED_TRACE(std::string("UNVETTED_NO_DIRECT_ORACLE__") + feature_name)` is now hardcoded inside `test_ibsi_ngldm_feature()`, but this same function is called by IBSI-verified tests (e.g. `test_ibsi_NGLDM_LDE`). When those tests fail, GTest will report the failure as `UNVETTED_NO_DIRECT_ORACLE__NGLDM_LDE` — the opposite of what it is.

Suggest mirroring the pattern used in `assert_2d_geomoment_features()`: pass a `review_prefix` string parameter to `test_ibsi_ngldm_feature()` and let each call site supply `"VERIFIABLE_WITH_3P_BUILTIN_ORACLE__"` or `"UNVETTED_NO_DIRECT_ORACLE__"` as appropriate.

tests/test_ibsi_gldzm.h
```diff
@@ -25,12 +25,16 @@ static std::unordered_map<std::string, double> ibsi_gldzm_gtruth
...
     {"GLDZM_ZP",        0.24},  // Zone percentage
-    {"GLDZM_GLM",       3.476190476190476}, // Grey level mean
     {"GLDZM_GLV",       3.97},  // Grey level variance
-    {"GLDZM_ZDM",       1.1071428571428572}, // Zone distance mean
     {"GLDZM_ZDV",       0.0816326530612245}, // Zone distance variance
     {"GLDZM_ZDE",       1.73}   // Zone distance entropy
 };
+
+static std::unordered_map<std::string, double> unvetted_nyxus_regression_gldzm_feature_golden_values
+{
```
sameeul: **Issue (Important): GLM and ZDM moved to unvetted table without explanation**

`GLDZM_GLM` and `GLDZM_ZDM` were in the IBSI reference table and are now in `unvetted_nyxus_regression_gldzm_feature_golden_values`. The IBSI standard does define Grey Level Mean and Zone Distance Mean, so the reason they cannot be verified against the IBSI phantom values (e.g. implementation difference, aggregation method mismatch) should be documented in a comment here so future maintainers don't inadvertently promote them back to oracle-verified.

tests/test_shape_morphology_2d.h
```diff
@@ -21,7 +21,7 @@
 #include "test_data.h"
 #include "test_main_nyxus.h"
 
-static std::unordered_map<std::string, double> shape2d_truth{
+static std::unordered_map<std::string, double> unvetted_nyxus_regression_shape2d_feature_golden_values{
 	{"AREA_PIXELS_COUNT", 26.0},
```
sameeul, line 33: **Issue (Minor): Geometrically trivial features classified as unvetted — rationale missing**

Features like `AREA_PIXELS_COUNT`, `CENTROID_X/Y`, `BBOX_*` are placed in `unvetted_nyxus_regression_shape2d_feature_golden_values`, which signals "no external oracle exists". But for a known fixture these values are trivially hand-verifiable (26 pixels, bounding box of a specific shape, etc.). If the intent is that the *test fixture itself* is not IBSI-standardized (which is valid), a brief comment explaining that distinction would prevent future readers from mistakenly treating these as unvalidated implementations.

tests/test_ibsi_gldzm.h
vjaganat90: IBSI values reconciled without changing formula. No accidental drift due to clear documentation now.

PR #354, opened 2026-07-01
fix(gldm): exclude background from the dependence matrix

darkclad:
The MATLAB grey-binning path maps off-ROI background (0) to level 1, so the GLDM zone loop's `pi == 0` guard (which tests the BINNED value) never rejected background pixels inside the bounding box. They were counted both as zones and as dependent neighbours, inflating Nz (154 -> 234 on the canonical ROI) and every count/dependence feature.

Skip background by the ORIGINAL intensity (imR / raw cloud value == 0) and require a neighbour to be an ROI pixel before counting it as a dependency, in both the trivial and out-of-core paths. Mirrors the GLCM background fix.

Adds tests/python/test_feature_bugs.py::test_gldm_background_not_counted, checking LDE/GLN/DN against the PyRadiomics reference on the canonical ROI.

### DIFF
--- a/tests/python/test_feature_bugs.py
+++ b/tests/python/test_feature_bugs.py
@@ -1,0 +1,70 @@
+"""Regression / bug-exposure tests for 2D feature defects found during oracle validation
+(2026-06). These exercise the PRODUCTION featurize() path on ROIs *with background* and at
+the DEFAULT settings - the conditions the C++ unit tests miss.
+
+This module covers the GLDM background-pollution defect (bug #14b).
+"""
+import re
+from pathlib import Path
+import numpy as np
+import pytest
+import nyxus
+
+
+# ----------------------------- helpers --------------------------------------
+def _canonical_roi():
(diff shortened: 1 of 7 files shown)


tests/python/test_feature_bugs.py
```diff
@@ -0,0 +1,70 @@
+"""Regression / bug-exposure tests for 2D feature defects found during oracle validation
+(2026-06). These exercise the PRODUCTION featurize() path on ROIs *with background* and at
+the DEFAULT settings - the conditions the C++ unit tests miss.
+
+This module covers the GLDM background-pollution defect (bug #14b).
```
sameeul: this comment and the file name need to be updated to reflect the test. Saying it `test_feature_bugs.py` will become misleading in the long run. Also, do we have an equivalent GTest?

darkclad: Fixed

PR #357, opened 2026-07-01
fix(convex-hull): Pick's-theorem hull area so SOLIDITY <= 1

darkclad:
ConvexHullFeature computed the hull area with the bare shoelace polygon area, which runs through pixel centres and under-counts the rasterised pixel coverage. The ROI area is a pixel count, so SOLIDITY = roi_area / hull_area could exceed 1 (1.3 on the canonical ROI) - impossible.

Use the Pick's-theorem pixel-count-equivalent hull area (shoelace + boundary_points/2 + 1, boundary_points = sum of gcd(|dx|,|dy|) over edges), matching scikit-image convex_area, plus a hull>0 guard. SOLIDITY is now <= 1 in both the trivial and out-of-core paths.

Refresh the vetted goldens in test_shape_morphology_2d.h (CONVEX_HULL_AREA 20->27, SOLIDITY 1.3->0.96296 = 26/27; CIRCULARITY and all other shape goldens unchanged) and add tests/python/test_feature_bugs.py::test_solidity_le_one.

### DIFF
--- a/src/nyx/features/convex_hull_nontriv.cpp
+++ b/src/nyx/features/convex_hull_nontriv.cpp
@@ -1,6 +1,7 @@
 #define _USE_MATH_DEFINES	// For M_PI, etc.
 #include <algorithm>
 #include <cmath>
+#include <numeric>
 #include "../dataset.h"
 #include "../feature_method.h"
 #include "../feature_settings.h"
@@ -8,6 +9,28 @@
 #include "convex_hull.h"
 
 using namespace Nyxus;
+
+// Number of integer lattice points lying ON the hull boundary = sum over edges of gcd(|dx|,|dy|).
+// By Pick's theorem the pixel-count-equivalent hull area (matching scikit-image convex_area, which
+// counts the pixels of the rasterised hull) is shoelace_area + boundary_points/2 + 1. The bare
+// shoelace area runs through pixel CENTRES and under-counts coverage, so for small/elongated ROIs
+// it falls below the ROI pixel count -> SOLIDITY > 1 (impossible). This correction fixes that.
+static long hull_boundary_points (const std::vector<Pixel2>& v)
+{
+	size_t n = v.size();
+	if (n < 2)
+		return 0;
+	long B = 0;
+	for (size_t i = 0; i < n; i++)
+	{
+		const Pixel2& p1 = v[i];
+		const Pixel2& p2 = v[(i + 1) % n];
+		long dx = (long)p2.x - (long)p1.x; if (dx < 0) dx = -dx;
+		long dy = (long)p2.y - (long)p1.y; if (dy < 0) dy = -dy;
+		B += std::gcd(dx, dy);
+	}
+	return B;
+}
 
 ConvexHullFeature::ConvexHullFeature() : FeatureMethod("ConvexHullFeature")
 {
@@ -32,12 +55,13 @@ void ConvexHullFeature::calculate (LR& r, const Fsettings& s)
 	// Build the convex hull
 	build_convex_hull(r.raw_pixels, r.convHull_CH);
 
-	// Calculate related features
-	double s_hull = polygon_area(r.convHull_CH),
-		s_roi = r.raw_pixels.size(), 
+	// Calculate related features. Pixel-count-equivalent hull area (Pick's theorem) so the hull is
+	// measured on the same basis as the ROI (pixel count) -> SOLIDITY <= 1.
+	double s_hull = polygon_area(r.convHull_CH) + hull_boundary_points(r.convHull_CH) / 2.0 + 1.0,
+		s_roi = r.raw_pixels.size(),
 		p = r.fvals[(int)Feature2D::PERIMETER][0];
 	area = s_hull;
-	solidity = s_roi / s_hull;
+	solidity = (s_hull > 0.0) ? s_roi / s_hull : 0.0;
 	circularity = sqrt(4.0 * M_PI * s_roi / (p*p));
 }
 
@@ -193,12 +217,12 @@ void ConvexHullFeature::osized_calculate (LR& r, const Fsettings& s, ImageLoader
 	// Build the convex hull
 	build_convex_hull (r.raw_pixels_NT, r.convHull_CH);
 
-	// Calculate related features
-	double s_hull = polygon_area(r.convHull_CH),
+	// Calculate related features (Pick's-theorem pixel-count-equivalent hull area; see calculate())
+	double s_hull = polygon_area(r.convHull_CH) + hull_boundary_points(r.convHull_CH) / 2.0 + 1.0,
 		s_roi = r.raw_pixels_NT.size(),
 		p = r.fvals[(int)Feature2D::PERIMETER][0];
 	area = s_hull;
-	solidity = s_roi / s_hull;
+	solidity = (s_hull > 0.0) ? s_roi / s_hull : 0.0;
 	circularity = sqrt(4.0 * M_PI * s_roi / (p * p));
 }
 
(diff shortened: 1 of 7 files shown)


darkclad:
# Justification — golden value changes for `main-convex-hull-solidity`

Branch: `main-convex-hull-solidity` (split from `main-feature-validation`, based on `main`)
Source fix: `src/nyx/features/convex_hull_nontriv.cpp`
Test file touched: `tests/test_shape_morphology_2d.h` (2 goldens: `CONVEX_HULL_AREA`, `SOLIDITY`)

## The bug

`ConvexHullFeature::calculate` computed the hull area with the bare shoelace polygon area
(`polygon_area(convHull)`), which runs through pixel **centres** and therefore under-counts the
rasterised pixel coverage of the hull. The ROI area (`s_roi`) is a **pixel count**. Dividing a
pixel-count ROI area by a centre-based hull area gives `SOLIDITY = s_roi / s_hull > 1` for
small/elongated ROIs — impossible (solidity is by definition ≤ 1). On the canonical fixture the old
golden was `SOLIDITY = 1.3`.

## The fix

Use the **Pick's-theorem pixel-count-equivalent** hull area so the hull is measured on the same basis
as the ROI (pixel count):

```
s_hull = shoelace_area + boundary_points/2 + 1      // boundary_points = sum over edges of gcd(|dx|,|dy|)
```

This matches scikit-image's `convex_area` (the pixel count of the rasterised hull). Result:
`SOLIDITY = s_roi / s_hull ≤ 1`, plus a `s_hull > 0` guard.

## Golden changes (2 keys)

| Golden               | Old  | New                          | Why                                                                                                                                                                    |
| -------------------- | ---- | ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `CONVEX_HULL_AREA` | 20.0 | **27.0**               | Pick's-theorem pixel-count hull area on the fixture. External oracle scikit-image`convex_area = 28` (≈4% high; the two hulls differ by one boundary lattice point). |
| `SOLIDITY`         | 1.3  | **0.9629629629629629** | `= AREA_PIXELS_COUNT / CONVEX_HULL_AREA = 26 / 27 = 0.962962…`. scikit-image gives `26/28 = 0.929`. Now ≤ 1.                                                     |

## Why this is not fudging

- **Internal consistency:** the new `SOLIDITY` is *exactly* `AREA_PIXELS_COUNT (26) / CONVEX_HULL_AREA (27)`; both the numerator (26, unchanged golden) and the new denominator are in the same table.
- **Nothing else moved:** the fix only changes `s_hull`. `CIRCULARITY` (which uses `s_roi` and the
  perimeter, not the hull) is **unchanged**, as are all other shape goldens. Only the 2 hull-derived
  keys changed.
- **Sign of correctness:** the old value 1.3 was provably wrong (solidity can never exceed 1); the new
  value is ≤ 1 and within ~4% of scikit-image's independent `convex_area`/`solidity`.
- Values are byte-identical to the oracle-validated `main-feature-validation` (copied, not invented);
  only the shared table-merge refactor was dropped (I kept `main`'s two-table structure and edited the
  two numbers in place).

## Regression guard

`tests/python/test_feature_bugs.py::test_solidity_le_one` asserts `SOLIDITY <= 1 + 1e-6` on the
canonical ROI through the production `featurize()` path — the hard correctness invariant that the old
code violated (1.3).

## CI

sameeul:
## SOLIDITY fix looks correct — one follow-up on the shared hull area

Fix is right: Pick's-theorem pixel-count area (`polygon_area + B/2 + 1`) makes SOLIDITY ≤ 1 provable (ROI pixels ⊆ hull lattice points → `s_hull ≥ s_roi`), both paths updated, degenerate cases fine. Keeping Pick's theorem is the right choice — deterministic, `O(#vertices)`, allocation-free, and it guarantees the ≤1 invariant.

Nit: `(s_hull > 0.0) ? ... : 0.0` is now dead (`s_hull ≥ 1` after `+1`). Harmless.

**The one thing to handle:** `CONVEX_HULL_AREA` is also read by `hexagonality_polygonality.cpp:37`, so this shifts `HEXAGONALITY_AVE`/`POLYGONALITY_AVE`/`HEXAGONALITY_STDDEV` too. Their goldens in `test_2d_remaining_features.h` (L62–64) aren't updated, and those keys are asserted bounds-only (`>0`, `≤10`, L259/268) — so CI passes while the values silently go stale.

This shift is *correct*, not a regression: the Polus reference computes `area_hull = area/solidity` = skimage `convex_area` (pixel count), so these features were always meant to use a pixel-count hull area. The fix moves them toward the reference and correctly keeps all three on one shared area — don't split.

**Asks:**
1. Refresh the three hexagonality/polygonality goldens in this PR **and switch their assertions from the bounds-only helper to a value check with tolerance**. Right now `assert_..._polygonality_score`/`_feature` (L252–268) only test `>0`/`≤10`, so the golden values on L62–64 are never actually compared — that's why the shift is silent. Point these keys at the value-comparison path (`agrees_gt(actual, golden, tol)`, as the other remaining2d features already use) against the refreshed values so real drift is caught going forward.
2. For `CONVEX_HULL_AREA`/`SOLIDITY` (skimage computes these directly, unlike the Polus-only hexagonality/polygonality scores): run skimage `regionprops` locally on the canonical ROI, set the goldens from skimage's `convex_area`/`solidity`, and assert nyxus within a tolerance that covers the Pick's-vs-rasterization gap (~1px / few %; 27 vs 28 today). Makes it a real skimage-vetted tolerance check instead of a self-referential snapshot; document the boundary-convention gap as the reason for the tolerance.

Minor: `test_convex_hull_oracle.py` asserts an invariant, not an oracle match — the `_oracle` name slightly overclaims. Test itself is good.

Verdict: take the fix; just refresh the hexagonality/polygonality goldens and add the tolerance cross-check in the same PR.

darkclad:
All three points addressed in this PR (tests only, no change to the fix itself). Full suites green after: C++ 683 passed / 0 failed, Python 49 passed / 1 skipped.

**1. Hexagonality/polygonality goldens + assertions.** These were going stale silently. Refreshed and pointed at the value-comparison path (`agrees_gt(actual, golden, tol)`) so the bounds-only helper no longer hides drift:

- `HEXAGONALITY_AVE` 6.4263 → 6.8823, `HEXAGONALITY_STDDEV` 0.3144 → 0.1850.
- `POLYGONALITY_AVE` is **unchanged** (2.0833): it depends only on neighbors/area/perimeter, not on `area_hull`, so only the two HEXAGONALITY keys move with the shared hull area. All three stay on the one shared `CONVEX_HULL_AREA` — no split.
- The `<=10` score bound is retained as a cheap secondary invariant on top of the value check.

**2. CONVEX_HULL_AREA / SOLIDITY vs skimage.** Moved both from the self-referential snapshot into the 3p-oracle table, goldens set from `skimage.regionprops` (`area_convex=28`, `solidity=0.9286`), asserted within a 10% tolerance that covers the gap. Characterised the gap so it's a documented convention, not a fudge factor: Nyxus hulls through pixel **centres** and reproduces `skimage.convex_hull_image(offset_coordinates=False) == 27` **exactly**; skimage's regionprops default (`offset_coordinates=True`) expands each pixel to its ±0.5 corners → 28. So SOLIDITY is now a real skimage-vetted `<=1` check.

**3. `_oracle` filename.** Renamed `test_convex_hull_oracle.py` → `test_convex_hull_invariants.py` — it asserts the `SOLIDITY <= 1` invariant, not an oracle match. Docstring updated to say so and to point at the C++ skimage cross-check for the value comparison.

Left the now-dead `(s_hull > 0.0) ? … : 0.0` guard as-is since it's harmless and outside these asks; can drop it in a follow-up if preferred.

PR #359, opened 2026-07-01
fix(neighbors): implement closest-neighbor dist/ang and cap PERCENT_T…

darkclad:
…OUCHING

Two production defects:

- CLOSEST_NEIGHBOR{1,2}_DIST/ANG and ANG_BW_NEIGHBORS_* need CENTROID_X/Y from Basic Morphology. When only neighbor features were requested, Basic Morphology was skipped in the trivial-ROI reduce, so the centroids stayed 0 and the closest-neighbor distances/angles came out 0. Run Basic Morphology whenever neighbor features are requested (reduce_trivial_rois.cpp).
- PERCENT_TOUCHING summed a per-neighbor touch count, so a contour pixel adjacent to several neighbors was counted repeatedly and the total could exceed 100% (e.g. 127%). Build a deduped per-ROI mask of contour pixels touching ANY neighbor (8-connected adjacency) and divide by contour length -> <= 100% (neighbors.cpp).

Refresh the one affected vetted golden in test_neighbors_2d.h (label 2 PERCENT_TOUCHING 33.33 -> 66.67; all closest-neighbor goldens unchanged) and add tests/python/test_feature_bugs.py::test_neighbor_distance_and_percent_touching.


darkclad:
# Justification — golden value change for `main-neighbors-closest-and-percent-touching`

Branch: `main-neighbors-closest-and-percent-touching` (split from `main-feature-validation`, based on `main`)
Source fix: `src/nyx/features/neighbors.cpp`, `src/nyx/reduce_trivial_rois.cpp`
Test file touched: `tests/test_neighbors_2d.h` (1 golden: label 2 `PERCENT_TOUCHING`)

## The bugs

- **#13 PERCENT_TOUCHING could exceed 100%** (`neighbors.cpp`). The old code added a per-neighbor
  touch count, so a contour pixel adjacent to several neighbours was counted once per neighbour and
  the total could exceed the contour length (e.g. 127%). Fix: build a **deduped** per-ROI mask of
  contour pixels touching ANY neighbour (8-connected adjacency, sqdist ≤ 2) and divide the distinct
  touching-pixel count by the contour length → guaranteed ≤ 100%.
- **#12 CLOSEST_NEIGHBOR{1,2}_DIST/ANG were 0 in production** (`reduce_trivial_rois.cpp`). Those
  features need `CENTROID_X/Y`, which are produced by Basic Morphology. When a user requested *only*
  neighbour features, Basic Morphology was not run, so the centroids stayed 0 and the closest-neighbour
  distances/angles came out 0. Fix: run Basic Morphology whenever neighbour features are requested.

## Golden change (1 key)

| Label | Golden               | Old      | New                | Why                                                                  |
| ----- | -------------------- | -------- | ------------------ | -------------------------------------------------------------------- |
| 2     | `PERCENT_TOUCHING` | 33.333… | **66.666…** | The deduped, adjacency-based touching fraction for this fixture ROI. |

Only this one value changed. Every other neighbour golden (`NUM_NEIGHBORS`, all
`CLOSEST_NEIGHBOR*_DIST/ANG`, `ANG_BW_NEIGHBORS_*`, and the other labels' `PERCENT_TOUCHING`) is
**unchanged** — the C++ unit test calls Basic Morphology explicitly before the neighbour reduce, so
its centroids are always populated and bug #12 (a production-only reduce-ordering defect) does not
manifest in the C++ fixture. That the whole suite passes with only label 2's `PERCENT_TOUCHING`
updated confirms the closest-neighbour reimplementation is value-preserving on this fixture.

## Why this is not fudging

- **Bounded invariant:** the fix's purpose is `PERCENT_TOUCHING <= 100%`; label 2 stays well within
  bounds. The old summing scheme is the one that produced impossible >100% values (exercised by the
  Python test on touching squares).
- **Minimal blast radius:** exactly one golden moved; all others are byte-identical to `main`.
- Value is byte-identical to the oracle-validated `main-feature-validation` (copied, not invented);
  only the shared two-table→one-table refactor was dropped.

## Regression guards (Python, production path)

`tests/python/test_feature_bugs.py::test_neighbor_distance_and_percent_touching` on two 4×4 squares
2 px apart asserts, through the production `featurize()` path:

- `CLOSEST_NEIGHBOR1_DIST > 0` for all ROIs — catches bug #12 (was 0 when centroids weren't computed).
- `PERCENT_TOUCHING <= 100` and `== 0` (a 2 px gap → neighbours but not touching) — catches bug #13.

## CI

sameeul:
Thanks @darkclad — the two intended fixes here are correct and I verified both: forcing Basic Morphology so `CENTROID_X/Y` exist (closest-neighbor dist/ang no longer read 0), and the deduped touch mask capping `PERCENT_TOUCHING ≤ 100%`.

While validating the neighbor family against external oracles (CellProfiler `MeasureObjectNeighbors` + a scipy analytic reference), I found a **remaining `PERCENT_TOUCHING` accuracy bug** that this PR *exposes* rather than introduces.

### Repro — two geometrically identical adjacent 4×4 squares

| | Nyxus (this PR) | scipy analytic | CellProfiler "Adjacent" |
|---|---|---|---|
| L1 (contact on its **right**) | **25.0%** | 33.33% | 33.33% |
| L2 (contact on its **left**) | 33.33% | 33.33% | 33.33% |

The error is position-dependent: a contact on the object's **right** edge counts 3/12 where it should count 4/12. Three collinear squares `1｜2｜3` give `{25.0, 58.33, 33.33}` vs the correct `{33.33, 66.67, 33.33}` (scipy + CellProfiler agree).

### Root cause

`Pixel2::min_sqdist` v2 (`src/nyx/features/pixel.cpp:36`) is an *approximate* nearest-distance search — a coarse unimodal hill-descent (`step = (b-a)/log(b-a)`, jump to best sample, narrow, repeat). Distance to a **closed contour is multi-modal**, so it can settle in the wrong basin and **overestimate** the true minimum. The exact full scan (v1) is disabled behind `#if 0`.

This PR tightens the touch test to `sqdist ≤ 2`, which is below v2's typical error, so the latent flaw surfaces: for the left square's pixel `(2,3)`, the true nearest right-square contour pixel is at sqdist **1**, but v2 returns **5** → `5 > 2` → the pixel is silently not marked touching → undercount by one.

Only `PERCENT_TOUCHING` is affected: `CLOSEST_NEIGHBOR*_DIST` uses centroid distance, and `NUM_NEIGHBORS` / the old within-radius touch used the loose `≤ radius² = 25` threshold that tolerated the error.

### Suggested fix

Use an **exact** nearest distance for the touch test (contours are tiny; cost is negligible) — re-enable v1 for this loop, add an `exact_min_sqdist`, or a short-circuit "any neighbor pixel within sqdist ≤ 2" scan. After fixing, the `test_neighbors_2d.h` golden (label 2 = `66.67`) should be **re-derived from the exact/oracle value** — it was measured with this bug present and likely encodes the undercount.

Broader flag: `min_sqdist` v2 is also used by `ROI_RADIUS`, `fractal_dim`, `circle`. Those rely on looser/aggregate distances so are probably fine, but any *exact*-contour-min-distance dependency carries the same latent risk — worth a follow-up audit.

### Self-contained reproducer (reproduces 4 vs 3 exactly)

```python
import math
def sqd(p, q): return (p[0]-q[0])**2 + (p[1]-q[1])**2
def exact(p, cloud): return min(sqd(p, c) for c in cloud)
def v2(p, cloud):                       # faithful replica of Pixel2::min_sqdist v2
    n = len(cloud); a, b = 0, n; ed = sqd(p, cloud[a]); ei = a
    step = int((b-a)/math.log(b-a))
    while True:
        i = a + step
        while i < b:
            d = sqd(p, cloud[i])
            if ed > d: ed = d; ei = i
            i += step
        stepL = step if ei >= step else ei
        stepR = step if ei+step < n else n-ei
        a = ei-stepL; b = ei+stepR
        step = 1 if (b-a) <= 10 else int((b-a)/math.log(b-a))
        if not (b-a > 2): break
    return ed
def trace(r0, c0):                      # clockwise boundary of a filled 4x4 square (12 px)
    return [(r0,c0),(r0,c0+1),(r0,c0+2),(r0,c0+3),(r0+1,c0+3),(r0+2,c0+3),
            (r0+3,c0+3),(r0+3,c0+2),(r0+3,c0+1),(r0+3,c0),(r0+2,c0),(r0+1,c0)]
A = trace(0, 0); B = trace(0, 4)        # adjacent squares
# failing pixel A(2,3): exact sqdist 1, v2 sqdist 5 -> missed
print("A(2,3):", "exact", exact((2,3), B), "v2", v2((2,3), B))
```

(Reproducers + the CellProfiler `MeasureObjectNeighbors` verification setup available on request.)

PR #367, opened 2026-07-11
test(intensity-histogram): vet the 21 untested IBSI IH dispersion/index features

sameeul:
## What

Adds `test_ih_dispersion_and_index_values` — value-check coverage for the **21 IBSI
Intensity-Histogram (IH) features that previously had no assertion at all**:

- **7 "…Value" dispersion features:** mean-absolute-deviation, robust-MAD, median-AD,
  interquantile range, coefficient of variation, quantile coefficient of dispersion, robust mean.
- **14 "…Index" variants:** variance / skewness / excess-kurtosis / interquantile-range /
  range / MAD / robust-MAD / median-AD / CoV / QCoD / entropy / uniformity / robust-mean, over
  the 1-based bin-index domain.

Test-only — no production code changes.

## Why

A test-suite vetting audit found that `test_ih_integer_domain_values` covered 26 of the 47
IH features, but these 21 dispersion / index features were never checked. A regression in any
of them (or in the robust-window trimming) would have gone undetected.

## How the ground truth is trustworthy

Ground truth is derived **independently from the IBSI first-order definitions** applied to the
discretised N-bin histogram — it is *not* copied from `intensity_histogram.cpp`. So a formula
regression fails the test rather than being silently pinned. For the "…Value" dispersion
features this is an oracle-grade check against the IBSI intensity-histogram definitions; the
"…Index" variants are the same statistics over the bin-index domain (Nyxus-specific) and are
pinned analytically.

Two ROIs are used so the robust window is exercised both degenerately and non-trivially:

| ROI | pixels | N | freq | robust window `[p10Idx, p90Idx]` |
|-----|--------|---|------|----------------------------------|
| 1 (existing 5-px fixture) | `{1,1,3,5,7}` | 3 | `{2,1,2}` | covers all bins → robust == full |
| 2 (new 17-px fixture) | `{0, 10×5, 20×6, 30×4, 40}` | 5 | `{1,5,6,4,1}` | trims both tail bins → robust ≠ full |

ROI-2 is specifically constructed so `robustMean`/`robust-MAD` differ from their full-range
counterparts, giving the trimming path real coverage.

## Verification

Ubuntu C++ GoogleTest CI is green:

```
[ RUN      ] TEST_NYXUS.TEST_IH_DISPERSION_AND_INDEX_VALUES
[       OK ] TEST_NYXUS.TEST_IH_DISPERSION_AND_INDEX_VALUES (0 ms)
...
[  PASSED  ] 696 tests.
```

The pass confirms nyxus's IH dispersion/index implementation agrees with the independently
derived IBSI definitions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/test_all.cc
+++ b/tests/test_all.cc
@@ -773,6 +773,11 @@ TEST(TEST_NYXUS, TEST_IH_REQUIRED_PREDICATE)
 	ASSERT_NO_THROW(test_ih_required_predicate());
 }
 
+TEST(TEST_NYXUS, TEST_IH_DISPERSION_AND_INDEX_VALUES)
+{
+	ASSERT_NO_THROW(test_ih_dispersion_and_index_values());
+}
+
 TEST(TEST_NYXUS, TEST_PIXEL_INTENSITY_VARIANCE_BIASED)
 {
 	ASSERT_NO_THROW(test_pixel_intensity_variance_biased());
--- a/tests/test_intensity_histogram.h
+++ b/tests/test_intensity_histogram.h
@@ -45,6 +45,19 @@ static const NyxusPixel intensityHistogramTestData[] =
     {0, 0, 1}, {1, 0, 1}, {2, 0, 3}, {3, 0, 5}, {4, 0, 7}
 };
 
+// Second ROI for the dispersion / robust family. 17 pixels, N=5 bins over [0,40]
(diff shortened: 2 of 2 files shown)


darkclad (approved):

sameeul:
closed in favor of #374

PR #369, opened 2026-07-13
`feat: --preserve-hu` — preserve CT / Hounsfield-Unit intensities

darkclad:


darkclad:
## Motivation

Nyxus stores every pixel as `PixIntens = unsigned int`. For CT images that silently
corrupts the data on load:

- **Signed-int16 TIFF / DICOM** are cast straight to `unsigned int`, so a HU of −1000
  wraps to ~4.29×10⁹. CT is effectively broken today.
- **Float/real images** are hard min–max rescaled into `[0, 10000]`, destroying the
  absolute HU scale.
- **DICOM `RescaleSlope` / `RescaleIntercept` were never applied**, so even the stored
  values weren't true HU.

This PR adds an opt-in mode that keeps Hounsfield values intact.

## Approach (offset-preserving, `unsigned-int`-friendly)

When `--preserve-hu` / `preserve_hu=True` is set, the load-time quantization is replaced
by a **slope-1 integer offset** instead of a min–max rescale:

```
u = round(HU − floor(global_min_HU))
```

So **1 HU = 1 grey level** (e.g. min → 0, water → +offset, bone → full span), negative CT
values no longer wrap, and the mapping is exactly invertible by adding `floor(min)` back.
This stays within the existing `unsigned int` pipeline — no rewrite of features / texture
matrices / GPU kernels — and reuses the `fp_phys_pivoxels` / `float_domain_map` scaffolding
that already existed. (This is the pragmatic "interpretation B"; a full float pipeline was
considered and rejected as disproportionately expensive.)

## What's included

- **Core primitive:** `SlideProps::uint_friendly_inten` HU branch; `IntensityHistogramFeatures::float_domain_map` reconstructs **true HU** for IH features.
- **CLI:** `--preserve-hu` → `FpImageOptions::preserve_hu()` → `Environment`, propagated onto every `SlideProps` (2D/3D, whole/segmented, montage).
- **Loaders:**
  - Float TIFF — shared `map_real_intensity` helper.
  - Signed-int16 TIFF — wraparound fix via the same `hu_offset` helper.
  - **DICOM** — reads `RescaleSlope`/`RescaleIntercept`, computes true HU, then offsets; scan keeps a true-HU shadow buffer so slide min/max live in the HU domain.
  - **NIfTI (3D)** — applies `scl_slope`/`scl_inter` then offsets.
- **MIN/MAX consistency:** `workflow_2d_whole` now applies the same offset (previously cast raw negative HU, wrapping the whole-slide MIN/MAX).
- **Python:** `preserve_hu=` kwarg on `Nyxus` / `Nyxus3D` / `ImageQuality` → `set_preserve_hu`.
- **Docs:** CLI reference + supported-formats notes.

## Supported range

- Negative HU fully supported (air ≈ −1000 and below); each image's min is rebased to 0.
- No fixed upper HU ceiling — the only limit is that the per-image span `max − floor(min)`
  fit in `unsigned int` (~4.29×10⁹ levels), which covers all real CT including dense metal.
- 1 HU = 1 grey level, integer-exact; sub-unit fractional precision is rounded (CT HU is
  integer-valued, so no loss).

## Testing

Runs in the existing CI jobs (no new CI wiring — tests live where CI already looks):

- **C++ GTest (`runAllTests`, `ALLEXTRAS=ON`): full suite green (700/700).** 14 dedicated HU
  tests: offset primitive, CLI parse, int16/float TIFF loaders, and **4 real-dcmtk DICOM
  decodes** including pydicom's real `CT_small.dcm` with independently-computed reference
  values. DICOM tests run on Ubuntu + Windows (Mac env excludes dcmtk).
- **PyTest (`pytest tests/python/`): green (69 passed, 1 pre-existing skip).** 3× 2D TIFF via
  `featurize_files` + 3× 3D NIfTI via `Nyxus3D` (non-unit `scl_slope` makes the HU rescale
  observable). All fixtures are committed, so nothing skips in CI.

Fixtures are reproducible via the committed `make_*_fixtures.py` generators.

## Known limitation (by design)

Non-IH intensity features report the **offset domain** (recoverable by adding `floor(min)`
back); only the intensity-histogram family reconstructs absolute HU automatically.

## Commits

- `feat(hounsfield): --preserve-hu CT/Hounsfield intensity preservation`
- `feat(hounsfield): wire --preserve-hu through the 3D NIfTI loader`
- `test(hounsfield): pydicom oracle for HU features; split HU tests by vetting kind; add pydicom to vetting catalog`

---

## Update — hardening the tests against the `tests/vetting` framework

Reviewed the HU tests against the oracle-vetting framework (`tests/vetting/SPEC.md`). The suite was
broad and used correct (exact) tolerances, but most assertions were `analytic` **self-consistent** —
the synthetic fixtures encoded the same offset formula the loader applies (SPEC §5.2), and the
user-facing HU *feature* values had **no external-tool oracle** (SPEC §3.1). This update closes that.

**1. New independent oracle — `test_hu_ct_small_pydicom` (SPEC §3.1, §4).**
Featurizes the true-HU field of pydicom's real `CT_small.dcm` slice with `preserve_hu`, reconstructs
absolute HU (`offset + floor(HU_min)`), and asserts `MIN/MAX/MEAN/INTEGRATED` equal values pinned
from **pydicom + numpy** — pixels and goldens come from an independent library, not the Nyxus
formula, so it is a genuine oracle rather than a snapshot. Matches exactly
(MIN −896, MAX 1167, MEAN −119.0738525390625, INTEGRATED −1950906). Fixtures `ct_small_hu.tif` /
`ct_small_mask.tif` are committed; the offline generator `make_ct_small_hu_fixture.py` needs pydicom,
but **CI does not** (it consumes the committed TIFFs) — SPEC §4's "reference tools are never CI deps".

**2. Kind-split per SPEC §6.** `test_hounsfield.h` → `test_hu_analytic.h` (closed-form primitive)

+ `test_hu_mechanics.h` (option-parse plumbing), so a file's name states what kind of check it holds.

**3. `pydicom` added to the SPEC §4 oracle catalog** (`tests/vetting/SPEC.md` §2 token set, §4 table,
notes; `tests/vetting/TOOLS.md` setup + coverage) — the DICOM decode + rescale → HU reference.

**Verification:** C++ UT **700/700**; Python CI **71 passed, 1 pre-existing skip** (the +2 are the new
pydicom oracle cases).

sameeul:
### Candidate follow-up PR — `--preserve-hu` refinements

From a review of this PR. The core design is sound (offset map, scan/load domain consistency via `get_dpequiv_pixel`/`hu_tile_`, sub-min clamp, per-workflow wiring). These three are non-blocking refinements that would make good a single follow-up PR — grouping them here rather than blocking the merge.

**F1 — `preserve_hu` + partial fp-options silently clamps negative HU to 0 (correctness).**
`FpImageOptions::empty()` (`src/nyx/cli_fpimage_options.cpp:61`) checks only `raw_min/max/target_dyn_range`, not `raw_preserve_hu`. So `--preserve-hu` **alone** works (`empty()==true` → offset base = scanned slide min). But combining it with any of `--fpimgmin/--fpimgmax/--fpimgdr` flips `empty()` to false, so the offset base becomes `fpopts.min_intensity()` — which defaults to **0** when `--fpimgmin` isn't given — at `src/nyx/image_loader.cpp:311` (DICOM) / `:322` (NIfTI) / the TIFF `fpmin`, and the IH reconstruction mirrors the same choice in `intensity_histogram.cpp::float_domain_map`. Net: every negative HU clamps to 0, silently, and IH "true HU" is wrong.
*Fix:* in `preserve_hu` mode, always use the scanned slide min as the offset base (ignore fp min/max/dr), or reject the flag combination with an error.

**F2 — Docs: "recoverable by adding the global minimum back" is false for nonlinear intensity features (accuracy).**
Holds for shift features (MEAN, MEDIAN, percentiles, MODE) and is a no-op for shift-invariant ones (VARIANCE, STDDEV, SKEWNESS, KURTOSIS, RANGE, IQR). But `ENERGY = Σx²`, `ROOT_MEAN_SQUARED`, `TOTAL_ENERGY`, and `INTEGRATED_INTENSITY = Σx` are **not** recoverable by "adding the min back" (they need N and the offset). The blanket claim appears in `README.md`, `docs/source/*.rst`, and the Python docstrings.
*Fix:* state which families are reported in true HU (IH_*), which are shift-invariant, and which are not simply recoverable.

**F3 — "global minimum" is actually per-slide (semantics / docs).**
The offset base is each slide's `min_preroi_inten`, not a dataset-global minimum. Per-slide IH reconstruction is correct, but offset-domain intensity features are **not comparable across slides** (each sits in its own offset domain). The docs say "global minimum," which is misleading for multi-image datasets.
*Fix:* correct the wording and note the cross-slide-comparability caveat.

Related: #373 hardens the **non**-`preserve_hu` signed-int16 path (clamp-to-0 to avoid the unsigned wraparound crash). It does **not** address F1, which is inside the `preserve_hu` offset path. Also note #373 and this PR edit the same native-cast lines in `grayscale_tiff.h`, so they'll conflict on merge — the clean end state is this PR's `preserve_hu ? hu_offset(...) : (DataType) v` with #373's signed-clamp in the `(DataType) v` branch.

PR #372, opened 2026-07-14
fix(geomoments): center 2D central moments on the true fractional cen…

darkclad:
…troid

centralMom() computed the origin as (int)originOfX / (int)originOfY, truncating the fractional centroid to whole pixels. That shifted the origin and corrupted every central, normalized, and Hu moment (shape and intensity), most visibly making mu10/mu01 non-zero when they are 0 by definition. The _nt sibling already used the untruncated centroid; this brings both overloads in line.

Testing:
- Regenerate the pinned central/normalized/Hu goldens against an independent numpy/skimage oracle (new tests/vetting/oracles/gen_geomoments_analytic.py, token: analytic); raw spatial moments are unaffected and unchanged.
- Add an asymmetric L-shape fixture whose centroid is a non-half fraction, so the truncation is caught at every moment order rather than only the symmetric-rectangle cases.
- Add an analytic invariant test asserting mu10 = mu01 = 0 for both shape and intensity moments.


darkclad:
## Central moments about the true centroid, not a truncated integer (`main-central-moment-centroid`)

**Type:** Bug fix (correctness) · defect #6 from the validation tracker

### Why

`BasicGeomoms2D::centralMom` computed each central moment as `(x - baseX - (int)originOfX)^p · (y - baseY - (int)originOfY)^q` — it **truncated the centroid to an integer** before subtracting. A
central moment is by definition taken about the *true* (fractional) centroid; truncating shifts the
origin by up to a pixel and corrupts **every** central/normalized/Hu moment. The tell-tale symptom: the
first central moments μ10 and μ01, which are **0 by definition**, came out non-zero. On the unit-test
fixture (centroid exactly (23.5, 19.5)) they were `0.5·m00 = 960` instead of 0. The intensity-weighted
overload had the same cast. The sibling implementation `2d_geomoments_basic_nt.cpp` already subtracts the
un-truncated `originOfX/Y`, so the two Nyxus code paths disagreed — corroborating that the cast is a bug.

A second finding: the existing `oracle_3p_*` moment goldens in `test_2d_geometric_moments.h` were
**mislabeled** — they were the buggy Nyxus output (`CENTRAL_MOMENT_10/01 = 960`), which no external
oracle (skimage/OpenCV give μ10=μ01=0) could have produced. They encoded the defect.

### What changed

- Dropped the two `(int)` casts in `centralMom` (both the plain and real-intensity overloads) in
  `src/nyx/features/2d_geomoments_basic.cpp`, so central moments are taken about the true fractional
  centroid (matching the `_nt` sibling).
- **Regenerated 100 central-moment-derived goldens** (shape + intensity: `CENTRAL_MOMENT_*`,
  `NORM_CENTRAL_MOMENT_*`, `HU_M*`, `IMOM_CM_*`, `IMOM_NCM_*`, `IMOM_HU*`, and the distance-weighted
  variants `WEIGHTED_CENTRAL_MOMENT_*`, `WT_NORM_CTR_MOM_*`, `WEIGHTED_HU_*`, `IMOM_WCM/WNCM/WHU_*`) in
  `test_2d_geometric_moments.h`. Raw moments (`SPAT_MOMENT_*`, `IMOM_RM_*`, the `NORM_SPAT_*`/`IMOM_NRM_*`
  and `WEIGHTED_SPAT_*`/`IMOM_WRM_*` sets) are **not** touched — the bug never affected them.
- Added an offline oracle generator `tests/vetting/oracles/gen_geomoments_analytic.py` (numpy reference,
  cross-checked against `skimage.measure`; token `analytic`) that reproduces every regenerated value, and
  a provenance comment pointing the golden set at it (per `tests/vetting` SPEC §6.4).
- Added an **analytic invariant test** `test_2d_geometric_moments_first_central_moment_analytic` asserting
  μ10 = μ01 = 0 for both shape and intensity — the definitional property the bug violated.
- Added an **asymmetric-mask oracle test** `test_2d_asymmetric_shape_central_moments_analytic` on an
  L-shaped fixture (30×20 rectangle minus a corner, centroid (12.786, 8.357)). Unlike the symmetric
  rectangle — whose odd-order central moments vanish by symmetry — this exercises **every** central and
  normalized-central order with a distinct non-zero value verified against the numpy oracle, so the
  truncation is caught at all orders, not just μ10/μ01.

### Impact on Nyxus

- **Bug fix / correctness:** ~22 shape columns plus the intensity and distance-weighted moment families
  are now correct — they are centred on the true centroid. Every regenerated value was verified against an
  independent numpy/skimage oracle to double precision (e.g. shape `CENTRAL_MOMENT_20` 368960→**368480**,
  `HU_M1` 0.169618→**0.169358**; intensity `IMOM_CM_10` 240984→**0**). The previously "oracle-verified"
  goldens that actually encoded the bug now genuinely match an external oracle.
- **Performance:** none to negligible — the change *removes* an `int` truncation from the inner moment
  loop; the arithmetic is otherwise identical (no new work, allocations, or branches). If anything it is a
  hair cheaper.

### Testing

Three layers, all against the checked-in numpy/skimage oracle (`gen_geomoments_analytic.py`):

1. **Corrected goldens** — the four `*_GEOMETRIC_MOMENTS_*` gtests now pass against the regenerated
   central/normalized/Hu values (rectangle fixture, shape + intensity). No raw-moment golden changed.
2. **Analytic invariant** — `test_2d_geometric_moments_first_central_moment_analytic` asserts μ10 = μ01 = 0
   (shape + intensity), the definitional property the truncation broke (was 960 / ~2.4e5).
3. **Asymmetric-mask oracle** — `test_2d_asymmetric_shape_central_moments_analytic` on the L-shape fixture
   (centroid (12.786, 8.357)) asserts all 16 `CENTRAL_MOMENT_*` + 7 `NORM_CENTRAL_MOMENT_*` against the
   oracle. Because the shape is asymmetric, **every** order is a distinct non-zero value the bug corrupts
   (whereas the symmetric rectangle zeroes all odd orders), so this catches the defect at all orders — the
   coverage the rectangle-only test lacked.

- **Full suites green:** 689/689 C++ gtests, 63 passed / 1 skipped pytest (incl. the Hu-moment pydicom
  test). Diff is line-ending-clean.

> **Spun-off finding (separate PR):** the asymmetric fixture also exposed a pre-existing bug in the Hu
> invariants `HU_M5` and `HU_M6` (`calcHu_imp`, 2d_geomoments_basic.cpp) — `h5` squares `3·(η30+η12)`
> instead of scaling `(η30+η12)²` by 3, and `h6` has a precedence error (`4·η11·(η30+η12)·η21 + η03`
> instead of `4·η11·(η30+η12)·(η21+η03)`). Independent of the centroid fix, tracked as PR-8
> `main-hu-invariants`. PR-2's asymmetric test therefore asserts the 23 CENTRAL/NORM_CENTRAL values only;
> Hu is asserted on the rectangle fixture (where M1–4,7 are correct).

darkclad:
Already implemented by other PR

PR #14, opened 2022-03-16
Build Windows wheels

sameeul:

PR #373, opened 2026-07-14
fix(loader): clamp negative signed-int pixels so int16 CT doesn't wra…

darkclad:
…p to a huge unsigned value

### DIFF
--- a/src/nyx/grayscale_tiff.h
+++ b/src/nyx/grayscale_tiff.h
@@ -12,6 +12,7 @@
 #endif
 #include <cmath>    // HU mode: std::floor / std::llround for the offset map
 #include <cstring>
+#include <type_traits>	// std::is_signed_v to guard signed->unsigned wraparound in loadTile
 #include <sstream>
 #include <limits.h> // for INT_MAX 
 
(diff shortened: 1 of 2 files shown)


darkclad:
## Clamp negative signed-integer pixels so int16 CT doesn't wrap (`main-hu-int16-guard`)

**Type:** Bug fix (robustness) · defect #14 from the validation tracker (surfaced by PR #369's macOS CI)

### Why

The pipeline's intensity type `PixIntens` is **unsigned**. A signed-integer TIFF (e.g. an int16 CT with
negative Hounsfield units) whose ROI contains negative values was cast straight to `PixIntens` in the TIFF
loader, so **−1024 wrapped to ~4.29e9**. That absurd maximum then sized the max-intensity-driven grey-bin /
histogram allocation and **segfaulted on macOS**; Windows/Linux "survived" by allocator luck but emitted
garbage `MIN`/`MAX`/`MEAN`/`RANGE` (measured: MAX/MEAN/RANGE ≈ 4.29e9 instead of the true 40 / −957 / 1064).
The in-flight `preserve_hu` work (PR #369) adds a proper offset map **behind a flag**, but its *no-flag*
path still does the wrapping cast — so a user who doesn't set the flag still hits the crash. This fix closes
that no-flag gap independently.

### What changed

- In `src/nyx/grayscale_tiff.h`, both the **tile loader** (`loadTile`) and the **strip loader** (`copyRow`)
  now clamp negatives of signed integer file types to 0 before the cast: `if constexpr (std::is_signed_v<FileType>) if (v < 0) v = 0;`. `if constexpr` makes it a compile-time no-op for the
  unsigned instantiations, so valid unsigned data is untouched.
- Added `#include <type_traits>`.

### Impact on Nyxus

- **Robustness / bug fix:** signed-int16 (and other signed-integer) images with in-ROI negatives no longer
  wrap; the crash-triggering ~4.29e9 maximum is gone and features are finite. Negatives are floored to 0 as
  a safe fallback for the no-`preserve_hu` path (proper HU offsetting remains the `preserve_hu` path in
  PR #369 — this change folds cleanly into that PR's `preserve_hu ? hu_offset(v) : (DataType)v` no-flag
  branch on merge).
- **No change to valid unsigned data** (the guard compiles out for unsigned types), and no change to the
  float/real-intensity path (which already rescales).
- **Performance:** negligible — one compile-time-guarded sign check per pixel on signed inputs only.

### Testing

- New `tests/python/test_signed_int16_loader.py` drives the **`featurize_files` (TIFF) path** the in-memory
  numpy path doesn't cover: a signed int16 image with negative HU across the ROI now yields finite
  `MAX=40 / MIN=0 / RANGE=40` (was ~4.29e9); a second test confirms valid unsigned int16 is unaffected.
- **Full suites green:** 687/687 C++ gtests, 65 passed / 1 skipped pytest. Diff is line-ending-clean
  (`git diff --stat` == `--stat --ignore-all-space`).

PR #378, opened 2026-07-17
fix(3d): compute 3ROBUST_MEAN (was hardcoded 0) + vet 18 3D first-ord…

darkclad:
…er features

Port the 2D robust-mean (mean of voxels in [P10,P90]) into D3_VoxelIntensityFeatures trivial + osized paths; it previously defaulted to 0. Update the 3D firstorder regression snapshot that had pinned the bug value (3ROBUST_MEAN 0 -> 1977.52). Vet 18 of the 19 3D first-order regression features against the Octave/MATLAB oracle (rel<1e-9) and mark them vetted in oracle_coverage.csv (firstorder 53->71 vetted). 3COVERED_IMAGE_INTENSITY_RANGE stays regression (bug-degenerate: slide_idx never set in workflow_3d_segmented.cpp).

Verified: gtest 702/702 pass; pytest tests/python 72 passed (7 arrow deselected, lean build).


darkclad:
## Summary

Fixes a 3D first-order feature that was silently emitting `0`, and records external-oracle
vetting for 18 of the 19 previously regression-only 3D first-order features.

## The bug: `3ROBUST_MEAN` always returned 0

`D3_VoxelIntensityFeatures` (`src/nyx/features/3d_intensity.cpp`) never computed
`val_ROBUST_MEAN` — it defaulted to `0` in both the trivial `calculate()` and the out-of-core
`osized_calculate()` normal paths (the only assignment was the degenerate zero-fill guard). The
2D `PixelIntensityFeatures` computes it correctly (`intensity.cpp:181/318`); it was simply never
ported to 3D.

**Fix:** port the 2D definition — mean of voxels within the `[P10, P90]` window — into both 3D
paths, using `p10_/p90_` from the histogram stats already computed there.

## Vetting: 18/19 3D first-order features

Vetted against an independent Octave/MATLAB oracle (harness in `octave/oracle_3d/`, external to
this repo): 12 raw-formula features + 5 percentiles + the fixed `3ROBUST_MEAN`, all agreeing to
< 1e-9 (percentiles reproduce Nyxus's 100-bin CDF-interpolation exactly; ~0.2% vs `prctile`,
definitional). These move `regression → vetted` in `tests/vetting/oracle_coverage.csv`
(firstorder **53 → 71** vetted; `coverage_report.md` regenerated).

## Deferred (out of scope here): `3COVERED_IMAGE_INTENSITY_RANGE`

Left as `regression` with `flag=bug-degenerate`. It always emits the degenerate `1` because
`roi.slide_idx` is never set in `workflow_3d_segmented.cpp` (2D sets it via `pixel_feed.cpp`), so
`3d_intensity.cpp` takes the `else val=1` branch. Fixing it means wiring `slide_idx` +
pre-ROI min/max through the 3D pipeline — a larger change warranting its own PR. (The
`preserve_hu` block at `phase2_3d.cpp:38` already reads `slide_idx` as if set, so 3D preserve-HU
likely shares this gap.)

## Testing

- **C++ gtest:** `702/702` pass. The 3D firstorder regression snapshot had pinned the bug value
  (`3ROBUST_MEAN = 0`); updated to the corrected `1977.52` (fixture mean is `1983.32`; robust mean
  trims slightly below, as expected).
- **Python:** `pytest tests/python/` — 72 passed, 1 skipped (7 arrow/parquet tests require an
  Arrow-enabled build; unrelated to this change).

## Files
| File | Change |
|---|---|
| `src/nyx/features/3d_intensity.cpp` | compute `3ROBUST_MEAN` in both paths |
| `tests/test_3d_coverage_common.h` | regression golden `0 → 1977.52` |
| `tests/vetting/oracle_coverage.csv` | 18 rows `regression → vetted`; `3COVERED` root-cause note |
| `tests/vetting/coverage_report.md` | regenerated (firstorder `53/19 → 71/1`) |

sameeul:
**Independent verification + a follow-up for the IO audit**

I re-verified the 18 promoted 3D first-order goldens with **intrinsic** Octave (`mean`/`std`/`var`, core `quantile`, `median`, `mode`, moment-based skewness/kurtosis — *not* a reimplementation of Nyxus's binned method), on the ROI (`ut_inten.nii` where `ut_mask57.nii == 57`, 274,432 voxels). **All 18 match to <0.25%** — percentiles ≤0.23% (canonical `quantile` vs the 100-bin CDF, the "~0.2% vs prctile" gap the PR already documents), and `3ROBUST_MEAN` = 1976.6 ≈ golden 1977.5 confirms the fix. **So the numbers are sound and this isn't a blocker** — flagging a systemic observation for later.

Reproducing the goldens required one non-obvious step: 3D first-order runs on **offset, integer-cast intensities** — `floor(raw) − (global slide min)`, i.e. `+1024` on this CT-like phantom (native `[-1024, 2000]`), so `3MEAN = 1983` is `native_mean (959) + 1024`. The tell is clean: shift-invariant features (`STD`, `VARIANCE`, `STD_ERROR`, `EXCESS_KURTOSIS`, `HYPERSKEW/FLAT`, `MAD`) match native directly; shift-sensitive ones (`MEAN`, `INTEGRATED`, `COV`, all percentiles, `QCOD`, `ROBUST_MEAN`, `UNIFORMITY_PIU`, `MODE`) only match once the offset is applied.

The reason this is worth surfacing is that the behavior isn't consistent across the IO paths (from code inspection):
- **3D NIfTI:** offset by `−min` when negatives exist (`raw_nifti.h::unhounsfield`) — and this predates the HU work (PR #274, not #369).
- **2D float TIFF:** min-max **rescale** to `[0, target_DR = 1e4]` (a scale, not a shift).
- **2D signed int:** negatives **clamped to 0** (#373).
- Within 2D, first-order reports on the mapped values while `IntensityHistogram` reconstructs to the native float domain (`float_domain_map`).

So the same signed/float image lands in different, non-native intensity domains depending on dimensionality, source type, and even feature family — and absolute-intensity features aren't comparable across 2D↔3D or to native values.

**Proposal (not for this PR):** capture this as an item for the upcoming **IO audit** — reconcile these intensity-domain conventions across loaders (2D/3D, int/float/signed) and feature families so the behavior is consistent and intuitive (one documented convention, predictable native-vs-offset semantics, 2D/3D parity, uniform reporting). In the interim, a one-line note on the affected `oracle_coverage.csv` rows and at `raw_nifti.h::unhounsfield` would keep the current offset domain from being mistaken for native.

*(Verification detail: 3D checked empirically via Octave intrinsic formulas; the 2D specifics above are from code inspection.)*

PR #381, opened 2026-07-17
fix(caliper): correct Martin/Nassenstein diameters + vet 12 features vs imea

darkclad:
Martin and Nassenstein shared byte-identical code that pushed both the
shortest and longest of a Y-grid of horizontal chords per rotation angle;
neither matched its named definition and the per-angle shortest (a spurious
near-apex ~0-length chord) drove STAT_NASSENSTEIN_DIAM_MIN/MODE to 0.0 --
impossible for a solid shape.

Reimplement on the rotated convex hull to the Pahl/Rumpf-1973 (imea)
definitions: Martin = area-bisecting horizontal chord; Nassenstein =
bottom-tangent vertical chord (one diameter per angle). Shared by the
in-RAM and out-of-RAM paths.

Vet vs imea on a filled ellipse (TEST_SHAPE2D_CALIPER_MARTIN_NASSENSTEIN_IMEA_ELLIPSE_ORACLE),
update the 8x8 regression goldens, promote 12 features (morphology 46->58).
Suite 704/704.


tests/test_remaining2d_common.h
```diff
@@ -236,6 +279,120 @@ static void calculate_remaining2d_polygonality_feature_values(std::unordered_map
...
+
+	fvals = roi.fvals;
+}
+
+// Assert a caliper stat agrees with imea within a relative tolerance (hull-vs-raster convention gap).
+static void assert_caliper_close_to_imea(
+	const std::vector<std::vector<double>>& fvals,
+	Nyxus::Feature2D feature,
+	const std::string& feature_name,
+	double reltol = 0.15)
```
sameeul: can we tighten this tolerance? 15% seems a lot for an oracle vetting

src/nyx/features/caliper_martin.cpp
```diff
@@ -1,9 +1,46 @@
+#include <algorithm>	// FIX (caliper reimpl): std::min/std::max used by the analytic chord helper
+#include <cmath>		// FIX (caliper reimpl): std::abs
```
sameeul: this is probably not needed here, I do not see any `std::abs`

src/nyx/features/caliper_nassenstein.cpp
```diff
@@ -50,144 +86,49 @@ void CaliperNassensteinFeature::save_value (std::vector<std::vector<double>>& fv
...
 	all_D.clear();
 	for (float theta = 0.f; theta < 180.f; theta += rot_angle_increment)
 	{
 		Rotation::rotate_around_center(convex_hull, theta, CH_rot);
-		auto [minX, minY, maxX, maxY] = AABB::from_pixelcloud(CH_rot);
-
-		std::vector<float> DA;	// Diameters at this angle
-
-		// Iterate y-grid
-		float stepY = (maxY - minY) / float(n_steps);
```
sameeul, line 67: `n_steps` is unused in the new code, can be removed from `caliper.h`, IMO.

sameeul:
**Follow-up (non-blocking): the new analytic caliper helpers inherit sub-pixel loss from integer hull rotation**

Nice fix — routing both the in-RAM and out-of-RAM paths through the corrected `calculate_imp` and reimplementing on the area-bisecting / bottom-tangent definitions is exactly right, and it verifies cleanly against imea.

One thing worth a follow-up PR (not this one): `hull_width_at_y` / `hull_height_at_x` now do real sub-pixel interpolation on the hull edges, but the hull they receive is integer-quantized. `Pixel2` extends `Point2i` (integer `x,y`), and `Rotation::rotate_around_center` truncates each rotated vertex back to int:

```cpp
// rotation.cpp
double x_rot = (p.x - cx) * c - (p.y - cy) * s + cx;
double y_rot = (p.y - cy) * c + (p.x - cx) * s + cy;
Pixel2 p_rot ((float)x_rot, (float)y_rot, p.inten);  // Pixel2(float,...) -> (StatsInt) truncation
```

So every rotated hull vertex is truncated toward zero before the analytic chord math runs. That injects a small, systematic *inward* bias (truncation, not rounding) into Martin/Nassenstein/Feret — all of them share this rotation path. It's **pre-existing behavior, not a regression from this PR**, and the ~1–2px hull-vs-raster convention gap you already document (and the 15% imea reltol) absorbs it, which is why the tests pass.

Suggested follow-up work, in rough priority order:

1. **Round instead of truncate** in the rotation output (`std::lround` before storing), which halves the worst-case bias for free and tightens *all* calipers at once — cheapest win.

2. **Better: give the caliper path a float-precision rotated hull.** `Point2f` already exists (`pixel.h:48`) and `circle.cpp` uses it throughout for exactly this reason — the min-enclosing-circle feature does its geometry in `Point2f` and is the one caliper-family feature this PR vets against imea to <0.1% (vs the 15% reltol the truncated Martin/Nassenstein/Feret need). Add a `Rotation::rotate_around_center` overload that outputs `Point2f` (or `double` pairs) instead of round-tripping through integer `Pixel2`, and template/overload `hull_width_at_y` / `hull_height_at_x` on the point type. Contained refactor, and it lets the imea tolerance drop substantially.

3. **Re-tighten the imea tolerance** in `TEST_SHAPE2D_CALIPER_*_IMEA_ELLIPSE_ORACLE` once the bias is gone — the 15% reltol could likely drop substantially, turning these into much stronger oracle checks.

Not asking you to expand this PR's scope; just flagging it so the sub-pixel-precision cleanup doesn't get lost now that the definitions are finally correct.

src/nyx/features/caliper_martin.cpp
darkclad: Fixed

src/nyx/features/caliper_nassenstein.cpp
darkclad: Fixed

sameeul (approved):

PR #385, opened 2026-07-20
test(ngldm): demote mislabeled 3D NGLDM `_ibsi` test to a regression drift-guard (no external oracle; MIRP disagrees up to ~10x)

darkclad:
Corrects a mislabeled "oracle": `test_3d_ngldm_ibsi.h` was named as an IBSI oracle but is not one.
No source or test-assertion changes — this is a re-classification (rename + registry/docs).

## Why
The `d3ngldm_GT` golden table in that file:
- has **no provenance** (no tool/version/config), contrary to SPEC 6.4;
- is evaluated on the **Nyxus coverage phantom** (`tests/data/nifti/phantoms/ut_inten.nii` +
  `ut_mask57.nii`), **not** the IBSI digital phantom, so IBSI consensus values cannot apply to it.

An independent MIRP NGLDM run on the *same* phantom at the *same* discretisation (fixed_bin_number
n=64, 3D; harness `morph_oracle/mirp_ngldm_3d.py`, offline / never in CI) **disagrees with every
comparable value**, several by an order of magnitude — e.g. GLNU 115443 vs 4350, HDE 261 vs 28,
DCENE 0.14 vs 0.0029. So the passing test is a **drift guard**, not an oracle.

## Changes (4 files, +83/-42, no source change)
- **Rename** `test_3d_ngldm_ibsi.h` -> `test_3d_ngldm_regression.h`; add a header documenting the
  lack of provenance and the MIRP disagreement table. The `_ibsi` suffix is freed for a genuine
  future 3D IBSI-phantom NGLDM oracle. All assertions preserved.
- `test_all.cc`: update the `#include` to the new filename.
- `oracle_coverage.csv`: the 18 3D NGLDM features **remain `status=regression`** (verdicts
  `unproven-reference`; `GLM`/`DCM` are `no-external-oracle` — MIRP emits no grey-level-mean /
  dependence-count-mean column, matching the 2D table's "not in IBSI" note). Each row records the
  MIRP delta. `3NGLDM_DCP` was already `vetted` (via `test_3d_ngldm_mirp.h`) and is unchanged. The
  2D NGLDM rows only have their `current_test` pointer updated to the renamed file.
- `MIGRATION.md`: note the correction and that the NGLDM kind label is now `regression`, not `_ibsi`.

## Net effect
No features are promoted or demoted in agreement; this fixes a misleading label so the registry no
longer implies 3D NGLDM is IBSI-vetted. Promotion still requires a documented, config-matched
external oracle (MIRP is the candidate).


sameeul (changes requested):
**Request changes.** After `ab977814` retracted the promotions, all 18 features stay `regression` — the diff does the opposite of the title. Please rewrite before merge; as written it'd stamp a false "ngldm → 100% vetted" onto history (also: nothing is vetted-ibsi, ngldm stays 20/38, there *is* a +33-line test change, and coverage_report.md is correctly untouched — not regenerated). Suggested: `docs(vetting): retract unproven 3D NGLDM "ibsi" promotions; document MIRP disagreement`.

The retraction itself is the right call.

**Also rename the file, don't just annotate it.** The comment admits it isn't IBSI (no provenance, wrong phantom, ~10× off MIRP), so `git mv test_3d_ngldm_ibsi.h test_3d_ngldm_regression.h` — that's the file the CSV's `current_test` already expected but which never existed. Update the `#include` (`test_all.cc:47`) and point `current_test` back to it. Frees `_ibsi` for a real 3D IBSI oracle later.

**Follow-up (non-blocking):** the test passes 10% against a table ~10× off MIRP — possible real 3D NGLDM bug. Worth a tracked MIRP check.

sameeul (approved):

PR #386, opened 2026-07-20
test(gldzm): promote 2D GLDZM_ZDV; keep 17 3D GLDZM regression -> gldzm 17->18 of 36

darkclad:
Validation pass over the grey-level-distance-zone (GLDZM) family. **Net effect: +1 vetted feature (17 → 18 of 36), not +18.** Registry + regenerated coverage docs only — no source or test-logic changes.

### Promoted (1)

- **2D `GLDZM_ZDV`** → `vetted`, via the already-passing IBSI test `test_gldzm_ibsi.h::TEST_GLDZM_MATRIX_ZDV` (reference `0.0816326530612245`).

### Investigated and left `regression` (all 18 3D `3GLDZM_*`)

- The 3D reference table `d3gldzm_GT` in `test_3d_gldzm_ibsi.h` has **no provenance** (SPEC §6.4 requires tool/version/config for an oracle golden) and is evaluated on the Nyxus coverage phantom `ut_inten.nii`, **not the IBSI digital phantom** — so IBSI consensus values cannot apply to it.
- An independent MIRP GLDZM run (MIRP implements IBSI GLDZM; pyradiomics has none) on the same phantom at the same discretisation (fbn n=64) **disagrees on every feature**, several by an order of magnitude (LDE 11.23 vs 314, SDE 0.382 vs 0.0224, ZDV 3.25 vs 79.7). The passing test is a **drift guard, not an oracle**. A provenance header documenting this is added to `test_3d_gldzm_ibsi.h`.
- `3GLDZM_ZDM` additionally has **no external oracle at all** (IBSI/MIRP emit no zone-distance mean) and a table value (222) that is self-inconsistent with its own variance (79.7 ⇒ std ≈ 8.9).
- `target_test` for all 18 3D rows stays `test_3d_gldzm_regression.h`, so the registry keeps recording the outstanding migration; **no `_ibsi` promotion is claimed** for them.

### Addressing review (@sameeul)

- **Title/body now match the diff** — 1 promotion, 17 investigated-and-kept-regression with a documented caveat, ZDM unvetted. Squashed to a single honest commit so the squash-merge message is accurate.
- **Taxonomy** — reverted `target_test` back to `test_3d_gldzm_regression.h` for the 17 rows that had been repointed at `test_3d_gldzm_ibsi.h`, keeping the "needs migration" signal in the registry rather than cementing the spec violation.

`coverage_report.md` delta: `gldzm 17 → 18` vetted, `19 → 18` regression.


tests/test_3d_gldzm_ibsi.h
```diff
@@ -7,6 +7,32 @@
 #include "../src/nyx/roi_cache.h"
 #include "../src/nyx/features/3d_gldzm.h"
 
+// PROVENANCE: UNKNOWN -- these are NOT IBSI consensus values and this is NOT an oracle table.
```
sameeul, line 10: This should move to tests_3D_gldzm_regression.h according to the spec we agreed upon.

sameeul: The PR title is also not correct. I would suggest to open a new PR with proper commit history so that we can keep a linear history of commits that actually works.

sameeul:
Two things here don't line up with what actually shipped in this diff — flagging both since they'd bite differently depending on how this lands.

## 1. Title/description doesn't match the diff

The title says `gldzm 17->35 of 36 vetted` and the body describes 17 3D features + 2D `GLDZM_ZDV` all being reconciled to `vetted`. Checking the actual `oracle_coverage.csv` diff:

```
grep "^+3D,3GLDZM" oracle_coverage.csv-diff | grep -c ",regression,"   → 18
grep "^+3D,3GLDZM" oracle_coverage.csv-diff | grep -c ",vetted,"      → 0
```

Only **`2D GLDZM_ZDV`** is actually promoted to `vetted` (via the already-passing `test_gldzm_ibsi.h::TEST_GLDZM_MATRIX_ZDV`, IBSI value `0.0816326530612245`). All 18 `3D,3GLDZM_*` rows stay `regression` — matching `coverage_report.md`'s own delta (`17→18`, i.e. **+1**, not +18). The new comment block in `test_3d_gldzm_ibsi.h` actually argues *against* promoting the 3D features (no provenance, wrong phantom, MIRP disagrees by up to 24x), which directly contradicts the PR's headline claim.

This matters beyond cosmetics: if this squash-merges, GitHub defaults the squash commit message to the PR title, so "gldzm 35/36 vetted" becomes the permanent history for a change that left 18/36 in `regression`. Could you retitle to something like `gldzm 17->18 of 36 vetted` and rewrite the body to reflect what actually happened — 1 promotion, 17 investigated-and-reverted with a documented caveat, ZDM left unvetted?

## 2. `target_test` papers over a taxonomy violation instead of fixing it

Per `SPEC.md` §2/§6.1, a `_<oracle>`-suffixed file (like `_ibsi`) makes a correctness claim and requires §6.4 provenance (tool + version + config + generator). The 17 `3GLDZM_*` values fail that — no provenance, wrong phantom, large MIRP disagreement — which by the SPEC's own taxonomy makes them regression-kind, belonging in a `_regression`-suffixed file, not `_ibsi`.

Before this PR, `target_test` for all 18 rows (including `ZDM`) pointed at `test_3d_gldzm_regression.h`. I checked — **that file doesn't exist**, on either `main` or this PR's own head branch (`darkclad/nyxus:main-gldzm`). So that value was never a real destination, it was a "needs migration" marker for work that's never been done. This PR changes `target_test` to `test_3d_gldzm_ibsi.h` for 17 of the 18 rows — i.e. it repoints the marker at where the values *currently, spec-violatingly, live* instead of doing the migration the marker was flagging. Only `3GLDZM_ZDM` keeps the old (aspirational-but-honest) `test_3d_gldzm_regression.h` target:

```
3GLDZM_SDE   -> target_test=test_3d_gldzm_ibsi.h        flag=unproven-reference
...(16 more, same pattern)...
3GLDZM_ZDM   -> target_test=test_3d_gldzm_regression.h  flag=no-external-oracle
```

`ZDM` and the other 17 are in the identical epistemic state (no oracle, `regression` status, same caveat language), but only `ZDM` still carries the outstanding-migration signal. The net effect of this PR is that the taxonomy violation — 17 unproven, regression-status values sitting in a file named like an oracle file — becomes permanent and un-flagged in the registry, rather than tracked as work still to do.

Could you either (a) actually create `test_3d_gldzm_regression.h` and move these 17 goldens + gtest cases into it, or (b) at minimum revert `target_test` back to `test_3d_gldzm_regression.h` for all 17 so the registry keeps recording this as unresolved rather than quietly closing it out?

Everything else — CSV parsing (quoted commas are fine, `check_coverage.py` uses `csv.DictReader`), test wiring (`test_all.cc:26,46` — both confirmed included), and the ZDM statistical argument (var=79.7 ⇒ std≈8.9, so mean=222 is impossible; 15.31 is self-consistent) — checks out.

PR #387, opened 2026-07-20
fix(3d-ngldm): DCNU/DCNUN summed the wrong marginal (== GLNU)

darkclad:
3D NGLDM Dependence Count Non-Uniformity computed the GREY-LEVEL (row) marginal s_{i.} = sum_j s_ij -- the outer loop iterated grey level i and summed over j, byte-identical to the GLNU loop directly above it. So 3NGLDM_DCNU == 3NGLDM_GLNU and 3NGLDM_DCNUN == 3NGLDM_GLNUN for every ROI, which is wrong: DCNU is the DEPENDENCE-COUNT (column) marginal s_{.j} = sum_i s_ij, then sum_j s_{.j}^2.

Transpose the loops (outer over dependence-count columns j, inner sum over grey levels i), matching the row-marginal's j in [1,Nr) convention. The 2D twin ngldm.cpp was already correct (it builds separate Sg/Sr marginal arrays).

On the 3D coverage phantom (ut_inten.nii label 57, GREYDEPTH=64): DCNU 115443.18 -> 85056.84, DCNUN 0.22575716 -> 0.16633456; GLNU/GLNUN unchanged, confirming the fix is isolated to DCNU. Updated the two pinned snapshots (test_3d_coverage_common.h, test_3d_ngldm_ibsi.h) to the corrected values. Full gtest suite 706/706.

NOTE: the corrected DCNU (85056.84) still does not match an independent MIRP run (40745) -- that is a separate NGLDM matrix-construction / binning divergence from IBSI, tracked as regression on branch main-ngldm; this commit only fixes the transposed-marginal bug and is not an oracle-agreement claim.


sameeul:
The fix itself is correct — verified independently, not just read-through:

- `3d_ngldm.cpp:305-320`: old code summed the row marginal `s_{i.}` for both the GLNU loop and the "DCNU" loop (byte-identical), so `DCNU`/`DCNUN` were always equal to `GLNU`/`GLNUN`. Confirmed pre-fix in `test_3d_coverage_common.h`: all four values were identical (`115443.18172715895` / `0.22575716076180957`). The fix transposes to the column marginal `s_{.j}`, matching the documented formula on line 353 (`F_DCNU = (1/Ns) Σ_j s_{.j}²`).
- Independent numeric check: since raw `DCNU`/`DCNUN` accumulate identically before dividing by `Ns`/`Ns²`, `DCNU/DCNUN` must equal the same `Ns` as `GLNU/GLNUN`. `115443.18.../0.2257... = 511360.00000000006` vs new `85056.84.../0.16633... = 511359.99999999994` — matches to 1e-6. That's real evidence the new golden values are self-consistent, not just plausible-looking.
- The claim that the 2D twin (`ngldm.cpp`) was already correct checks out — it uses separate `Sg`/`Sr` marginal arrays (`ngldm.cpp:308-316`), no shared bug there.

## But this branch is stale, and merging as-is won't actually fix anything on `main`

`darkclad/nyxus:main` is **25 commits behind** `PolusAI/nyxus:main` (0 ahead). In that gap, `tests/test_3d_ngldm_ibsi.h` was already renamed to `tests/test_3d_ngldm_regression.h` on `main`, with a header already added: *"REGRESSION / drift-guard tests — NOT an oracle"*, citing the same GLNU/DCNU-vs-MIRP disagreement numbers this PR's description cites independently. So the registry's provenance/naming concern (regression-status values living in an `_ibsi`-suffixed file) is already resolved upstream — just not in this branch.

Concretely:
- This PR updates golden values in `test_3d_ngldm_ibsi.h`, which **doesn't exist at that path on current `main`**.
- The actual current file, `test_3d_ngldm_regression.h`, **still has the unfixed buggy values** (`3NGLDM_DCNU: 115443.0`, same as GLNU) — this PR never touches it.
- The source bug is still present, unfixed, on current `main`'s `src/nyx/features/3d_ngldm.cpp` (verified directly), so the fix itself still applies cleanly.

Net effect if merged as-is: either the merge can't cleanly apply (path doesn't exist), or it resurrects the old misnamed file at the old path — undoing the rename — while `test_3d_ngldm_regression.h` keeps asserting the old wrong value against newly-correct source output, so CI would still fail (or silently pass a stale golden, depending on which file ends up wired into `test_all.cc`).

Could you rebase `main-ngldm-dcnu-fix` onto current `PolusAI/nyxus:main` and reapply the golden-value updates to `test_3d_ngldm_regression.h`? The source fix in `3d_ngldm.cpp` shouldn't need to change.

PR #390, opened 2026-07-23
test(neighbor): oracle-vet 6 features (analytic + CellProfiler) -> neighbor 8/9

darkclad:
Vet the 2D neighbor family against independent/external oracles, taking it from 2/9 to 8/9 vetted.

CellProfiler (real cellprofiler.modules.MeasureObjectNeighbors, Adjacent method, run offline via tests/vetting/oracles/gen_neighbor_cellprofiler.py):
  - NUM_NEIGHBORS, CLOSEST_NEIGHBOR1_DIST: CP reproduces Nyxus exactly on the fixture. Upgrades these two rows from tracker-only to a checked-in generator + gtest (test_neighbor_cellprofiler.h).

Analytic (independent numpy recomputation of Nyxus' documented closed forms in tests/vetting/oracles/gen_neighbor_analytic.py; test_neighbor_analytic.h, 1e-4):
  - CLOSEST_NEIGHBOR2_DIST, CLOSEST_NEIGHBOR1_ANG, CLOSEST_NEIGHBOR2_ANG, ANG_BW_NEIGHBORS_MEAN/STDDEV/MODE. These are deterministic closed forms of the (CP-vetted) neighbor graph. CP's AngleBetweenNeighbors is a subtended angle and its SecondClosestDistance ranks all objects globally -- different definitions, so CP is not a valid oracle for them.

PERCENT_TOUCHING stays regression: audited against real CellProfiler and found to diverge on 3/5 ROIs. Nyxus = distinct contour pixels 8-adjacent to a neighbor / contour length; CP = object outline pixels overlapping a disk(distance+0.5)-dilated neighbor / perimeter. No CP method (Adjacent/Expand/Within) reproduces Nyxus. Documented convention gap.

Removes the two superseded "unvetted_no_direct_oracle" angle regression functions (now tightly vetted by the analytic oracle). Registry updated; coverage regenerated (604 -> 610 / 758). gtest 717/717; pytest 72 passed (7 arrow fails are the tiff-only local build, not regressions).


sameeul:
Files vs. functions are inconsistent with each other here, and only one of them follows `SPEC.md`.

## File names: fine

`test_neighbor_analytic.h` / `test_neighbor_cellprofiler.h` / `test_neighbor_regression.h` all match §6.1 (`test_<family>_<kind-or-oracle>.{h,py}`, oracle token last).

## Test/function names: don't match §6.2

§6.2 says gtest cases follow `test_<family>[_<subject>]_<oracle>` — oracle token **last**, "so the oracle suffix makes vetting status self-evident" (the spec's own stated reason). Examples given there: `test_glcm_ave_pyradiomics`, `test_ih_dispersion_analytic` — subject, then oracle.

The new names here put the oracle token in the middle instead, followed by the subject:

```
test_all.cc:19   TEST_NEIGHBORHOOD2D_CELLPROFILER_COUNTS_FIRST_DISTANCE
test_all.cc:26   TEST_NEIGHBORHOOD2D_ANALYTIC_SECOND_DISTANCE_AND_ANGLES
test_neighbor_cellprofiler.h:172   test_neighborhood2d_cellprofiler_counts_and_first_distance()
test_neighbor_analytic.h:113       test_neighborhood2d_analytic_second_distance_and_angles()
```

Per the spec's own examples these should read subject-then-oracle, e.g. `test_neighborhood2d_counts_first_distance_cellprofiler` / `test_neighborhood2d_second_distance_angles_analytic` — which would also make the function names end the same way their files do, instead of the file ending in the oracle token and the function burying it mid-name.

Small additional inconsistency while in there: the gtest macro isn't a clean uppercase of its own function — `test_neighborhood2d_cellprofiler_counts_and_first_distance` drops "AND" and becomes `..._COUNTS_FIRST_DISTANCE`. §6.2 says the macro should just be the uppercased function name.

Doesn't affect correctness — everything here computes and vets exactly what it claims to (verified independently, see other comment) — but worth a rename pass for consistency with the rest of the registry's naming, since the whole point of the suffix convention is that vetting status is legible from the name alone.

PR #391, opened 2026-07-23
test(morphology): vet DIAMETER_EQUAL_PERIMETER + ORIENTATION + EROSIONS_2_VANISH -> 80/113

darkclad:
Vet three more 2D morphology features against analytic / scikit-image oracles (morphology 77 -> 80), and document why four others in the "future-vettable" list are not cleanly promotable.

Promoted:
  - DIAMETER_EQUAL_PERIMETER -> analytic. == PERIMETER/pi (contour.cpp), pure double arithmetic on the vetted PERIMETER; recomputed at 1e-9 in the existing documented-formula-conformance test (same treatment as CIRCULARITY/ROUNDNESS).
  - ORIENTATION -> skimage. skimage regionprops measures the major-axis angle from the row axis; Nyxus measures the same ellipse from the x axis, so NYXUS == 90 - degrees(skimage.orientation) = 70.4174. Matches to 10 decimals because the ellipse angle is invariant to the pixel-size second-moment correction (it shifts mu20 and mu02 equally). New gtest test_shape2d_skimage_orientation_and_erosions + tests/vetting/oracles/gen_morphology_skimage.py.
  - EROSIONS_2_VANISH -> skimage. Nyxus' 3x3 (8-connected) erosion count == skimage binary_erosion(square(3)) = 1; disk(1)/4-connected gives 2, so the test also pins the connectivity convention.

Documented (stay regression, with reasons in the registry notes):
  - GEODETIC_LENGTH: geo_len_thickness.cpp truncates PERIMETER to size_t and uses integer division, so its value (10.0) deviates ~11% from the real-valued rectangle formula (~11.13). Vetting it analytically would bake in the truncation -- a latent precision issue to fix first.
  - ROI_RADIUS_MEAN/MAX: mean/max per-pixel min SQUARED distance to the ROI's own contour pixels (roi_radius.cpp); scipy distance_transform_edt uses a different reference (distance to background, not squared), so no standard tool reproduces it.
  - EROSIONS_2_VANISH_COMPLEMENT: degenerate 0 on this fixture; needs a holed fixture.
  - MAJOR/MINOR_AXIS_LENGTH, ECCENTRICITY: ~1.4% moment-normalization gap vs skimage, left regression.

gtest 717/717; pytest 72 passed (7 arrow fails are the tiff-only local build, not regressions). Coverage regenerated (604 -> 607 / 758).


sameeul:
Two smaller things, verified by diffing this commit against its own parent locally (rebased a copy onto current `main` to isolate the real content change from noise).

## 1. `DIAMETER_EQUAL_PERIMETER`'s new oracle test is in the wrong file

The registry promotes this to `vetted,analytic` — a real vetting claim (documented-formula conformance: `DIAMETER_EQUAL_PERIMETER == PERIMETER/pi`, recomputed at 1e-9). But the actual assertion is added into an *existing* function inside `test_morphology_regression.h`:

```cpp
// test_shape2d_documented_formula_conformance_no_external_oracle(), inside test_morphology_regression.h
const double dep_formula = P / PI;
ASSERT_NEAR(fvals[static_cast<int>(Nyxus::Feature2D::DIAMETER_EQUAL_PERIMETER)][0], dep_formula, 1e-9)
	<< "DIAMETER_EQUAL_PERIMETER does not match PERIMETER/pi";
```

Per `SPEC.md` §6.1, a file suffixed `_regression` makes no correctness claim — `_<oracle>` (here, `_analytic`) does. This PR already creates `test_morphology_skimage.h` for the two skimage-vetted features; the analytic one should get the equivalent treatment (a `test_morphology_analytic.h`, or at minimum a clearly separated section) rather than living inside the no-external-oracle regression file whose whole purpose is to *not* make correctness claims. Right now the file name says "no oracle" while one of the assertions inside it explicitly is one.

## 2. Undisclosed CRLF→LF rewrite of `oracle_coverage.csv`

This commit converts the entire 759-line `oracle_coverage.csv` from CRLF to LF. Checked directly: diffing this commit against its parent with line endings intact shows ~1520 changed lines (the whole file, add+delete); stripping `\r` first drops that to the real 24 lines of actual content change (the 3 promotions + 2 note rewrites). Noting it for the record — consistent LF is fine, no change needed here.

sameeul:
`DIAMETER_EQUAL_PERIMETER` has a real external oracle available too: `imea.measure_2d.macro.perimeter_equal_diameter(perimeter)` (`perimeter / pi`) — the same formula this PR uses, independently implemented in a third-party package. Verified it against this PR's own PERIMETER value (26.9349412836191): imea gives `8.573658094355881`, matching the pinned `DIAMETER_EQUAL_PERIMETER` (`8.57365809435587`) to double precision.

Full writeup + the same finding for `GEODETIC_LENGTH`/`THICKNESS` + a concrete implementation recipe is on #392: https://github.com/PolusAI/nyxus/pull/392#issuecomment-5175550938 — same ask applies here: vet against `imea` instead of (or alongside) `analytic`.

PR #392, opened 2026-07-23
fix(morphology): geodetic length/thickness from real perimeter (was size_t-truncated)

darkclad:
geo_len_thickness.cpp assigned PERIMETER (a double) into a size_t and then used integer
division, truncating the perimeter to its floor and rounding every intermediate term. On
the 8x8 shape2d fixture this made GEODETIC_LENGTH 10.0 and THICKNESS 3.0 -- ~11% off the
rectangle-model formula (geodetic ~11.13, thickness ~2.34). Keep perimeter and area as
double and use floating-point division, so both features equal the documented side lengths
of the rectangle with the same area and perimeter (the roots of x^2 - (P/2)x + A = 0).

**Vetted against the external `imea` oracle** (per review), not a self-derived formula.
`imea.measure_2d.macro.geodeticlength_and_thickness(area, perimeter)` is a third-party
implementation of the same DIN ISO 9276-6 rectangle model. Fed the Nyxus AREA_PIXELS_COUNT
(26) and PERIMETER (26.9349412836191), imea 0.3.5 returns 11.13182483477333 and
2.3356458070362205 -- the fixed Nyxus values bit for bit (|diff| = 0 for both), so the
goldens are pinned at the "exact" tier (abs 1e-9) in `test_morphology_imea.h`, next to the
DIAMETER_EQUAL_PERIMETER claim, with a checked-in offline generator
(`tests/vetting/oracles/gen_morphology_imea.py`; reference tools stay out of CI).

**Scope of the claim.** This vets the rectangle-model *transform*, not the area and
perimeter it consumes. imea derives its own perimeter from `cv2.arcLength` (12.6569 vs
Nyxus' chain-code 26.9349), which drives `(P/4)^2 - A` negative -- imea clamps the root, so
its *end-to-end* geodeticlength and thickness both collapse to 3.1642 and agree with
neither Nyxus value (differences 7.97 and 0.83). The entire gap is inherited from
PERIMETER, which stays a regression row of its own. Same scope as the
DIAMETER_EQUAL_PERIMETER row from #391. Recorded in the test comment, the generator
docstring and both registry notes.

Registered as `oracle=imea` only, not `imea`+`analytic` as two rows: imea evaluates
literally the same closed form the analytic assertion would recompute, so two rows would
count one formula twice rather than two independent references. Follows #391's precedent for
DIAMETER_EQUAL_PERIMETER, and `test_morphology_analytic.h` notes these are not re-derived
there.

GEODETIC_LENGTH regression -> vetted (imea); THICKNESS keeps `oracle=imea` but is now backed
by a real assertion (its prior row was tracker-only, and the loose 1000x snapshot had pinned
the truncated 3.0). Goldens updated; the redundant loose geodetic gtest removed. No
correctness claim is added to `test_morphology_regression.h`, which claims nothing per SPEC
§2 -- both features were pulled out of it.

Test results: gtest 722/722; pytest 72 passed (the 7 arrow failures are a tiff-only local
build; no Python test covers these features); ASan+UBSan on Linux/gcc 13.3 clean --
gtest 718/718 (722 minus the 4 DCMTK-gated cases) and pytest 72 passed, with
AddressSanitizer / LeakSanitizer / CHECK-failed / UBSan-runtime-error counts all zero.
Negative control: restoring the pre-fix size_t arithmetic makes the new assertion fail by
1.13, so it is not vacuous. Coverage 614 -> 615 / 758; morphology 80 -> 81 of 113.


tests/test_morphology_common.h
```diff
@@ -99,8 +99,11 @@ static std::unordered_map<std::string, double> oracle_3p_shape2d_feature_golden_
 	{"DIAMETER_EQUAL_PERIMETER", 8.57365809435587},
 	{"DIAMETER_CIRCUMSCRIBING_CIRCLE", 12.3317073399088},
 	{"DIAMETER_INSCRIBING_CIRCLE", 0.828486893405308},
-	{"GEODETIC_LENGTH", 10.0},
-	{"THICKNESS", 3.0},
+	// Real-valued rectangle-model roots after the geo_len_thickness.cpp perimeter-truncation fix
+	// (were the integer-truncated 10.0 / 3.0). Now vetted by analytic conformance in
```
vjaganat90: GREAT! I have always suspected these

tests/test_all.cc
```diff
@@ -1012,14 +1012,14 @@ TEST(TEST_NYXUS, TEST_SHAPE2D_CONTOUR_FEATURES)
 	ASSERT_NO_THROW(test_shape2d_contour_features());
 }
 
-TEST(TEST_NYXUS, TEST_SHAPE2D_VERIFIABLE_WITH_3P_BUILTIN_ORACLE_CONTOUR_DIAMETER_EQUAL_PERIMETER)
+TEST(TEST_NYXUS, TEST_SHAPE2D_CONVEX_HULL_FEATURES)
```
vjaganat90, line 1019: Not a blocker. If all the oracle supported tests are in tests/vetting then we should releax the naming of test groups and fixtures to original names, We should discuss this offline.

vjaganat90 (approved):
Overall LGTM!

sameeul:
Before this lands as `analytic`-only: `GEODETIC_LENGTH`/`THICKNESS` have a real external oracle available, and it's a stronger vet than what's here. Already tried it and it matches exactly — writing down the recipe below.

## What I checked

`imea` (the same package already used elsewhere in this repo for Feret/min-enclosing-circle/caliper vetting) has this in `imea/measure_2d/macro.py:248`:

```python
def geodeticlength_and_thickness(area, perimeter):
    # DIN ISO 9276-6: rectangle model,
    # area = geodetic_length * thickness, perimeter = 2*(geodetic_length + thickness)
    v = max((perimeter / 4) ** 2 - area, 0)
    geodetic_length = perimeter / 4 + sqrt(v)
    thickness = perimeter / 2 - geodetic_length
    return geodetic_length, thickness
```

That's the identical rectangle-model formula this PR uses (`P/4 + sqrt(P²/16 − A)`), sourced from a named external standard (DIN ISO 9276-6) and implemented independently by a third party — not just Nyxus checking its own formula against itself. Ran it directly against this PR's own AREA/PERIMETER inputs (26.0, 26.9349412836191):

```
imea geodetic_length: 11.13182483477333   -> matches pinned GEODETIC_LENGTH exactly
imea thickness:        2.3356458070362205  -> matches pinned THICKNESS exactly
```

Exact match to full double precision.

## Why this is worth doing instead of / in addition to `analytic`

The `analytic` oracle here just re-derives Nyxus' own documented formula in the C++ test, so it can only catch a transcription bug — it can't confirm the formula itself is the standard one (it happens to be, but nothing here proves that). `imea` is independently-written, named-standard-sourced, and already trusted elsewhere in this registry. Vetting against it would let both features carry `oracle=imea` (or `imea`+`analytic` both, which per `SPEC.md` §3.1 is the "vetted by ≥2 oracles" higher-confidence tier).

It also closes real tracker debt: the *old* `THICKNESS` row already claimed `oracle=imea`, but its own notes say it was "tracker-only... `test_morphology_imea.h` was never created." Someone already correctly flagged imea as the right oracle for this — it just never got implemented. `GEODETIC_LENGTH` wasn't flagged before, but since `geodeticlength_and_thickness()` returns both from the same two inputs, vetting one gets the other for free.

## Concrete recipe (matching the pattern this PR already uses for skimage)

1. `tests/vetting/oracles/gen_morphology_imea.py` — offline generator, same shape as `gen_morphology_skimage.py`: build the fixture, call `imea.measure_2d.macro.geodeticlength_and_thickness(area, perimeter)` with Nyxus' own AREA/PERIMETER, assert against the pinned goldens, print pass/fail.
2. `tests/test_morphology_imea.h` — new file, `TEST(TEST_NYXUS, TEST_SHAPE2D_GEODETIC_LENGTH_THICKNESS_IMEA)` asserting `GEODETIC_LENGTH`/`THICKNESS` against the imea-derived goldens (`11.13182483477333` / `2.3356458070362205`), included from `test_all.cc`.
3. `oracle_coverage.csv`: set `oracle=imea` (or add a second row per feature for `imea`+`analytic` if you want both tracked), `current_test`/`target_test` → `test_morphology_imea.h`.
4. Pull `GEODETIC_LENGTH`/`THICKNESS` (and honestly `ROUNDNESS`/`DIAMETER_EQUAL_PERIMETER` too, per the other comment) out of `test_morphology_regression.h` — none of them belong in a file whose name says "no oracle."

PR #393, opened 2026-07-23
fix(gabor): keep filter-response magnitudes real-valued, then vet GABOR vs scikit-image

darkclad:
The Gabor pipeline stored each filter-response magnitude in a PixIntens (unsigned int) image and thresholded that, so sub-integer responses were floored to 0 -- e.g. several ROIs reported GABOR = 0.0 at the higher frequencies where the true response is small but nonzero. Keep the response real-valued end to end:
  - CPU (gabor.cpp): GaborEnergy writes double; calculate() scores from a std::vector<double>
    instead of a PixIntens ImageMatrix.
  - non-trivial (gabor_nontriv.cpp): GaborEnergy_NT2 drops the (PixIntens) cast on the magnitude. (Also flags a separate pre-existing bug: osized_calculate leaves originalScore at 0, so the out-of-core baseline score is not wired in.)
  - GPU (gabor.cu, cache.h/.cpp, gabor.cpp): the energy montage and host buffers become real-valued -- RealPixIntens (float, 4 bytes like PixIntens, so device/host buffer sizes are unchanged) for the device montage; double for the host scoring buffers. NOTE: the GPU changes are UNVERIFIED here (no CUDA toolchain in this environment) and need a GPU-CI compile/run; CPU and non-trivial paths are unchanged in behavior relative to each other.

Then VET GABOR vs scikit-image (was regression, flagged no-mainstream-oracle):
  - Nyxus' Gabor kernel is the canonical Gabor filter: it equals skimage.filters.gabor_kernel (frequency = f0/2pi, sigma_x = sig2lam*2pi/f0, sigma_y = sigma_x/gamma) to ~1e-18.
  - An independent reimplementation of the L1-normalize / full-convolve / crop / score pipeline reproduces all 16 golden values (4 DSB2018 ROIs x 4 frequencies) exactly.
  - New generator tests/vetting/oracles/gen_gabor_skimage.py (offline; needs scikit-image). gabor_truth updated to the corrected real-valued scores; the gtest is renamed test_gabor_skimage. gabor 0/1 -> 1/1.

gtest 717/717; pytest 72 passed (7 arrow fails are the tiff-only local build; the pytest gabor tests exercise parameter plumbing / CPU-GPU consistency, not feature values). Coverage 604 -> 605 / 758.


sameeul:
Reviewed this rigorously — the core fix (keeping filter-response magnitudes real-valued instead of truncating to `PixIntens`) is correct, and the vetting math checks out. A few non-blocking things worth attention.

## 1. Undisclosed CRLF/whitespace churn

`git diff` shows 17 lines changed in `oracle_coverage.csv` (16 GLSZM/GLDM rows + the GABOR row), but stripping `\r` (`diff <(tr -d '\r' < old) <(tr -d '\r' < new)`) shows only the GABOR row's content actually changed — the other 16 are byte-identical once corrupted duplicate `\r\r\r`/`\r\r` sequences (already present on `main` today) get normalized. Harmless, but given this project's history with undisclosed line-ending rewrites masking real changes, worth a one-line mention in the PR body.

## 2. Naming convention (SPEC §6.1)

The skimage-vetted GABOR assertion lives in `test_gabor_regression.cc`/`.h`. The function itself is correctly named `test_gabor_skimage` (§6.2-compliant, oracle token last) — but per §6.1/§2, a file suffixed `_regression` makes no correctness claim. Same pattern already flagged on `ROUNDNESS`/`DIAMETER_EQUAL_PERIMETER`/`GEODETIC_LENGTH`/`THICKNESS` in #391/#392. Suggest renaming to `test_gabor_skimage.cc`/`.h` (or splitting oracle vs. regression content into separate files).

## 3. Blank `tolerance`, and a labeling nuance in the generator

The CSV `tolerance` column is left blank despite `status=vetted`. The actual gtest assertion has no explicit tolerance and falls back to the project default (~rel 1e-3, per `test_main_nyxus.h`), looser than the "machine precision" the notes imply — worth filling in `tolerance=rel=1e-3` (or tighter, see below).

Related: `gen_gabor_skimage.py`'s Part B (the code path that actually produces the 16 pinned goldens) builds Nyxus' kernel from a hand-derived numpy formula, not a direct call to `skimage.filters.gabor_kernel` — only Part A (an isolated kernel-shape check) calls the real function. Checked whether this matters: Part A's own comparison reports `max|kernel diff|` of `1.7e-18`–`3.5e-18` (floating-point noise, nowhere near the script's own declared `5e-3` tolerance — the parameter mapping is essentially exact). Went further and substituted the real `skimage.filters.gabor_kernel` output into the *full* feature pipeline end-to-end for all 4 ROIs — every one of the 16 values still matches the pinned goldens to `0.000e+00`. So this changes nothing functionally, but the registry claims `oracle=skimage` while the value-generating code never actually calls it — only a side-check does. Given how much this repo cares about provenance being self-evident (§6.4), suggest having Part B call `skimage.filters.gabor_kernel` directly rather than relying on Part A to vouch for Part B's formula. Not free: skimage's kernel comes back a different array shape (needs Part A's crop/resample/normalize logic in the value path too), plus the same `f0=0` degenerate-case handling Part A already special-cases (`gabor_kernel` can't take `frequency=0`).

## Verified correct, no action needed

- The `PixIntens`→`double`/`RealPixIntens` fix and GPU buffer sizing are consistent with `pixel.h`'s typedefs.
- The `calculate_gpu_multi_filter`→`calculate_gpu` test-call arg-count fix matches `gabor.h`'s actual signature (the old call had the wrong arity).
- No duplicate/stale golden stores elsewhere in the tree for this feature.
- No scope creep — diff is Gabor-only.
- `oracle=skimage` is a valid SPEC §4 token.

PR #394, opened 2026-07-23
fix(imq): correct FOCUS_SCORE overflow + vet 4 image-quality features vs analytic oracle

darkclad:
FOCUS_SCORE and LOCAL_FOCUS_SCORE were dominated by an unsigned-integer overflow: focus_score.cpp's laplacian() multiplied PixIntens (unsigned) pixels by the negative Laplacian kernel weights, so e.g. -4 wrapped to ~4.29e9 and FOCUS_SCORE reached 2.81e18. Cast the pixel to double before the multiply -> FOCUS_SCORE 12.109, LOCAL_FOCUS_SCORE 2.681.

Vet 4 of the 6 imq features against an independent numpy/scipy reimplementation of their documented algorithms (tests/vetting/oracles/gen_imq_analytic.py; reproduces every value exactly):
  - FOCUS_SCORE       = variance of the Laplacian magnitude (scipy convolution) = 12.109
  - LOCAL_FOCUS_SCORE = same on the top-left h/2 x w/2 tile / 4 = 2.681
  - MIN_SATURATION    = fraction of pixels at the image minimum = 18/96 = 0.1875
  - MAX_SATURATION    = fraction of pixels at the image maximum = 16/96 = 0.16667
imq 0/6 -> 4/6.

The other two stay regression with documented reasons (oracle_coverage.csv notes):
  - POWER_SPECTRUM_SLOPE: 0.0 on this fixture (rps() early-returns when min(h,w)/8 < 3), and its radial binning uses the FFT value as the radius index instead of the frequency radius -- a real bug; needs a rewrite + a >=24 px fixture.
  - SHARPNESS: correct (all-double DOM sharpness port, no bug); needs a reference-DOM oracle.

gtest 717/717; pytest 72 passed (7 arrow fails are the tiff-only local build; the pytest imq tests check the ImageQuality class runs, not feature values). Coverage 604 -> 608 / 758.


src/nyx/features/focus_score.cpp
```diff
@@ -254,7 +254,10 @@ void FocusScoreFeature::laplacian(const std::vector<PixIntens>& image, std::vect
 
                     if(ii >= 0 && ii < m_image && jj >= 0 && jj < n_image &&
                        ikFlip >= 0 && jkFlip >=0 && ikFlip < m_kernel && jkFlip < n_kernel){
-                        out[i* n_image + j] += image[ii * n_image + jj] * kernel[ikFlip * n_kernel + jkFlip];
+                        // Cast the pixel to double before multiplying: image is PixIntens (unsigned),
+                        // so the negative kernel weights (e.g. -4) were converted to a huge unsigned
+                        // value, wrapping the Laplacian (focus score reached ~1e18 from the overflow).
+                        out[i* n_image + j] += (double)image[ii * n_image + jj] * kernel[ikFlip * n_kernel + jkFlip];
```
vjaganat90: please use static_cast<double>. If not now then open a PR where we port ALL C style casts to C++ safe casts.

tests/test_imq_regression.h
```diff
@@ -15,17 +15,22 @@
...
 
 void test_focus_score_feature() {
-    
+
     FocusScoreFeature f;
-    double truth_value = 2.810e18;
+    // FOCUS_SCORE = variance of the Laplacian magnitude, vetted vs scipy (gen_imq_analytic.py).
+    // Was 2.810e18: focus_score.cpp multiplied PixIntens (unsigned) by the negative kernel weights,
+    // wrapping the Laplacian; now cast to double.
+    double truth_value = 12.109375;
```
vjaganat90: is this the oracle value or regression value? In either case why did it change?

vjaganat90 (changes requested):

tests/test_imq_regression.h
vjaganat90: It's scipy oracle. This is resolved for me.

vjaganat90 (approved):
LGTM. With the view that we will have a PR which fixes all the C-style casting redundant keywords in the codebase.

sameeul:
`FOCUS_SCORE`/`LOCAL_FOCUS_SCORE` still have a bug after this PR — not the overflow this PR fixes, a different one in `variance()`:

```cpp
double FocusScoreFeature::variance(const std::vector<double>& image) {
    double abs_image_mean = mean(|x| for x in image);
    return mean((|x| - abs_image_mean)^2 for x in image);   // variance of |x|, not variance of x
}
```

This takes the absolute value of the Laplacian *before* computing variance. The metric this feature is named after — and the one every standard tool implements — is the variance of the raw, signed Laplacian (Pech-Pacheco et al. 2000; `cv2.Laplacian(img, cv2.CV_64F).var()`; `skimage.filters.laplace(img).var()`). Taking `|x|` first is a different statistic: `Var(|X|) = E[X²] − E[|X|]²`, which is smaller than the true `Var(X) = E[X²] − E[X]²` whenever the raw Laplacian's mean isn't exactly zero (which it won't be near ROI boundaries, since zero-padding shifts the mean away from 0).

Quantified on this PR's own fixture (`im_quality_intensity`, reconstructed from `test_data.h`):

```
cv2.Laplacian(img, cv2.CV_64F, ksize=1, borderType=cv2.BORDER_CONSTANT).var()  = 34.956597   (textbook)
Nyxus FOCUS_SCORE (this PR's pinned value)                                     = 12.109375
relative difference: 65%
```

That's not a tolerance-band disagreement, it's a different formula.

**The convolution itself is already provably correct** — I ran the real `cv2.Laplacian` call above and it matches Nyxus' kernel + boundary handling (zero-padding) exactly:
```
cv2 zero-pad Laplacian == Nyxus' Laplacian convolution: max abs diff = 0.0
```
So the only change needed is dropping the `abs()` in `variance()` to compute the plain `mean((x - mean(x))^2)`.

**Ask:** fix `variance()` to compute the true variance (drop the `|x|`), then vet both features against real OpenCV instead of `analytic`:
- `FOCUS_SCORE` → `cv2.Laplacian(roi, cv2.CV_64F, ksize=1, borderType=cv2.BORDER_CONSTANT).var()`
- `LOCAL_FOCUS_SCORE` → the same call on the extracted top-left tile, divided by `scale²` — it's the identical convolution on a crop, so the same exact-match proof applies.

This upgrades both from a self-referential `analytic` check (which currently just re-encodes Nyxus' own — buggy — formula and calls it vetted) to a real external-tool oracle, and fixes a real 65% numeric error at the same time.

sameeul:
Correction to something I said in an earlier comment on this thread: I claimed no standard tool matches `MIN_SATURATION`/`MAX_SATURATION`'s convention of checking against the image's own observed min/max (rather than a fixed format threshold like 255). That was wrong — I hadn't actually checked, and a real oracle exists.

## CellProfiler already computes this exact metric

`MeasureImageQuality`'s saturation metrics (`measureimagequality.py:1266`, `calculate_saturation()`):

```python
number_pixels_maximal = numpy.sum(pixel_data == numpy.max(pixel_data))
number_pixels_minimal = numpy.sum(pixel_data == numpy.min(pixel_data))
percent_maximal = 100.0 * number_pixels_maximal / pixel_count
percent_minimal = 100.0 * number_pixels_minimal / pixel_count
```

`numpy.max(pixel_data)`/`numpy.min(pixel_data)` — the image's **own observed** extremum (masked to the ROI when a mask is present), not a fixed bit-depth threshold. That's the identical convention `MIN_SATURATION`/`MAX_SATURATION` use.

Ran it on this PR's actual fixture:

```
CellProfiler PercentMaximal / 100:  0.16666666666666669
CellProfiler PercentMinimal / 100:  0.1875

Nyxus pinned MAX_SATURATION: 0.166667
Nyxus pinned MIN_SATURATION: 0.1875
```

Exact match.

## Ask

`MIN_SATURATION`/`MAX_SATURATION` don't need to stay `analytic`-only — they can be promoted to `oracle=cellprofiler` with a real, verified external match, the same way `NUM_NEIGHBORS`/`CLOSEST_NEIGHBOR1_DIST` already are in this registry (`gen_neighbor_cellprofiler.py` is the existing pattern to copy — same `cellprofiler_core`/`cellprofiler.modules` setup, just swap in `MeasureImageQuality` instead of `MeasureObjectNeighbors`).

This is on top of the separate `FOCUS_SCORE`/`LOCAL_FOCUS_SCORE` finding already posted — that one's an actual bug (65% off from the textbook `cv2.Laplacian().var()` formula); this one's just an available upgrade from `analytic` to a stronger, already-matching external oracle.

PR #401, opened 2026-08-03
fix: move away from C-style casts to C++ static, reinterpret and const casts

vjaganat90:


sameeul:
Close this for now since we need to actually fix all the tests before we do this change. In terms of priority, this is at the very bottom.

PR #429, opened 2026-08-14
fix(intensity_histogram): report IH_ROBUST_MEAN_IDX as a 1-based bin index

darkclad:
IH_ROBUST_MEAN_IDX was emitted 0-based while the family's other eight index-location features all carry the +1, and docs/source/featurelist.rst documents the _IDX half as 1-based. Reported values move up by exactly 1.

Confirmed three ways on two fixtures: the trimmed mean of the 1-based bin indices is 1 higher than the reported value; the family's own centre map VAL = lo + (IDX - 0.5) * binWidth only closes on IH_ROBUST_MEAN_VAL with the corrected index; and coefficientOfVariationIndex in the same function already divides by (meanIndex + 1). The fix is at the emit site on the shared compute() path, so the in-RAM and out-of-core implementations move together.

It survived because its only assertion was a golden labelled "hand-computed ... derived independently of intensity_histogram.cpp" that in fact matched the code. Both goldens are corrected; the phantom one now comes out of a generator.

Vetting for the whole 2D intensity_histogram family: 47 registry rows, 43 vetted (23 mirp, 20 analytic) and 4 demoted to regression.

- test_2d_intensity_histogram_mirp.h pins 23 _IDX goldens from mirp 2.6.0, which Nyxus reproduces to double precision (worst 1.3e-15). A coverage invariant fails if an _IDX feature is added later without a golden.
- test_2d_intensity_histogram_analytic.h pins 19 _VAL and bookkeeping goldens against closed forms written from the published definitions rather than from intensity_histogram.cpp (worst 7.5e-15).
- IH_P10_VAL, IH_P90_VAL, IH_INTERQUANTILE_RANGE_VAL and IH_QUANTILE_COEFFICIENT_OF_DISPERSION_VAL are demoted to status=regression. They use the grouped-data percentile, which no reference implementation reproduces: on the phantom Nyxus reports P90_VAL=4.3125 while all nine of numpy's and all nine of Octave's native methods answer between 4.0 and 5.0, their defaults giving 4.0 - which is what the _IDX half and mirp report - and Octave prctile giving 4.2, the interpolated variant IBSI names as non-reference. A closed form agreeing with Nyxus to 1e-16 only restates Nyxus' own method, so the goldens move to test_2d_intensity_histogram_regression.h as drift guards claiming no oracle. The _IDX percentiles are unaffected.
- test_2d_intensity_histogram_common.h holds the one shared phantom fixture, so the IBSI consensus, the mirp goldens and the closed forms are read off a single computation at a single configuration.
- oracle_coverage.csv: all 47 rows move to source=in-tree, config_recipe= ih.ibsi_fbn, tolerance=rel=1e-9 and target_test cleared. 31 of them previously claimed analytic or ibsi from an offline audit with no generator behind it.
- config_recipes.md: _VAL is not "an affine transform of the IDX distribution". Five conventions coexist and the measured table replaces the claim. The unreferenced ih.mirp_fbn recipe is folded into ih.ibsi_fbn.

Include hygiene over the family: every IH header, plus test_data.h and test_main_nyxus.h, now compiles as the only thing in a translation unit. test_data.h named FTABLE_RECORD without including output_writers.h, which defines it; test_2d_intensity_histogram_regression.h carried an unused <cmath> and an unused test_2d_remaining_common.h; the common and mirp headers called Nyxus-namespaced helpers unqualified and only compiled because another header's file-scope using-directive reached them through test_all.cc's include order.

Audit artifacts under tests/vetting/audit/intensity_histogram_2d_*: the coverage CSV, one report per oracle, and a golden-regeneration guide. scan_intensity_histogram_coverage.py regenerates the CSV from the test sources and --check enforces that the registry and the tests agree row for row and that no gtest function is left unregistered. TOOLS.md gains the conda recipe for mirp and the Octave quantile() argument-order gotcha.


tests/test_2d_intensity_histogram_analytic.h
```diff
@@ -38,16 +39,10 @@ void test_2d_intensity_histogram_dispersion_robust_analytic() {
...
-    ASSERT_TRUE(agrees_gt(fv[(int)F::IH_ROBUST_MEAN_VAL][0], 19.46666667, 1e4));         // oracle=analytic
+    ASSERT_TRUE(agrees_gt(fv[(int)F::IH_ROBUST_MEAN_VAL][0], 19.46666667, 1e4));
     ASSERT_TRUE(agrees_gt(fv[(int)F::IH_ROBUST_MEAN_ABSOLUTE_DEVIATION_IDX][0], 0.6222222222, 1e4));
-    ASSERT_TRUE(agrees_gt(fv[(int)F::IH_ROBUST_MEAN_IDX][0], 1.933333333, 1e4));         // oracle=analytic
-    // QCoD_VAL: not IBSI-anchorable (needs unexposed P25/P75 sum) -> analytic golden here.
-    ASSERT_TRUE(agrees_gt(fv[(int)F::IH_QUANTILE_COEFFICIENT_OF_DISPERSION_VAL][0], 0.3178294574, 1e4)); // oracle=analytic
-    // IQR_VAL: no IBSI anchor -- IDX is the IBSI discrete percentile, VAL interpolates, so they
-    // differ by design (not a defect); vetted analytically here.
-    ASSERT_TRUE(agrees_gt(fv[(int)F::IH_INTERQUANTILE_RANGE_VAL][0], 12.3, 1e4)); // oracle=analytic
+    ASSERT_TRUE(agrees_gt(fv[(int)F::IH_ROBUST_MEAN_IDX][0], 2.933333333, 1e4));
```
sameeul: Should we have a ref val table here to keep things organized?

PR #431, opened 2026-08-14
test(morphology): assert the 2D morphology features against MATLAB, scikit-image and imea

darkclad:
Closes the 2D morphology rows that read status=vetted in oracle_coverage.csv while no in-tree oracle test asserted them, by running each claimed oracle and pinning what it produces.

The planned gap was 17 rows. It is 19: AREA_PIXELS_COUNT and MAJOR_AXIS_LENGTH were credited to test_2d_morphology_analytic.h, which reads them only as inputs to its CIRCULARITY and ROUNDNESS formula checks and asserts neither. Running the oracles turned up a second, unplanned gap of 7 more rows in the imea family, so 26 rows are settled here.

Promoted (22)
- 17 vs MATLAB (GNU Octave 11.3.0 + image 2.20.0 regionprops) on the shape2d fixture: the areas, centroids, weighted centroids, bounding box, aspect ratio, extent, the ellipse triple and the Euler number. Every one agrees to ~1e-15. MATLAB applies the same +1/12 pixel finite-size second-moment correction Nyxus does, which is why MAJOR/MINOR_AXIS_LENGTH and ECCENTRICITY vet here where scikit-image (which omits it, differing 1.4%) could not.
- DIAMETER_EQUAL_AREA vs scikit-image equivalent_diameter_area, exact.
- 4 vs imea on the caliper ellipse: the three caliper _STDDEV statistics and ALLCHORDS_MIN.

Demoted to status=regression, with the measurement in the audit reports (4)
- ROI_RADIUS_MEDIAN. ROI_RADIUS_* report SQUARED distances: RoiRadiusFeature::calculate feeds Pixel2::min_sqdist() straight into the mean/max/median. Measured on filled disks, MAX = (R-1)^2 (82, 362, 1522 for R = 10, 20, 40) - it scales as R^2, so no skimage radius statistic can reproduce it. The units defect, and the fact that min_sqdist() is the approximate hill-descent its own header documents as overestimating (exact_min_sqdist() sits beside it), are behaviour changes and are left for their own branch.
- The three caliper _MODE statistics. The mode of a caliper distribution is an artifact of the angular sampling step: imea's own mode ranges over 19..24 as dalpha goes 5 -> 30, further than the Nyxus-imea gap, so no tolerance separates agreement from sampling noise.

Corrected attributions
- PERIMETER was asserted in test_2d_morphology_matlab.h under a recipe documented as nnz(bwperim(...)) = 846. That recipe returns 846 perimeter PIXELS and regionprops('Perimeter') returns 952.848, while the golden was 999.26 - which is scikit-image's measure.perimeter (999.259018078045, agreeing with Nyxus to 3.8e-15). The assertion moved to the skimage file at full precision and the registry now names the oracle that backs it.
- morphology_2d_imea_shape2d_ref_vals pinned 19 caliper/chord statistics that had never been compared to imea; imea's own values on that 8x8 raster differ by 3.9-79.3%. They are Nyxus snapshots and now live in test_2d_morphology_regression.h. An oracle-named test asserting a snapshot table passes every automated check in the repo, because the oracle is attributed from the function-name suffix - only running the tool finds it.
- The imea ellipse goldens mixed dalpha 9 and 10 and one (Nassenstein MIN = 16.0) matched no run at any step. Regenerated at dalpha=10, the step Nyxus' own calipers sweep (rot_angle_increment, caliper.h), which drops the worst residual from 8.9% to 4.99% and lets the assertion tolerance tighten from 10% to 6%.
- DIAMETER_EQUAL_PERIMETER, GEODETIC_LENGTH and THICKNESS were each pinned twice, in a ref-vals table and again as a literal in the assertion, with the assertions reading the literals; the two copies of DIAMETER_EQUAL_PERIMETER had already drifted. The assertions now read the tables.

Registry hygiene for the whole family, per the revet.txt checklist: every one of the 99 rows now carries a real config_recipe id (three new recipes - morphology.shape2d_native, morphology.perimeter_circles, morphology.caliper_ellipse - plus morphology.fractal_blob512), 28 rows whose target_test was already satisfied have it cleared, and the fraclac goldens gained the SPEC 6.4 provenance pointer they were missing. The six CellProfiler rows are left as the family's one remaining source=tracker claim: no CellProfiler version, config or generator is recorded, so the honest closure is a generator run or a demotion, not a relabel.

Include hygiene: each morphology header now includes the standard headers it uses directly (<string>, <vector>, <unordered_map>, <algorithm>, <cmath>) instead of relying on test_2d_morphology_common.h to drag them in, and imea's include comment named two symbols that live elsewhere or not at all.

Also repairs one pre-existing registry row: 3D/3ROBUST_MEAN was written with an unquoted comma inside candidate_oracle ("[P10,P90]"), splitting it into 15 fields and shifting every column after it, so source read empty and notes read "tracker". coverage_report.md is regenerated, which also corrects a stale intensity_histogram count left from #429.

Generators re-verify every pinned golden in the header they feed and exit non-zero on a mismatch or on a pin they cannot produce: gen_morphology_matlab.m 33/33, gen_morphology_skimage.py 6/6, gen_morphology_imea.py 20/20. All three read their fixtures out of tests/test_data.h rather than carrying a second copy.

Verified: runAllTests 779/779 on Windows and 779/779 under ASan+UBSan on Linux (0 AddressSanitizer reports, 0 runtime errors, 0 leaks); pytest tests/python/ 92 passed, 1 skipped (Arrow-enabled build); check_coverage.py --check, check_test_names.py --check and audit/scan_morphology_coverage.py --check all clean.


tests/vetting/audit/morphology_2d_coverage.csv
```diff
@@ -0,0 +1,100 @@
+Dim,Family,FeatureName,List_of_Oracles,Test_Names,Regression,Reg_Test_Name,Notes
+2D,morphology,AREA_PIXELS_COUNT,matlab,test_2d_morphology_basic_matlab,Y,test_2d_morphology_basic_regression,
```
vjaganat90: Blocking: the Regression=Y entries here are exact duplicate assertions for the same fixture and configuration already covered by the third-party oracle tests. This applies to AREA_PIXELS_COUNT, AREA_UM2, CENTROID_X/Y, WEIGHTED_CENTROID_X/Y, BBOX_XMIN/YMIN/WIDTH/HEIGHT, ASPECT_RATIO, EXTENT, MAJOR_AXIS_LENGTH, MINOR_AXIS_LENGTH, ELONGATION, ECCENTRICITY, EULER_NUMBER, and DIAMETER_EQUAL_AREA (18 assertions total). Both sides call calculate_shape2d_feature_values(), so the snapshot adds no distinct benchmark/configuration signal once the MATLAB or scikit-image assertion is present. Under the SPEC taxonomy, please remove these entries from the regression tables/functions and retain regression coverage only for genuinely different production-only configurations or benchmarks. The separate circles/ellipse/caliper cases are not part of this request. It's NOT even re-computation.

tests/test_2d_morphology_matlab.h
```diff
@@ -1,9 +1,106 @@
...
 
+#include <string>
+#include <vector>
+
 #include "test_2d_morphology_common.h"
 
 #include "test_ref_vals.h"
 
+// MATLAB regionprops goldens for the shape2d fixture (SPEC 6.4 provenance).
+// tool=GNU Octave 11.3.0 + image 2.20.0 (license-free MATLAB stand-in, TOOLS.md);
```
vjaganat90, line 11: Non-blocking: using GNU Octave as the MATLAB proxy is reasonable here, and this line already says so. Could we preserve or mirror that explicit wording in the audit and registry notes wherever oracle=matlab is presented? That will make it clear that these goldens were produced by Octave acting as a MATLAB-compatible proxy, rather than by licensed MATLAB itself.

tests/vetting/audit/morphology_2d_coverage.csv
sameeul: I think this is OK, regression=Y means that we also have a regression test along with the oracle vetted test. We have discussed and agreed upon this.

tests/test_2d_morphology_matlab.h
darkclad: done. SPEC §4 now states that the matlab token names MATLAB regionprops semantics run as GNU Octave + image, and that oracle=matlab means an Octave run unless a table's provenance header says otherwise

tests/vetting/audit/morphology_2d_coverage.csv
vjaganat90: OK, Can we at least agree to reorganize them somewhwere else? In many cases they are pure noise, I fully support implementation drift guards when computation is complicated or convoluted but in many of these cases it is not.

vjaganat90 (approved):
LGTM

PR #433, opened 2026-08-14
test(3d glcm): assert the _AVE features, and restore 26 assertions that never run

darkclad:
Closes the 29 3GLCM_*_AVE rows that read status=vetted with no in-tree oracle assertion, and wires in 26 assertions that existed in the tree but were unreachable.

The _AVE gap was an attribution problem
assert_3d_glcm_feature_pyradiomics() ends in calc_ave(r.fvals[fcode]) - the average over the 13 3D angles - compared against the PyRadiomics golden. That is the right quantity: PyRadiomics reports one value per feature over its whole direction set, which is the Nyxus *_AVE aggregation. But the assertion was booked against the per-angle base feature, while save_value separately writes the same number to the feature the gap rows actually name (3d_glcm.cpp:148). Nothing asserted the stored *_AVE features.

23 of them now assert fvals[..._AVE][0] directly rather than recomputing calc_ave, so a defect in how save_value populates *_AVE fails there and nowhere else. All pass.

The remaining 6 have no PyRadiomics golden of their own - it deprecates DIS as equivalent to DifferenceAverage and does not report ENERGY, ENTROPY, HOM1, SUMVARIANCE or VARIANCE under their own names - so they are vetted through the identities DIS=DIFAVE, ENERGY=ASM, ENTROPY=JE, HOM1=ID, SUMVARIANCE=CLUTEND, VARIANCE=JVAR. Those were already asserted at 1e-6 on the per-angle values; test_3d_glcm_ave_equivalence_pyradiomics now asserts them on the stored *_AVE features and re-checks that the twin still matches its golden, so neither end of the chain can rot silently.

Five rows carried an incoherent oracle
--------------------------------------
3GLCM_{DIFAVE,DIS,IDN,IDMN,INFOMEAS2}_AVE read oracle=mirp while their base rows read oracle=pyradiomics for the same measured quantity, with no notes on either. No mirp run has ever existed for 3D GLCM in this tree - no generator, no goldens, no test file. They now name the tool whose golden the assertion actually compares against.

26 assertions that existed but never executed
---------------------------------------------
test_3d_glcm_regression.h (25 tests) was not #included and had zero TEST() registrations. The cause is mechanical: it carried its own definition of get_3d_segmented_phantom(), which redefines the one in test_3d_glcm_pyradiomics.h inside the single test_all.cc translation unit. The live 3D headers forward-declare it; this one now does too.

Wired in, all 25 failed - and the pins were not merely stale but impossible. ID, IDM, IDN, IDMN and JMAX are bounded in [0,1] by construction and were pinned at 2.5, 2.4, 3.8, 3.9 and 1.86. The old table also broke the SUMVARIANCE == CLUTEND identity the family asserts elsewhere (18057.4 against 18057.0). The regenerated values hold that identity exactly, give DIS == DIFAVE to 3.6e-15, and are all in range. They come from the current implementation, which the sibling PyRadiomics file - which does run - vets independently on the compat phantom. Tolerance tightened 10% -> rel=1e-9, and test_3d_glcm_dump_regression() regenerates the table so the next refresh is a gtest filter rather than a hand transcription.

test_3d_glcm_jvar_pyradiomics was a complete assertion with no TEST() calling it (not_covered.md B.2). Registered; it passes.

The goldens had never been reproduced, and one was wrong -------------------------------------------------------- The family had no generator at all - gen_glcm_{mirp,pyradiomics}.py are 2D-only - so SPEC 6.4's "generator script path" was unmet and the numbers rested on a hand-written CLI recipe in a comment. That comment also contradicted itself, naming the ut_ phantom while every assertion uses the compat one.

gen_glcm3d_pyradiomics.py now runs PyRadiomics 3.0.1 against the compat phantom and re-verifies every pinned golden. 19 of the 23 reproduce bit-for-bit, which confirms the table's origin. Five did not:

  3GLCM_IDN    pinned 0.9822362042997563  ->  0.9067759330416398   (7.7%)
  3GLCM_IDM    pinned 0.4040020605537021  ->  0.3726945904589868   (7.8%)
  3GLCM_ID     pinned 0.47211428859469606 ->  0.4459415317170447   (5.5%)
  3GLCM_IDMN   pinned 0.9822362042997563  ->  0.9797065356412845   (0.26%)
  3GLCM_CORRELATION 0.4305477709920443    ->  0.43309121847659515  (0.59%)

3GLCM_IDN and 3GLCM_IDMN were pinned to byte-identical values - the IDMN number pasted into the IDN slot. PyRadiomics reports different values for the two. It survived because the 10% band absorbed a 7.7% error, which is the "tolerance loose enough to pass a known-bad value" SPEC 7 warns about. The other four are consistent with the inverse-difference family's known version sensitivity.

All five are re-pinned to the fresh run. Nyxus still agrees with the corrected goldens inside the same 10% band - that band covers a real convention gap, Nyxus' asymmetric offset-1 cooc matrix against PyRadiomics' symmetric default.

Registry
--------
All 59 rows get a real config_recipe id (two new: glcm3d.pyradiomics_bincount20 and glcm3d.regression_ut_phantom, which differ in bin count and are therefore not comparable to each other), target_test cleared, source moved off tracker, and current_test synced to the files that actually cover each feature - it previously named test_3d_glcm_regression.h for all 59 when that file asserts 25 and did not run at all.

Noted for PRs 8-10: test_3d_{gldm,glrlm,glszm,ngtdm}_regression.h and test_3d_firstorder_matlab.h are unreachable the same way, but unlike the GLCM file their assert bodies are wrapped in #if 0, so wiring them in yields tests that assert nothing until the body is restored. test_3d_ngldm_regression.h is a third variant - included and running, but its body is #if 0 too, so its 19 tests have been passing without checking anything.

Verified: runAllTests 824/824 on Windows and under ASan+UBSan on Linux; pytest tests/python/ full directory; check_coverage.py --check, check_test_names.py --check and audit/scan_glcm3d_coverage.py --check all clean.


tests/test_3d_glcm_pyradiomics.h
```diff
@@ -1,20 +1,37 @@
...
+// test_3d_glcm_regression.h, at a different bin count; the two benchmarks are not comparable.)
+//
+// Nyxus side: 100 grey levels, GLCM offset 1, asymmetric cooc matrix. PyRadiomics' GLCM is symmetric
+// by default, which is the convention gap the 10% band covers.
 //
-// Getting Pyradiomics ground truth values:
-//      pyradiomics mri.nii.gz liver.nii.gz --param settings1.yaml
+// PyRadiomics reports ONE value per feature over its whole direction set, i.e. the Nyxus *_AVE
+// aggregation over the 13 3D angles -- not a per-angle value. Each golden below is therefore the
+// reference for both the per-angle base feature (through calc_ave) and the stored *_AVE feature.
```
vjaganat90: Blocking: this over-credits the PyRadiomics aggregate. PyRadiomics returns one scalar over its direction set, so comparing it to calc_ave(r.fvals[fcode]) validates only the averaged quantity (and the stored *_AVE slot), not the 13 directional values in the unsuffixed/base feature output. Compensating per-direction errors can leave the mean unchanged. Marking both the base and *_AVE assertions as oracle-vetted therefore conflicts with SPEC's assertion-level feature x config x reference model. Please either keep only *_AVE oracle-vetted and classify the directional vector separately, or add a genuine per-direction reference and assert every element. If direction-level validation is NOT possible then so be it.

tests/vetting/audit/glcm_3d_pyradiomics_vetting_report.md
```diff
@@ -0,0 +1,156 @@
...
+| Test | `test_3d_glcm_pyradiomics.h` |
+| Tolerance | `rel=1e-1`, matching the existing per-angle assertions on the same goldens |
+
+## The gap was an attribution problem, not a missing measurement
+
+`assert_3d_glcm_feature_pyradiomics()` ends in
+
+```cpp
+double atot = f.calc_ave (r.fvals[fcode]);          // average over the 13 3D angles
+ASSERT_TRUE(agrees_gt(atot, glcm_3d_pyradiomics_ref_vals[fname], 10.));
```
vjaganat90, line 23: Blocking: a global 10% band is not a valid oracle recipe. The provenance above explicitly says Nyxus is using an asymmetric GLCM while PyRadiomics is symmetric, so these are not necessarily the same quantity. SPEC requires matching configurations; **tolerance cannot be used to absorb a convention mismatch**.  Please configure both sides equivalently (or limit the oracle to the unaffected subset) and use measured per-feature tolerances, normally <=1% for cross-tool agreement. Any genuinely larger method divergence needs a feature-specific band plus citation; affected unmatched features should remain regression. The same hard-coded 10.0 also appears in the *_AVE and equivalence checks below.

tests/vetting/oracle_coverage.csv
```diff
@@ -588,65 +588,65 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
-3D,3GLCM_SUMENTROPY_AVE,glcm,vetted,pyradiomics,agreed,,,test_3d_glcm_coverage.h;test_nyxus.py;test_3d_glcm_regression.h,test_3d_glcm_regression.h,mirp,,tracker,"vetted by equivalence: 3GLCM_*_AVE = calc_ave(per-direction base), which is exactly the value test_3d_glcm_pyradiomics.h asserts == pyradiomics for the vetted base feature"
-3D,3GLCM_SUMVARIANCE_AVE,glcm,vetted,pyradiomics,agreed,,,test_3d_glcm_coverage.h;test_3d_glcm_regression.h,test_3d_glcm_regression.h,mirp,,tracker,"vetted by equivalence: numerically identical to pyradiomics-vetted 3GLCM_CLUTEND_AVE (rel<1e-6, gtest TEST_3D_GLCM_EQUIVALENCE_DUMP_PYRADIOMICS asserts it). ENTROPY/HOM2 also required the /sum_p fix"
+3D,3GLCM_ACOR,glcm,vetted,pyradiomics,agreed,glcm3d.pyradiomics_bincount20,,test_3d_glcm_coverage.h;test_3d_glcm_pyradiomics.h;test_3d_glcm_regression.h,,,,audit,test_3d_glcm_regression.h is #included in test_all.cc and its 25 assertions run; it pins Nyxus' own values as a drift guard and claims no oracle - the vetting claim on this row rests on test_3d_glcm_pyradiomics.h
```
vjaganat90: Blocking: this row records the unsuffixed 3GLCM_ACOR feature as vetted even though the PyRadiomics assertion reaches it only by averaging its directional vector. That is another assertion about the aggregate, not evidence for the base feature's 13 directional outputs. The semicolon-separated current_test field also mixes coverage, oracle, and regression tests into one feature row and supplies no individual test name or benchmark. SPEC section 3 requires one row per assertion (feature x config x reference), precisely so these claims cannot be conflated. Please keep the PyRadiomics claim on the 3GLCM_ACOR_AVE assertion, leave the base row non-vetted unless there is a per-direction oracle, and normalize the sibling rows to identify each concrete test and benchmark separately.

vjaganat90 (changes requested):

tests/test_3d_glcm_pyradiomics.h
darkclad: Agreed and fixed

tests/vetting/audit/glcm_3d_pyradiomics_vetting_report.md
darkclad: Agreed and fixed

tests/vetting/oracle_coverage.csv
darkclad: Agreed and fixed

vjaganat90:
The major concerns have been addressed. I would urge the header include cleanup pass once more as in [PR 432](https://github.com/PolusAI/nyxus/pull/432)

tests/test_3d_glcm_regression.h
```diff
@@ -1,63 +1,78 @@
 #pragma once
 
 #include <gtest/gtest.h>
+#include <iomanip>
+#include <iostream>
+#include <string>
```
vjaganat90: again `<string>` and `<vector>` not needed at all.

vjaganat90:
I think this is good to go after rebase.

PR #434, opened 2026-08-14
test(3d morphology): assert the five PCA axis features against MIRP, and two MATLAB rows nothing checked

darkclad:
Closes the five 3D morphology rows that read status=vetted, oracle=mirp with no in-tree oracle assertion, plus two more the family plan did not count.

The five axis features
----------------------
3MAJOR_AXIS_LEN, 3MINOR_AXIS_LEN, 3LEAST_AXIS_LEN, 3ELONGATION and 3FLATNESS carried a mirp verdict from an offline harness run (source=tracker) with target_test naming a _regression file. Nothing in the tree compared them against MIRP; the only thing touching them was the coverage sweep, which pins Nyxus' own output as a snapshot.

gen_morphology3d_mirp.py now runs MIRP 2.6.0 against the segmented phantom and test_3d_morphology_mirp.h asserts all five. Same definition on both sides - 4*sqrt of the mask covariance eigenvalues, and their ratios - so the agreement is exact to double precision:

  3MAJOR_AXIS_LEN   104.70681271508683   rel 0
  3MINOR_AXIS_LEN    88.30145986864228   rel 0
  3LEAST_AXIS_LEN    71.51449974198198   rel 2.0e-16
  3ELONGATION         0.8433210559938976 rel 0
  3FLATNESS           0.6829975804590384 rel 1.6e-16

Asserted at rel=1e-9, far tighter than SPEC 7's rel=1e-3 same-definition tier, because that is what the measurement supports.

The generator checks the identities the old defect broke -------------------------------------------------------- This family carried an eigenvalue-ordering defect that produced 3LEAST_AXIS_LEN > 3MAJOR_AXIS_LEN and 3FLATNESS > 1, both structurally impossible. It is fixed in 3d_surface.cpp and these are the corrected values. The generator now asserts, on MIRP's own output, that MAJOR >= MINOR >= LEAST > 0, that ELONGATION and FLATNESS lie in [0,1], and that each equals its defining ratio - so a misconfigured oracle run is caught before anything is pinned. No _invariant file is added: the five oracle pins fix the exact values on this fixture and are strictly stronger than the bounds would be there.

Two rows the plan did not count
-------------------------------
3VOXEL_VOLUME and 3VOLUME_CONVEXHULL read status=vetted, oracle=matlab and had a MATLAB golden and a stated band in morphology_3d_matlab_ref_vals/_ref_tols - but no oracle-named test asserted them. test_3d_morphology_matlab.h registered exactly one function, for 3MESH_VOLUME. The only thing comparing the other two against MATLAB was the parameterized sweep in test_3d_morphology_coverage.h, whose case names carry no oracle token, so the claim rested on a test that does not say what it checks (SPEC 6.2). The report generator misses this the same way it missed two rows in 2D morphology: a feature name in an oracle file counts as coverage whether an assertion reads it or only a table does. Both now have an assertion through the existing helper, goldens and bands. Both pass.

MATLAB cannot be re-run, so MIRP cross-checks it
------------------------------------------------
Three rows claim oracle=matlab and revet step 3 says to run every claimed oracle. Not possible here - no licence, and Octave's image package has no regionprops3 - and the goldens have no in-repo generator (not_covered.md C). The generator prints the corresponding MIRP quantities instead, unpinned:

  voxel volume        MIRP 274432.0   MATLAB 274432.0   exact
  convex-hull volume  MIRP 496958.32  MATLAB 497824.0   0.17%

Both MATLAB goldens are independently corroborated, which also relocates the disagreement: Nyxus' 3VOLUME_CONVEXHULL of 478516 is 3.6% from MATLAB and 3.7% from MIRP while the two tools sit 0.17% apart, so the difference is Nyxus' discrete voxel hull against two triangulated ones - which is what the existing 5% band documents, now with a second measurement behind it.

The surface-area convention gap stays open
------------------------------------------
3AREA counts exposed voxel faces (59992) where MIRP integrates a marching-cubes mesh (46739), a 28% convention difference that no tolerance turns into an agreement. 3AREA and the five features derived from it stay status=regression, with the reason recorded per row rather than left as a bare absence. Settling it changes six public feature values and belongs on its own branch.

Test file cleanup
-----------------
test_3d_morphology_common.h carried a dead #if 0 block - a superseded copy of the whole fixture, ending in a line of asterisks that is not valid C++ and only compiled because it was disabled. Removed. It also included test_ref_vals.h while declaring no reference table. Include hygiene across the family: _common.h relied transitively on <string>, <tuple>, <vector> and helpers/fsystem.h; _matlab.h named Pixel3 and Nyxus::calc_eigvals without including features/pixel.h or helpers/helpers.h, and used std::abs/std::vector without <cmath>/<vector>; _regression.h relied on <string>. All now direct. Three headers carried "Migrated from test_3d_shape.h (Wave 8)" history, which belongs in the audit report.

Registry
--------
All 14 rows get a real config_recipe id (three new: morphology3d.mirp_ibsi, morphology3d.matlab_regionprops3, morphology3d.regression_ut_phantom - one fixture, three benchmarks, so unusually they ARE comparable to each other, which is what makes the MIRP cross-check meaningful), a real tolerance, target_test cleared, source moved off tracker, and current_test synced to the files that actually cover each feature.

Verified: runAllTests 780/780 on Windows and under ASan+UBSan on Linux; pytest tests/python/ full directory (92 passed, 1 skipped); check_coverage.py --check, check_test_names.py --check and audit/scan_morphology3d_coverage.py --check all clean; gen_morphology3d_mirp.py re-verifies all five goldens and all four structural identities against a fresh MIRP 2.6.0 run.


vjaganat90 (commented):
1. **Remove the two redundant regression assertions.** `3VOXEL_VOLUME` is now covered by `TEST_3D_MORPHOLOGY_VOXEL_VOLUME_MATLAB`, but `TEST_3D_MORPHOLOGY_VOXEL_VOLUME_REGRESSION` still checks the same feature through the same `calculate_3d_morphology_feature_value()` helper. Likewise, `3VOLUME_CONVEXHULL` is covered by `TEST_3D_MORPHOLOGY_VOLUME_CONVEX_HULL_MATLAB` while `TEST_3D_MORPHOLOGY_VOLUME_CONVEX_HULL_REGRESSION` checks the same point. All four tests use `ut_inten.nii + ut_mask57.nii`, label 57, `GREYDEPTH=128`, and `IBSI=true`; there is no distinct production-only configuration for the regression tests to protect. Per the oracle-vs-regression convention, please remove the two regression functions/registrations. I would recommend quarantining them but, Honestly these "backup" tests are NOT worth the hassle.

2. **Remove the dead `3MESH_VOLUME` regression golden.** `morphology_3d_regression_ref_vals` contains `{"3MESH_VOLUME", 478516}`, but there is no `test_3d_morphology_mesh_volume_regression()` and no corresponding GoogleTest registration. The only executable assertion for this feature is the MATLAB test, and the coverage CSV already records `Regression=N`.

3. **Complete MATLAB golden provenance.** The comment above `morphology_3d_matlab_ref_vals` explicitly says `generated = offline; no in-repo generator`. That leaves all three MATLAB goldens noncompliant with SPEC §6.4, which requires tool, version, exact configuration, and a checked-in runnable generator path at the definition site. Please add the MATLAB/Octave generator script and exact invocation/configuration (or reclassify these values if they cannot be regenerated).

4. **Cite the 5% method-divergence band.** The comments and MIRP report measure and explain the approximately 3.6% hull difference, but SPEC §7 requires a stated band **plus a citation** for known method divergence. Please cite the MATLAB `regionprops3` `ConvexVolume` definition and the documented Nyxus/IBSI/MIRP hull or mesh-volume convention used to justify applying 5% to `3VOLUME_CONVEXHULL` and `3MESH_VOLUME`.

darkclad:
> 1. **Remove the two redundant regression assertions.** `3VOXEL_VOLUME` is now covered by `TEST_3D_MORPHOLOGY_VOXEL_VOLUME_MATLAB`, but `TEST_3D_MORPHOLOGY_VOXEL_VOLUME_REGRESSION` still checks the same feature through the same `calculate_3d_morphology_feature_value()` helper. Likewise, `3VOLUME_CONVEXHULL` is covered by `TEST_3D_MORPHOLOGY_VOLUME_CONVEX_HULL_MATLAB` while `TEST_3D_MORPHOLOGY_VOLUME_CONVEX_HULL_REGRESSION` checks the same point. All four tests use `ut_inten.nii + ut_mask57.nii`, label 57, `GREYDEPTH=128`, and `IBSI=true`; there is no distinct production-only configuration for the regression tests to protect. Per the oracle-vs-regression convention, please remove the two regression functions/registrations. I would recommend quarantining them but, Honestly these "backup" tests are NOT worth the hassle.
> 2. **Remove the dead `3MESH_VOLUME` regression golden.** `morphology_3d_regression_ref_vals` contains `{"3MESH_VOLUME", 478516}`, but there is no `test_3d_morphology_mesh_volume_regression()` and no corresponding GoogleTest registration. The only executable assertion for this feature is the MATLAB test, and the coverage CSV already records `Regression=N`.
> 3. **Complete MATLAB golden provenance.** The comment above `morphology_3d_matlab_ref_vals` explicitly says `generated = offline; no in-repo generator`. That leaves all three MATLAB goldens noncompliant with SPEC §6.4, which requires tool, version, exact configuration, and a checked-in runnable generator path at the definition site. Please add the MATLAB/Octave generator script and exact invocation/configuration (or reclassify these values if they cannot be regenerated).
> 4. **Cite the 5% method-divergence band.** The comments and MIRP report measure and explain the approximately 3.6% hull difference, but SPEC §7 requires a stated band **plus a citation** for known method divergence. Please cite the MATLAB `regionprops3` `ConvexVolume` definition and the documented Nyxus/IBSI/MIRP hull or mesh-volume convention used to justify applying 5% to `3VOLUME_CONVEXHULL` and `3MESH_VOLUME`.

1. Keeping the two regression assertions for now — Regression=Y means a drift snapshot alongside the oracle assertion, which is the convention we settled on (see @sameeul's note on the same point in #431). It's a cross-family question rather than a 3D-morphology one, so I've filed it as a single decision to make once rather than resolving it differently per PR.

2. Fixed — the 3MESH_VOLUME regression golden is gone. Good catch: the file defines eight regression tests and the table carried nine entries, so that one was never read by anything.

3. MATLAB can't be re-run from this tree (no licence, and Octave's image package has no regionprops3), so rather than claim a generator that cannot exist, I gave both quantities a reference from a tool that is runnable here. MIRP's morph_vol_approx reproduces 3VOXEL_VOLUME exactly (274432), and its convex-hull volume — backed out of IBSI's volume density (convex hull) = V_mesh/V_convex — gives 496958.32 against regionprops3's 497824. Both are pinned in test_3d_morphology_mirp.h, asserted, and re-verified by gen_morphology3d_mirp.py on every run; the two rows now read oracle=mirp with the MATLAB assertion kept as the second opinion. The MATLAB session still isn't repeatable, but the values it produced are now reproducible from the tree.

4. Cited, and with a second measurement behind it: MIRP's hull and MATLAB's agree with each other to 0.17% while Nyxus sits 3.7% below both, so the band covers a discrete voxel hull against two independently triangulated ones — a definitional difference, not tool noise. That is now stated at the band's definition site in both files and in the MIRP report.

vjaganat90:
> > 3. **Complete MATLAB golden provenance.** The comment above `morphology_3d_matlab_ref_vals` explicitly says `generated = offline; no in-repo generator`. That leaves all three MATLAB goldens noncompliant with SPEC §6.4, which requires tool, version, exact configuration, and a checked-in runnable generator path at the definition site. Please add the MATLAB/Octave generator script and exact invocation/configuration (or reclassify these values if they cannot be regenerated).
> 3. MATLAB can't be re-run from this tree (no licence, and Octave's image package has no regionprops3), so rather than claim a generator that cannot exist, I gave both quantities a reference from a tool that is runnable here. MIRP's morph_vol_approx reproduces 3VOXEL_VOLUME exactly (274432), and its convex-hull volume — backed out of IBSI's volume density (convex hull) = V_mesh/V_convex — gives 496958.32 against regionprops3's 497824. Both are pinned in test_3d_morphology_mirp.h, asserted, and re-verified by gen_morphology3d_mirp.py on every run; the two rows now read oracle=mirp with the MATLAB assertion kept as the second opinion. The MATLAB session still isn't repeatable, but the values it produced are now reproducible from the tree.

so they should move to mirp_refvals no? We just need agreement with one oracle for every feature value. It's one thing to have backup regression guard but two oracle tests is not spec compliant. What happens if there's a bug in the future version of one of the oracle, which oracle value would you believe?

tests/test_3d_morphology_mirp.h
```diff
@@ -0,0 +1,143 @@
+#pragma once
+
+#include <cmath>
+#include <string>
```
vjaganat90: test_ref_vals.h already includes `#include <string>`

tests/test_all.cc
```diff
@@ -531,6 +532,45 @@ TEST(TEST_NYXUS, TEST_3D_MORPHOLOGY_MESH_VOLUME_MATLAB) {
 	ASSERT_NO_THROW(test_3d_morphology_mesh_volume_matlab());
 }
 
+TEST(TEST_NYXUS, TEST_3D_MORPHOLOGY_VOXEL_VOLUME_MATLAB) {
```
vjaganat90: like I mentioned, one oracle suffices. It would be preferable just from a pure operational POV to prefer an open source oracle.

PR #437, opened 2026-08-17
test(gabor): vet both GABOR default configs against skimage, not just the compiled-in one

darkclad:
The family's single vetted row claimed scikit-image on a golden table whose header described the documented defaults (gabor_freqs=[4,16,32,64], gabor_thetas=[0,45,90,135]). Running skimage fresh at exactly that recipe does not reproduce those goldens: they come from the compiled-in GaborFeature::f0_theta_pairs default, which stores its pairs in the opposite order from the one the option parser builds and every consumer reads. The two configs differ by up to 0.84 absolute on the same ROI, and the one the Python API runs had no in-tree assertion at all.

Both config points are now pinned and asserted:

  gabor.cpp_static_defaults    f0={0,pi/4,pi/2,3pi/4}, theta={4,16,32,64} rad
                               TEST_2D_GABOR_SKIMAGE (the existing 16 values, unchanged)
  gabor.documented_defaults    f0={4,16,32,64}, theta={0,pi/4,pi/2,3pi/4} rad
                               TEST_2D_GABOR_DOCUMENTED_DEFAULTS_SKIMAGE (new, 16 values)

Every one of the 32 pins was re-derived from a fresh scikit-image 0.26.0 run (max |diff| 0.000e+00), and the kernel mapping was cross-checked against the canonical closed form at all seven non-degenerate (f0, theta) points (<= 1.1e-16).

The order mismatch in gabor.cpp is a defect in the feature, not in the test. It changes public values, so it is measured and documented here and left to its own branch.

Also in this pass:

- test_gabor_truth.h deleted; its table moved next to the assertions that read it, renamed gabor_2d_skimage_cpp_defaults_ref_vals and declared through the SPEC 6.3.1 ref_vals_list alias. As an exempt "fixture" file it was outside both the naming rule and the table check, so a table of oracle goldens sat unchecked; SPEC 6.1 now says the exemption covers fixtures, not goldens.
- Include hygiene in test_2d_gabor_skimage.{h,cc}: dropped <fstream> and features/intensity.h (unused) and the commented-out misspelled gtest include; added the headers the file actually uses (featureset.h, feature_settings.h, roi_cache.h, test_dsb2018_data.h, test_ref_vals.h).
- gen_gabor_skimage.py now parses both tables out of the .cc and re-verifies every pinned value, and reports 1/baseline_count per ROI - the resolution limit that justifies the rel=1e-3 band (4.5e-3 .. 1.1e-2, i.e. four to eleven times the tolerance).
- The scope of "vetted vs skimage" is measured, not asserted: scoring the same feature off skimage's own filtering (skimage.filters.gabor, untruncated support) diverges by up to 1.0, because gamma=0.1 makes the analytic kernel far wider than the 16x16 window - 7.9% of its L1 mass survives the crop at the lowest frequency. The crop and the zero-padded full convolution are recorded as part of both recipes, and generator part D reprints the mass table each run.
- Registry: one GABOR row became two, one per config recipe; config_recipe ids replace the generic "not mode-specific" boilerplate; current_test drops test_nyxus.py (its two GABOR tests compare GPU against CPU and validate parameter parsing - neither asserts an oracle value); target_test cleared; source audit -> in-tree.
- config_recipes.md defines both recipes; TOOLS.md records that skimage needs no env of its own (nyxus_mirp already carries it) plus the gabor_kernel gotchas.

Audit artifacts under tests/vetting/audit/: gabor_2d_coverage.csv, gabor_2d_skimage_vetting_report.md, gabor_2d_golden_regen.md.

Verified: fresh skimage rerun of every pin (generator exit 0), gtest 778/778 on Windows (MSVC, Release), 85 passed / 1 skipped in pytest tests/python (the 7 arrow failures are the known ALLEXTRAS=OFF build, not a regression), check_test_names.py and check_coverage.py clean, and both negative controls run: perturbing a pin fails the gtest, and perturbing it in the .cc fails the generator. ASan/UBSan on Linux: 774/774, exit 0, no sanitizer reports.


vjaganat90 (changes requested):
This cleanly exposes the static/default mismatch, fixes the `agrees_gt` argument order, and moves regenerated goldens beside their assertions.

I think the direction is right; these claims just need tightening:

1. **Name the behavior being tested, not the unresolved contract.** `tests/test_2d_gabor_skimage.cc:64-96` injects 4/16/32/64 as raw angular `f0`, but `src/nyx/python/nyxus/nyxus.py:97-104` documents denominators of π and `src/nyx/environment.cpp:219-228` gives another CLI default. `src/nyx/cli_gabor_options.cpp:32-38` stores the raw numbers while `src/nyx/features/gabor.cpp:107-112,393-398` consumes them as angular frequency, so please rename this recipe (for example `gabor.python_raw_defaults`) or first resolve the units contract.

2. **Narrow the scikit-image oracle claim.** `tests/vetting/oracles/gen_gabor_skimage.py:19-29,151-164` says the score pipeline is Nyxus's own reimplementation, and `:44-49` says `f0=0` is analytic rather than from scikit-image; nevertheless `tests/vetting/audit/gabor_2d_coverage.csv:2` labels both complete configs as `skimage`. Please classify this as hybrid/kernel-level, split the analytic point, or use WND-CHARM for a full-feature oracle.

3. **Make the native-gabor negative control reproducible.** `tests/vetting/config_recipes.md:82-87` says generator Part D reruns `skimage.filters.gabor()`, but `tests/vetting/oracles/gen_gabor_skimage.py:61-64,216-236` imports only `gabor_kernel` and prints retained kernel mass. Please add the stated score comparison or trim that reproducibility claim.

Minor:
- `tests/test_2d_gabor_skimage.cc:135-147` skips all GPU value assertions, while `tests/test_all.cc:722-727` runs it under the skimage oracle test name; move this size-only path to a mechanics test.
- `tests/test_2d_gabor_skimage.cc:162-169` manually restores process-global `f0_theta_pairs`; an RAII guard would prevent state leakage on an exception.
- `tests/vetting/SPEC.md:203-204,348-349` requires the family matrix and benchmark registry; `tests/vetting/matrix/gabor.md` and `tests/vetting/benchmarks.md` are still missing. Please also regenerate the conflicted `tests/vetting/coverage_report.md`.

Once the oracle scope and naming are tightened, the additional coverage here will be valuable.

PR #438, opened 2026-08-20
test(2d gldm): vet the family against PyRadiomics, and band the entropy it exposed

darkclad:
## What was cleaned up

- **One fixture for the family.** `test_2d_gldm_common.h` holds the settings recipe and the four-slice phantom averaging that the ibsi, regression and mechanics files each carried their own copy of. Fixtures only, no reference data (SPEC §6.3.1). Two of those copies sized slice 2's and slice 3's mask with `sizeof` of the **intensity** array; all the arrays are 20 entries, so the counts happened to agree.
- **Both golden tables are `const` and read through `.at()`** behind a `count()` guard. They were non-`const` and read with `operator[]`, which default-inserts `0.0` for a missing key and then compares against it with a tolerance of `0/frac = 0`.
- **`../src/nyx/features/pixel.h` leaves all four files** — nothing in them named a symbol from it; `NyxusPixel` comes from `test_data.h`. The three remaining nyxus headers arrive through the new `_common.h`, which is the only file that spells them.
- **Stale cross-references fixed.** Both mechanics files pointed at `test_gldm_oracle.{h,py}`, names nothing in the tree has carried for some time.
- **Every row's `current_test` named `test_3d_gldm_regression.h` and `test_3d_gldm_pyradiomics.h`** — 3D files, on 2D rows, for a different implementation (`3d_gldm.cpp`). The first is additionally one of the headers `test_all.cc` never includes, so it runs nothing at all.

## What was measured, and banded

**This PR changes no C++ source.** `git diff <merge-base> -- src/` is empty.

**`GLDM_DE` is computed with an approximate logarithm** and misses its reference by **7.9e-4** on the phantom mean and up to **1.3e-3** per slice, while every other feature in the family reproduces PyRadiomics to 1e-16. `GLDMFeature::calc_DE()` reads `Nyxus::fast_log10`, a *float* second-order polynomial approximating `log2` over `[0.75, 1.5)`; measured against `log2` across that reduction range its worst absolute error is **8.9e-3**, at x = 0.75.

| slice | PyRadiomics | Nyxus (`fast_log10`) | rel |
|---|---|---|---|
| z1 | 3.0464393446710125 | 3.0425443887710575 | **1.3e-3** |
| z2 | 2.3789919869000604 | 2.376873304969386 | 8.9e-4 |
| z3 | 2.6098501660289402 | 2.6088697770062619 | 3.8e-4 |
| z4 | 2.8218881955261761 | 2.820316738552517 | 5.6e-4 |
| mean | 2.7142924232815471 | 2.7121510523248058 | 7.9e-4 |

Reproduced independently of Nyxus: re-summing PyRadiomics' own dependence matrix with the `helpers.h` polynomial transcribed into Python — float arithmetic preserved operation by operation — lands on 2.7121510523248058, Nyxus' value to the last bit. The residual is the approximation and nothing else; the dependence matrix, which is what this family implements, is exact.

**`GLDM_DE` therefore asserts at `rel=2.5e-3`, twice its measured worst, per SPEC §7's documented-residual tier.** The other 13 features assert at `rel=1e-9`. This follows #433, which named `fast_log10` for the same reason and banded its five GLCM entropy features at twice their measured residuals: one helper, one convention across every family that reads it, and one change when it changes.

An earlier revision of this PR took the source fix instead — one line, `std::log2(p)` for `fast_log10(p + EPS) / LOG10_2`, which reproduces PyRadiomics at 9.8e-16. It was withdrawn on review, because it moves a public feature value and that does not belong in an oracle PR. For the record, that value is 2.7142924232815497 — the same number `NGLDMfeature`'s dependence-count entropy already produces and the same one mirp's `ngl_dc_entr` is pinned at in `test_2d_ngldm_mirp.h`.

**The band is not a free pass.** The other 13 features hold `rel=1e-9` against the same dependence matrix, per slice and on the mean, so a real error in the matrix fails on 13 assertions at once. Verified non-vacuous: tightening `GLDM_DE` back to `rel=1e-9` fails with `abs(3.04254 - 3.04644) = 0.00389 > 3.05e-09`.

*Left alone deliberately:* `fast_log10` is also read by `glcm.cpp`, `glrlm.cpp`, `glszm.cpp` and their four 3D counterparts. Each carries the same error into its family's entropies and belongs to that family's re-vetting.

## What was promoted

**All 14 features, `oracle=ibsi` → `oracle=pyradiomics`, `rel=1e-2` → `rel=1e-9`** (`rel=2.5e-3` for `GLDM_DE`).

PyRadiomics defines GLDM and names all 14 features. On the four IBSI phantom slices at `binWidth=1` (identity binning on the integer phantom), `gldm_a=0`, `distances=[1]`, `force2D`, it reproduces Nyxus to **1.6e-16** worst case over the 13 exact-tier features.

**Per slice, not just the mean.** The phantom is four slices, each its own ROI with its own scalar; IBSI's "2D, averaged" aggregation publishes only their mean, so the assertion compared one average against another — which passes whenever per-slice errors cancel. PyRadiomics computes each slice on its own (the averaging is the generator's), so `gldm_2d_pyradiomics_slice_ref_vals` pins all 14 × 4 values alongside the 14 means, and every assertion checks both. Measured per slice over the 13 exact-tier features: 8 exact on all four, worst 2.2e-16. `GLDM_DE`'s per-slice spread is wider than its mean (1.3e-3 against 7.9e-4), which is exactly the case the mean cannot see — and why its band is sized on the per-slice worst. **Verified non-vacuous:** moving `GLDM_LDE_z1` +0.001 and `GLDM_LDE_z3` −0.001 leaves the mean unchanged to its last digit; the per-slice assertion fails and names `GLDM_LDE_z1`, while `test_2d_gldm_ibsi.h` (mean only) passes.

The IBSI file stays. Its goldens are the reference manual's **NGLDM** values under GLDM names, and that attribution is right — a Nyxus GLDM dependence count is `1 +` the number of 8-neighbours sharing the centre's binned grey level, which is IBSI's `j = k + 1` at alpha=0, d=1, so the two families are the same statistic. The identity is measured rather than assumed: 13 of Nyxus' 14 values are bit-identical to the mirp-vetted NGLDM goldens already pinned in `test_2d_ngldm_mirp.h`, and the fourteenth differs by exactly the `fast_log10` residual above — two implementations disagreeing about arithmetic, not two families disagreeing about the statistic. Published to three significant figures, those values can only ever assert at `rel=1e-2`, so the two files are complementary: **IBSI fixes the definition, PyRadiomics fixes the digits.**

## What was demoted

**`tests/python/test_2d_gldm_pyradiomics.py` → `test_2d_gldm_mechanics.py`.** Its three pinned numbers *are* PyRadiomics output and reproduce exactly; what does not hold is the comparison. It ran Nyxus at the production default (`ibsi=false`, MATLAB grey binning at `coarse_gray_depth=64`) against a PyRadiomics run binning at `binWidth=1`, so the two built their dependence matrices over different level assignments:

| feature | PyRadiomics | Nyxus | rel |
|---|---|---|---|
| GLDM_LDE | 1.88311688312 | 2.06493506494 | **0.088** |
| GLDM_DV | 0.263282172373 | 0.344408837915 | **0.24** |
| GLDM_LGLE | 0.0160307720963 | 0.0081913915026 | **0.96** |
| GLDM_SDLGLE | 0.0136053972358 | 0.00653737856421 | **1.08** |
| GLDM_LDHGLE | 2317.33766234 | 3165.14285714 | **0.27** |

(all 14 in the report). The file's `rel=0.20`/`rel=0.15` bands passed all three assertions anyway. SPEC §5 does not let a tolerance absorb a configuration mismatch, so this was not an oracle test at any band. Its real subject — the bug #14b background-pollution guard on the production path, which the IBSI-mode oracle tests cannot reach — is a SPEC §2 mechanics test, re-pinned against Nyxus' own values at `rel=1e-9` with the `GLDM_LDE < 5.0` bound that fails on the pre-fix inflated ~16.6. No registry row changes `status`: the family is vetted on the phantom in IBSI mode, which is a different config point.

## What the review changed

- **The source fix is gone.** `calc_DE()` is byte-identical to `main`; `GLDM_DE`'s public value does not move, and there is no prerequisite PR to depend on. The commit that made the change is dropped from the branch rather than left with a revert on top.
- **The mean now averages the slices the oracle just asserted.** `assert_gldm_feature_against_golden_values()` took a feature id and featurised the four phantom slices itself, so every PyRadiomics assertion computed them twice and averaged the second run — the mean and the per-slice checks were reading different featurisations, and only determinism kept them agreeing. It is now `assert_gldm_mean_against_golden_values()`, taking the slice vector, with a guard against averaging an empty one.
- **SPEC §5.1 and §6.3 completed.** `tests/vetting/matrix/gldm.md` added: GLDM has no angle set, no offset and no alpha/distance knob, and its two remaining axes are coupled because `calculate()` overwrites `GREYDEPTH` with `0` under `IBSI=true`, so the cross-product is four points, not a grid. The radiomics-binning path is recorded NOT MEASURED rather than omitted. `benchmarks.md`'s existing `ibsi_digital_phantom` entry now names the `gldm.ibsi_phantom_2d` recipe and this family's tests, and all 14 registry rows fill the `benchmark` and `test_name` columns.

## New audit artifacts

`tests/vetting/audit/`: `gldm_2d_coverage.csv`, `gldm_2d_pyradiomics_vetting_report.md`, `gldm_2d_ibsi_vetting_report.md`, `gldm_2d_golden_regen.md`, `scan_gldm_coverage.py`. `tests/vetting/oracles/gen_gldm_pyradiomics.py` prints both tables and **re-verifies every pin in the header it feeds**, exiting non-zero on a mismatch or on a pin it cannot produce. Recipe `gldm.ibsi_phantom_2d` added to `config_recipes.md`; config matrix in `matrix/gldm.md`; the pytest finding recorded in `not_covered.md`; the `force2D` setup gotcha added to `TOOLS.md`.

## How it was verified

Rebased onto current `main` and re-run from a fresh build:

- `gen_gldm_pyradiomics.py` (pyradiomics v3.0.1, env `nyxus_oracle`): **70 verified, 0 failed, 0 unproducible** — both tables, 14 means and 56 per-slice values
- gtest **860/860** on Windows
- `pytest tests/python/` **88 passed, 1 skipped**; the 7 `@pytest.mark.arrow` failures are a tiff-only local build (`USE_ARROW=OFF`), not this branch
- `check_test_names.py` 0 violations; `check_coverage.py --check` clean and `--write` moves nothing; `scan_gldm_coverage.py` idempotent over 14 rows; `test_vetting_mechanics` passes
- `oracle_coverage.csv` parses clean at 761 rows × 16 columns, no ragged rows
- **No sanitizer run, because there is nothing to sanitize:** this branch changes no C++ source. The ASan/UBSan results quoted in the previous revision of this description covered the withdrawn source fix and no longer apply.


src/nyx/features/gldm.cpp
```diff
@@ -440,8 +441,11 @@ double GLDMFeature::calc_DE()
 	{
 		for (int j = 1; j <= Nd; j++)
 		{
-			double entrTerm = fast_log10(P.matlab(i, j) / double(Nz) + EPS) / LOG10_2;
-			f += P.matlab(i, j) / double(Nz) * entrTerm;
+			// log2 of the exact probability: an empty cell contributes nothing, so no epsilon
```
vjaganat90: The exact-log change is mathematically sound, but it also changes the public `GLDM_DE` value. Please split `5164e1a` into the prerequisite source-fix PR used by this vetting series, then rebase this oracle PR on it.

tests/test_2d_gldm_pyradiomics.h
```diff
@@ -0,0 +1,184 @@
...
+    {
+        const std::string key = feature_name + "_z" + std::to_string (z + 1);
+        SCOPED_TRACE ("pyradiomics " + key);
+        ASSERT_TRUE (gldm_2d_pyradiomics_slice_ref_vals.count(key) > 0) << key;
+        ASSERT_TRUE (agrees_gt (per_slice[z], gldm_2d_pyradiomics_slice_ref_vals.at(key),
+                                gldm_2d_pyradiomics_frac_tolerance)) << key;
+    }
+
+    // then the four-slice mean, the quantity IBSI publishes
+    assert_gldm_feature_against_golden_values (feature_, feature_name, gldm_2d_pyradiomics_ref_vals,
```
vjaganat90: `per_slice` is already computed and asserted above, but this helper recomputes all four slices. Please derive the mean from that vector (or pass it into the helper) so the mean covers the exact values just checked.

tests/vetting/config_recipes.md
```diff
@@ -120,6 +120,31 @@ chosen reference tool (SPEC 5). Oracle tests reference a recipe by id; this file
   `FRACT_DIM_PERIMETER` is cross-method — Nyxus uses a divider walk against FracLac's edge box
   count — so it carries a stated ~3% band, per SPEC 7's "known method divergence" row.
 
+## gldm.ibsi_phantom_2d
```
vjaganat90, line 123: This adds the recipe, but SPEC §5.1 still requires `tests/vetting/matrix/gldm.md`, and §6.3 requires the IBSI phantom in `tests/vetting/benchmarks.md`. Please add both so this family migration is SPEC-complete.

src/nyx/features/gldm.cpp
darkclad: Fixed

tests/test_2d_gldm_pyradiomics.h
darkclad: Fixed

tests/vetting/config_recipes.md
darkclad: Fixed

PR #439, opened 2026-08-20
test(2d gldzm): vet the family against mirp, and fix the zone connectivity it was hiding

darkclad:
Next family in the vetting series (after #438, 2D GLDM). Seventeen of the eighteen 2D GLDZM rows
have read `oracle=mirp` with `target_test=test_2d_gldzm_mirp.h` since the tracker was imported —
**a file that did not exist**. Running that oracle for the first time found Nyxus disagreeing with
it on **all 18 features**, by 1.3% to 60%, with every test green.

## What was fixed

**GLDZM zones were 4-connected.** `GLDZMFeature::calculate` grew a zone by testing East, South,
West, North. A GLDZM zone is 8-connected — two same-level pixels touching only at a corner are one
zone. That is what IBSI defines, what mirp implements, and what **Nyxus' own GLSZM already used**
(`glszm.cpp` tests E/SE/S/SW, the forward half of the 8-neighbourhood). The two families disagreed
inside the same codebase about what a zone is.

| feature        | Nyxus (before) | mirp           | rel            |
| -------------- | -------------- | -------------- | -------------- |
| GLDZM_ZDV      | 0.081632653061 | 0.051020408163 | **0.60** |
| GLDZM_LDLGLE   | 0.500413359788 | 0.386078042328 | **0.30** |
| GLDZM_GLNU     | 1.642857142857 | 1.414285714286 | **0.16** |
| GLDZM_ZP       | 0.265389576883 | 0.240389576883 | **0.10** |
| GLDZM_LDE      | 1.321428571429 | 1.214285714286 | 0.088          |
| …the other 13 |                |                | 0.013–0.068   |

**The published IBSI values sided with mirp.** The file's own table pins `GLDZM_SDE` at the manual's
`0.946`; mirp gives 0.94643 and Nyxus gave 0.91964. Two independent references agreed with each
other and disagreed with Nyxus, so this was never a mirp-configuration artifact.

**Every disagreement lived in one slice.** Phantom slices z2, z3 and z4 reproduced mirp exactly,
digit for digit; z1 — the only slice with two same-level pixels touching at a corner and nowhere
else — carried all of it. Two diagonal links split five true zones into seven:

|                      | zones on z1 | GLDZM_ZP | GLDZM_GLNU |
| -------------------- | ----------- | -------- | ---------- |
| 4-connected (before) | 7           | 0.35     | 2.714      |
| 8-connected (after)  | 5           | 0.25     | 1.800      |
| mirp                 | 5           | 0.25     | 1.800      |

With the four diagonals, Nyxus reproduces mirp on all 16 mirp-exposed features **on every slice** to
a worst residual of **8.9e-16**, and matches all fourteen published IBSI values.

*Left alone deliberately:* `3d_gldzm.cpp` has the identical four-direction loop and therefore the
same defect (in 3D that is 26-connectivity, so the fix is not a copy-paste). It is a separate family
in the re-vetting order; changing it here would move pinned goldens this PR does not own. Filed with
its measurement as item 17 in the project's deferred-defects list.

**On shipping a source fix inside a vetting PR.** The standing rule in this series is that a change
to a public feature value gets its own branch — that is where the `fast_log10` entropy defect, the
`3MESH_VOLUME` alias and the 3D angle-slot mislabel all sit. This one is deliberately different, and
the difference is that those three leave their families vettable with a stated band while this one
does not: at 4-connectivity Nyxus matches no reference on any of the 18 features, so there is no
honest tolerance to ship. The alternative is demoting all 18 rows to `regression` and filing the
defect, which is a worse outcome for the same work. Happy to split it into a fix branch plus a
vetting branch on top if you would rather review it that way — commit 1 is already exactly that
split.

## The two things that should have caught it

**The tolerance was ±50%.** `agrees_gt(aveTotal, reference, 2.)` — `agrees_gt` divides the golden by
its third argument, so `2.` is a band of ±50%, about 40× looser than the worst error it was covering.
Same defect the 2D NGLDM re-vet fixed; the sibling family was never swept. Now `rel=1e-2` on the IBSI
file (set by its three significant figures) and `rel=1e-9` on the mirp file.

**The published-matrix test was encoding the bug.** `TEST_2D_GLDZM_MATRIX_CORRECTNESS_IBSI` compares
against `ibsi_fig3_17c_gldzm_reference_matrix` in `test_data.h` — which is wrong independently of
this fix: it summed **9 zones over a 16-pixel ROI**, and mirp finds 8. mirp pins that fixture three
ways over, and all three agree with the 8-connected reading:

| mirp on fig. 3.17a | value | implies                                  |
| ------------------ | ----- | ---------------------------------------- |
| `dzm_z_perc_2d`  | 0.5   | 8 zones over 16 pixels                   |
| `dzm_glnu_2d`    | 2.25  | per-grey zone counts 3, 1, 2, 2          |
| `dzm_zdnu_2d`    | 6.25  | 7 zones at distance 1, one at distance 2 |

The grey-level-2 row read `2, 0` and is `1, 0`. Corrected with that arithmetic recorded beside it.

Note the source fix alone does **not** turn the suite red — the ±50% band absorbs it. That is the
whole problem in one sentence.

## What was promoted

**16 features vetted against mirp at `rel=1e-9`**, in the `test_2d_gldzm_mirp.h` that seventeen rows
had targeted all along.

**Per slice, not just the mean.** The phantom is four slices, each its own ROI; IBSI's "2D, averaged"
aggregation publishes only their mean, so the assertion compared one average against another — which
passes whenever per-slice errors cancel. This family is the argument for taking that seriously: the
defect lived entirely in z1 and the mean diluted it by four. `gldzm_2d_mirp_slice_ref_vals` pins all
16 × 4 values alongside the 16 means, and every assertion checks both.

*Verified non-vacuous:* moving `GLDZM_HGLZE_z1` +0.002 and `GLDZM_HGLZE_z3` −0.002 leaves the mean
unchanged to its last digit; the per-slice assertion fails and names `GLDZM_HGLZE_z1`, while
`..._IBSI` (mean only) passes.

## What was demoted

**The IBSI table held four Nyxus snapshots.** Fourteen entries were 3-significant-figure published
values; `GLDZM_SDLGLE`, `GLDZM_GLM`, `GLDZM_ZDM` and `GLDZM_ZDV` carried **sixteen digits** and
matched Nyxus' own pre-fix output exactly. The manual does not publish at that precision — that is
what identifies them — and two were wrong: mirp gives `SDLGLE` 0.3674768518518518 (5.3% away) and
`ZDV` 0.05102040816326531 (60% away). Per SPEC §6.3.1 a table mixing published values with snapshots
has no honest oracle for its name, so it is split:

- `SDLGLE` and `ZDV` → vetted against mirp at the exact tier; the IBSI file never held a published
  value for either.
- `GLM` and `ZDM` → `status=regression`, drift-pinned in `test_2d_gldzm_regression.h`. Neither is an
  IBSI GLDZM feature and mirp exposes no column for either, so nothing independent covers them.
  `GLDZM_ZDV` was also the one row reading `oracle=ibsi`, citing that snapshot as its evidence.

## Also cleaned up

- `test_2d_gldzm_common.h` holds the settings recipe and four-slice scaffolding the family's files
  each copied. Fixtures only, no reference data (SPEC §6.3.1). The copy in the IBSI file sized every
  slice's mask with `sizeof` of the **intensity** array; all arrays are 20 entries, so the counts
  happened to agree.
- Both golden tables are `const`; `<unordered_map>` leaves the IBSI file (`test_ref_vals.h`, two
  lines below, already supplies it); the `VERIFIABLE_WITH_3P_BUILTIN_ORACLE__` trace prefix becomes
  the oracle token actually compared against.
- Every row's `current_test` had named `test_3d_gldzm_regression.h` — a 3D file, on 2D rows, for a
  different implementation. Dropped.

## New audit artifacts

`tests/vetting/audit/`: `gldzm_2d_coverage.csv`, `gldzm_2d_mirp_vetting_report.md`,
`gldzm_2d_ibsi_vetting_report.md`, `gldzm_2d_golden_regen.md`, `scan_gldzm_coverage.py`.
`tests/vetting/oracles/gen_gldzm_mirp.py` prints both tables and **re-verifies every pin in the
header it feeds**. Recipe `gldzm.ibsi_phantom_2d` added to `config_recipes.md`; the split-table
finding recorded in `not_covered.md`.

## How it was verified

Fresh rebuild and fresh oracle reruns, not just "tests pass":

- `gen_gldzm_mirp.py` (mirp 2.6.0, env `nyxus_mirp`): **80 verified, 0 failed, 0 unproducible** —
  16 means and 64 per-slice values
- gtest **816/816** on Windows (802 before: −4 IBSI cases whose goldens left the table, +16 mirp,
  +2 regression)
- ASan+UBSan on Linux, gtest: **812/812**, exit 0, no sanitizer diagnostics
- ASan on Linux, `pytest tests/python/`: **85 passed, 1 skipped**, no sanitizer diagnostics
- `pytest tests/python/` on Windows **85 passed, 1 skipped**; the 7 `@pytest.mark.arrow` failures are
  the tiff-only local build, not this branch
- `check_test_names.py --check` 0 violations, `check_coverage.py --check` clean,
  `scan_gldzm_coverage.py --check` clean over 18 rows
- all 82 pinned goldens checked against the bounds the features hold by construction
  (`SDE`/`ZP`/`GLNUN`/`ZDNUN` in [0,1], `LDE`/`ZDM` ≥ 1, normalised ≤ raw, `LDE ≥ SDE` per
  slice) — no violations


sameeul (approved):

src/nyx/features/gldzm.cpp
```diff
@@ -210,6 +210,91 @@ void GLDZMFeature::prepare_GLDZM_matrix_kit (SimpleMatrix<unsigned int>& GLDZM,
 					continue;
 				}
 
+				// A zone is 8-connected: two same-level pixels touching only at a corner belong to one
```
vjaganat90, line 213: This changes every public 2D GLDZM value. Please keep the series boundary used elsewhere: move `089b50f` to a prerequisite source-fix PR, then rebase this vetting PR on it.

tests/test_data.h
```diff
@@ -256,10 +256,16 @@ const static NyxusPixel ibsi_fig3_17a_gldzm_sample_image_mask[] =
 	{1, 4, 1}, {2, 4, 1}, {3, 4, 1}, {4, 4, 1}
 };
 
+// Rows are the four grey levels, columns the two zone distances, and the table sums to 8 zones
+// over the 16-pixel ROI. Grey level 2 holds ONE zone: its five pixels are linked through the
+// (2,1)-(3,2) diagonal, and a GLDZM zone is 8-connected. mirp 2.6.0 on this fixture pins the
```
vjaganat90: These MIRP scalars constrain the total and row/column marginals, but do not uniquely determine the 4x2 matrix or which grey level owns the distance-2 zone. Please call this corroboration and support the exact cells with direct matrix output or an explicit analytic zone enumeration.

tests/test_2d_gldzm_mirp.h
```diff
@@ -0,0 +1,200 @@
...
+    {"GLDZM_GLV_z3", 4.222222222222222}, {"GLDZM_GLV_z4", 4.222222222222222},
+
+    {"GLDZM_ZDV_z1", 0.0}, {"GLDZM_ZDV_z2", 0.20408163265306123},
+    {"GLDZM_ZDV_z3", 0.0}, {"GLDZM_ZDV_z4", 0.0},
+
+    {"GLDZM_ZDE_z1", 1.5219280948873621}, {"GLDZM_ZDE_z2", 2.2359263506290326},
+    {"GLDZM_ZDE_z3", 1.5849625007211563}, {"GLDZM_ZDE_z4", 1.5849625007211563}
+};
+
+// rel=1e-9: agrees_gt divides the golden by this, so a larger argument is a tighter band
```
vjaganat90: `rel=1e-9` is not SPEC §7's `exact` tier; that row is `abs 1e-9`. Either use an absolute `ASSERT_NEAR(..., 1e-9)` here or stop labeling this relative check exact.

tests/test_2d_gldzm_mirp.h
```diff
@@ -0,0 +1,200 @@
...
+    {
+        const std::string key = feature_name + "_z" + std::to_string (z + 1);
+        SCOPED_TRACE ("mirp " + key);
+        ASSERT_TRUE (gldzm_2d_mirp_slice_ref_vals.count(key) > 0) << key;
+        ASSERT_TRUE (agrees_gt (per_slice[z], gldzm_2d_mirp_slice_ref_vals.at(key),
+                                gldzm_2d_mirp_frac_tolerance)) << key;
+    }
+
+    // then the four-slice mean, the quantity IBSI publishes
+    assert_gldzm_feature_against_golden_values (feature_, feature_name, gldzm_2d_mirp_ref_vals,
```
vjaganat90: `per_slice` is already computed and asserted above, but this helper reruns all four slices. Please derive the mean from that vector (or pass it into the helper) so the mean covers the exact values just checked.

tests/vetting/config_recipes.md
```diff
@@ -120,6 +120,26 @@ chosen reference tool (SPEC 5). Oracle tests reference a recipe by id; this file
   `FRACT_DIM_PERIMETER` is cross-method — Nyxus uses a divider walk against FracLac's edge box
   count — so it carries a stated ~3% band, per SPEC 7's "known method divergence" row.
 
+## gldzm.ibsi_phantom_2d
```
vjaganat90, line 149: This recipe is present, but SPEC §5.1/§6.3 still require `tests/vetting/matrix/gldzm.md` and the IBSI phantom in `tests/vetting/benchmarks.md`. Please add both for a complete family migration.

vjaganat90 (changes requested):

src/nyx/features/gldzm.cpp
sameeul: This is fine IMO

tests/test_data.h
darkclad: Fixed

tests/test_2d_gldzm_mirp.h
darkclad: Fixed

tests/test_2d_gldzm_mirp.h
darkclad: Fixed

tests/vetting/config_recipes.md
darkclad: Fixed

vjaganat90 (approved):
LGTM

PR #440, opened 2026-08-20
test(2d glszm): vet the family against mirp, per slice and averaged

darkclad:
Ten of the sixteen 2D GLSZM registry rows have claimed `oracle=pyradiomics` with
`target_test=test_2d_glszm_pyradiomics.h` since the tracker was imported — **a file that does not
exist**. The other six claimed `ibsi`, against a file that does and asserts all sixteen features at
±1%. No independent implementation had ever been run against this family.

This PR runs two. **mirp 2.6.0 and PyRadiomics 3.0.1 agree with each other to 7.0e-16** on all 80
quantities, and both reproduce Nyxus **bit for bit on fifteen of the sixteen features**, so those
move to `rel=1e-9`. The sixteenth, `GLSZM_ZE`, misses *both* by 2.5e-3 — the `fast_log10`
approximation, measured and banded here rather than fixed, because fixing it changes a public
feature value.

Only mirp's values are pinned; PyRadiomics is the corroboration that says so. Running the claimed
oracle mattered: swapping in mirp alone would have closed the row while leaving the question of
whether the two tools agree unasked — and that question is exactly what shows the one disagreement
is Nyxus' logarithm and not a mirp convention.

## What was promoted

All sixteen rows go from `oracle=pyradiomics`/`ibsi` with a dangling or ±1%-only backing to
`oracle=mirp` at the SPEC §7 exact tier, asserted **per slice and on the four-slice mean**:

| features                                                                                                                                              | tolerance    | agreement measured                                          |
| ----------------------------------------------------------------------------------------------------------------------------------------------------- | ------------ | ----------------------------------------------------------- |
| `SAE`, `LAE`, `LGLZE`, `HGLZE`, `SALGLE`, `SAHGLE`, `LALGLE`, `LAHGLE`, `GLN`, `GLNN`, `SZN`, `SZNN`, `ZP`, `GLV`, `ZV` | `rel=1e-9` | **2.0e-16** vs mirp, **2.6e-16** vs PyRadiomics |
| `ZE`                                                                                                                                                | `rel=4e-3` | 2.5e-3 per slice, 2.0e-3 on the mean, against both tools    |

**Nothing was demoted.** No feature's oracle claim failed on re-run, so no row moves to
`status=regression` and nothing leaves the oracle files. The one feature that does not reach the
exact tier keeps its oracle claim with a measured, cited band rather than losing it.

Recipe `glszm.ibsi_phantom_2d`: the four IBSI digital-phantom slices from `test_data.h`, each
featurised on its own with `IBSI=true` and `GREYDEPTH=128`; mirp reaches the same configuration with
`by_slice=True` and `base_discretisation_method="none"` (the phantom is already discrete 1..6).

The published-consensus file stays alongside at `rel=1e-2`. The two are complementary rather than
redundant: the IBSI values are quoted to three significant figures and fix the **definition**; mirp
fixes the **digits**. Dropping the IBSI file would leave the family tied to one implementation and
nothing published; dropping the mirp file would leave a band seven orders of magnitude looser than
the measured agreement — which is exactly the gap `GLSZM_ZE`'s 2.5e-3 had been sitting inside.

## The one residual, and why it is not fixed here

`glszm.cpp` computes zone entropy as `fast_log10(p / sum_p + EPS) / LOG10_2`. `Nyxus::fast_log10` is
a float-precision quadratic approximation of the logarithm, not `std::log10`, and that is the whole
of the 2.5e-3:

| slice | Nyxus              | mirp               | PyRadiomics        | rel    |
| ----- | ------------------ | ------------------ | ------------------ | ------ |
| z1    | 2.3185803890228271 | 2.321928094887362  | 2.321928094887361  | 1.4e-3 |
| z2    | 2.2318551199776784 | 2.2359263506290326 | 2.2359263506290312 | 1.8e-3 |
| z3    | 1.5809745788574219 | 1.5849625007211563 | 1.5849625007211552 | 2.5e-3 |
| z4    | 1.5809745788574219 | 1.5849625007211563 | 1.5849625007211552 | 2.5e-3 |

Slices z3 and z4 make it plain: both references give exactly `log2(3) = 1.5849625007`, which is what
a three-zone ROI with equal probabilities must produce. Nyxus returns 1.5809745789.

The normalisation is correct here — the probability is `p / sum_p`, not a raw count — so the
logarithm is the only difference from the reference. The family is fully vettable with the residual
stated, so this PR states it: `GLSZM_ZE` carries a feature-specific `rel=4e-3` band with its cause
named at the site, and the other fifteen features stay at the exact tier. The band is deliberately
too tight to absorb anything larger.

The same approximation is live in `glcm.cpp`, `glrlm.cpp` and four 3D twins. `gldm.cpp` was switched
to `std::log` when the 2D GLDM family was re-vetted, so the codebase is currently mixed. Finishing
that sweep is a public-value change and belongs on its own branch.

## Two tables, because a mean cannot vet the parts

The four-slice mean is the quantity IBSI publishes, but it cannot vet the four values behind it: two
slice errors that cancel leave it unmoved, and a defect confined to one slice reaches it quartered.
`test_2d_glszm_mirp.h` pins mirp's value for each slice as well as the mean, and every assertion
checks both.

**Negative control.** Adding +0.02 to the pinned `GLSZM_SAE_z1` and −0.02 to `GLSZM_SAE_z2` leaves
the mean untouched:

```
[ RUN      ] TEST_NYXUS.TEST_2D_GLSZM_SAE_IBSI
[       OK ] TEST_NYXUS.TEST_2D_GLSZM_SAE_IBSI (0 ms)          <- mean only: passes
[ RUN      ] TEST_NYXUS.TEST_2D_GLSZM_SAE_MIRP
	abs of (actual=0.105556 - groundtruth=0.125556)=0.02 > tolerance=1.25556e-10
  ... GLSZM_SAE_z1
[  FAILED  ] TEST_NYXUS.TEST_2D_GLSZM_SAE_MIRP (0 ms)          <- per slice: fails, and names it
```

## A drift guard that had stopped guarding

`test_2d_glszm_regression.h` pins Nyxus' output in **default** mode (`IBSI=false`, 64 grey levels) —
a different config point, not a duplicate of the oracle files. In default mode Nyxus weights the
grey-level-dependent features by raw intensity instead of by the grey-level index, so `GLSZM_HGLZE`
is 16.44 in IBSI mode and 1497.57 here on the same fixture. Nothing reproduces that, and it is what
a caller gets without asking for IBSI compliance, so a snapshot is the right kind of test for it.

It was asserting at ±1%, and one pin had drifted underneath that band without failing:

| feature          | pinned   | computed now        | rel              |
| ---------------- | -------- | ------------------- | ---------------- |
| `GLSZM_SALGLE` | 0.109386 | 0.10956092298348039 | **1.6e-3** |

Every other entry was a truncation of the current value to 1e-5 or better. All sixteen are
re-recorded at full precision and the band tightened to `agrees_gt`'s `rel=1e-3` default. Which of
the two `GLSZM_SALGLE` values is right is not a question this recipe can answer — nothing vets
default mode — so what is recorded is that the value moved and the test did not notice.

## Also cleaned up

- **A shared fixture header.** `test_2d_glszm_common.h` holds the settings recipe, the per-slice
  featurisation and the mean-against-a-table assertion; the two existing files had each carried
  their own copy of four hand-unrolled load/calculate/save sequences. Fixtures only, no reference
  data (SPEC 6.3.1) — each file keeps its table beside its own assertions.
- **Include hygiene** in the direction the reviewer asked for: the fixture header spells out what it
  includes and why, and the files including it repeat none of it. The direct `roi_cache.h`,
  `features/pixel.h` and `environment.h` includes are gone from both existing files.
- **The `_ibsi` table was checked for digit count** — all sixteen entries carry three significant
  figures, so it is a transcription of the published consensus and not a run pasted under a
  published-value name.
- **Range, identity, digit-count and cross-table checks run over all 112 pins in the family**, not
  spot-checked: unit-interval and non-negativity ranges, `normalised ≤ raw`, `LGLZE ≤ HGLZE`,
  `SAE ≤ LAE`, each IBSI pin within `rel=1e-2` of the corresponding mirp mean, and each mirp mean
  equal to the arithmetic mean of its own four per-slice pins within `rel=1e-12` — the check that
  stops a per-slice table and a mean table drifting apart in a copy-paste. All pass.
- **Two registry errors beyond the dangling target.** `current_test` listed
  `test_3d_glszm_regression.h` and `test_3d_glszm_pyradiomics.h` on every **2D** row, neither of
  which asserts a 2D feature; and the note recording that `test_3d_glszm_regression.h` is orphaned
  (never `#include`d, so it never runs) sat on the 2D rows, where correcting them would have erased
  it. That is a 3D GLSZM problem — the 3D rows still carry it and the audit report restates it so
  the 3D re-vet inherits it.

## New audit artifacts

| file                                                         | what it holds                                                                                                                                                                                                      |
| ------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `tests/vetting/oracles/gen_glszm_mirp.py`                  | offline generator; re-verifies**both** pinned tables (16 means + 64 per-slice) and exits non-zero on a mismatch or an unproducible pin. Current run: 80 verified, 0 failed, 0 unproducible, all at `rel=0` |
| `audit/glszm_2d_mirp_vetting_report.md`                    | full comparison, the residual's cause, negative control, range/identity checks, the drift finding                                                                                                                  |
| `audit/glszm_2d_ibsi_vetting_report.md`                    | the published table checked entry by entry, and why both oracle files stay                                                                                                                                         |
| `audit/glszm_2d_golden_regen.md`                           | runnable regeneration steps for all three tables, incl. the mirp→Nyxus column map                                                                                                                                 |
| `audit/scan_glszm_coverage.py` + `glszm_2d_coverage.csv` | feature→test map scanned out of the sources;`--check` runs the acceptance check and is clean on all 16 rows                                                                                                     |
| `config_recipes.md`, `TOOLS.md`, `not_covered.md`      | the recipe, mirp's GLSZM column naming, and the closed backlog entry                                                                                                                                               |

## How it was verified

| gate                                                                             | result                                                                                                    |
| -------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------- |
| gtest, Windows/MSVC                                                              | **814 passed**, 798 before — exactly the 16 cases added                                            |
| ASan + UBSan gtest, Linux/gcc (`-fsanitize=address,undefined`, RelWithDebInfo) | **814 passed, 0 sanitizer diagnostics**, exit 0; 48 GLSZM cases (16 ibsi + 16 mirp + 16 regression) |
| `pytest tests/python/`                                                         | 85 passed, 1 skipped; the 7`@pytest.mark.arrow` failures are the tiff-only local build                  |
| `gen_glszm_mirp.py`                                                            | 80 verified, 0 failed, 0 unproducible                                                                     |
| PyRadiomics 3.0.1 corroboration                                                  | 7.0e-16 worst case against the pinned mirp goldens                                                        |
| range / identity / digit-count / cross-table checks                              | 112 pins, 0 violations                                                                                    |
| `check_test_names.py`                                                          | 0 naming violations                                                                                       |
| `check_coverage.py --check`                                                    | clean                                                                                                     |
| `scan_glszm_coverage.py --check`                                               | 16 rows, clean                                                                                            |

Fresh oracle rerun, not just "tests pass": every golden in `test_2d_glszm_mirp.h` was produced by
the generator and re-verified by it against a live mirp run at the recipe's config.


tests/test_2d_glszm_mirp.h
```diff
@@ -0,0 +1,214 @@
...
+// mirp goldens for the 2D GLSZM family (SPEC 6.4 provenance).
+// tool=mirp 2.6.0; env=nyxus_mirp (conda); recipe=glszm.ibsi_phantom_2d, i.e. the four IBSI
+// digital-phantom slices from test_data.h featurised one at a time, with by_slice=True and
+// base_discretisation_method="none" (the phantom is already discrete 1..6), which is what Nyxus
+// computes in IBSI mode; generator=tests/vetting/oracles/gen_glszm_mirp.py.
+//
+// Fifteen of the sixteen features are bit-identical to mirp (2.0e-16 worst case), so they assert at
+// the SPEC 7 "exact" tier. test_2d_glszm_ibsi.h pins the same quantities as published IBSI consensus
+// values, quoted to three significant figures, and therefore asserts at rel=1e-2: IBSI fixes the
+// definition, mirp fixes the digits.
```
vjaganat90: `2.0e-16 worst case` is floating-point agreement, not bit identity. Please say “agree within 2.0e-16” here and in the report/registry so the wording matches the values shown.

tests/test_2d_glszm_mirp.h
```diff
@@ -0,0 +1,214 @@
...
+
+    {"GLSZM_ZV_z1", 2.8}, {"GLSZM_ZV_z2", 5.63265306122449},
+    {"GLSZM_ZV_z3", 34.888888888888886}, {"GLSZM_ZV_z4", 40.666666666666664},
+
+    {"GLSZM_ZE_z1", 2.321928094887362}, {"GLSZM_ZE_z2", 2.2359263506290326},
+    {"GLSZM_ZE_z3", 1.5849625007211563}, {"GLSZM_ZE_z4", 1.5849625007211563}
+};
+
+// rel=1e-9: agrees_gt divides the golden by this, so a larger argument is a tighter band
+static const double glszm_2d_mirp_frac_tolerance = 1e9;
```
vjaganat90: `rel=1e-9` is not SPEC §7’s `exact` tier; that row is `abs 1e-9`. Either use an absolute `ASSERT_NEAR(..., 1e-9)` for these 15 features, or keep the relative check and stop labeling it exact.

tests/test_2d_glszm_mirp.h
```diff
@@ -0,0 +1,214 @@
...
+    {
+        const std::string key = feature_name + "_z" + std::to_string (z + 1);
+        SCOPED_TRACE ("mirp " + key);
+        ASSERT_TRUE (glszm_2d_mirp_slice_ref_vals.count(key) > 0) << key;
+        ASSERT_TRUE (agrees_gt (per_slice[z], glszm_2d_mirp_slice_ref_vals.at(key), frac_tolerance))
+            << key;
+    }
+
+    // then the four-slice mean, the quantity IBSI publishes
+    assert_glszm_feature_against_golden_values (feature_, feature_name, glszm_2d_mirp_ref_vals,
```
vjaganat90: `per_slice` already holds and asserts all four values, but this helper reruns the feature four more times for the mean. Compute the mean from `per_slice` so the averaged assertion covers the same values just checked.

tests/vetting/config_recipes.md
```diff
@@ -158,3 +158,26 @@ chosen reference tool (SPEC 5). Oracle tests reference a recipe by id; this file
   `test_3d_glcm_dump_regression()`.
 - A different binCount from `glcm3d.pyradiomics_bincount20`, so the two benchmarks are not
   comparable to each other by construction.
+## glszm.ibsi_phantom_2d
```
vjaganat90: This recipe is present, but SPEC §5.1/§6.3 still require `tests/vetting/matrix/glszm.md` and the benchmark entry in `tests/vetting/benchmarks.md`. Please add both for a complete family migration.

tests/test_2d_glszm_mirp.h
darkclad: Fixed

tests/test_2d_glszm_mirp.h
darkclad: Fixed

tests/test_2d_glszm_mirp.h
darkclad: Fixed

tests/vetting/config_recipes.md
darkclad: Fixed

PR #441, opened 2026-08-20
Rehabilating old 3d_glcm regression tests and untangling few other tests

sameeul:
## Summary

`test_3d_glcm_coverage.h` ran a generic `TEST_P`-parameterized sweep (internally "Wave-9") over every 3D family's featureset, instantiated two ways:

- `GLCM_WITH_3P_EMBEDDED_GT` — asserted GLCM values against embedded goldens. Redundant for GLCM: `test_3d_glcm_pyradiomics.h` already asserts the same features against PyRadiomics, and this fixture's "finite value after the full pipeline runs" check used the identical bypass-`FeatureManager`, hand-construct-the-calculator pattern the named tests already use, so it caught nothing a failing named test wouldn't.
- `GLCM_UNVETTED_LOCAL_REGRESSION` — a table-driven drift guard on 36 values with no oracle claim, whose failures reported as a sanitized test-param string instead of a feature name.

Both instantiations also incidentally covered two things they weren't designed for: confirming every feature name resolves, and confirming every registered `Feature3D` code has exactly one provider (`FeatureManager::check_11_correspondence()` had a `//xxxx what about 3D and IMQ?` TODO — it only ever checked `Feature2D`). Retiring the file meant giving both of those a real home first.

This PR does that in three steps, then a fourth pass fixing what code review found in the process.

## What changed

**1. Port the drift guards to named tests** (`25378826`)

`glcm_3d_regression_coverage_ref_vals` (36 features, grey64 profile) is ported verbatim into `test_3d_glcm_regression.h` as individually named `*_grey64_regression` tests — same phantom, same settings, same numbers, just named so a failure points at the feature instead of a sanitized param string. `test_3d_coverage_common.h` gets a `families_with_individually_ported_regression()` escape valve so its completeness guard (`TEST_3D_FEATURE_COVERAGE_COUNTS`) stops expecting a local coverage table for `glcm`. `oracle_coverage.csv`'s `current_test` column is updated on the 34 affected rows to credit `test_3d_glcm_regression.h`.

**2. Retire `test_3d_glcm_coverage.h` entirely** (`968f8fa2`)

Every one of GLCM's 59 public features now has an independently named test with a real golden comparison (23 in `test_3d_glcm_pyradiomics.h`, 36 from step 1) — verified by cross-referencing the enum's name strings against every assert call site and `TEST()` registration. The file and its `#include` are deleted; `scan_glcm3d_coverage.py` drops the now-unnecessary SWEEP blanket-credit mechanism; `oracle_coverage.csv` drops the dead `current_test` reference on all 59 GLCM rows.

**3. Close the FeatureManager correspondence gap** (`b4c446e6`, `09a6e08c`)

`FeatureManager::check_11_correspondence()` is extended with a second loop over `Feature3D`'s code range (`Feature3D::COV == (int)Feature2D::_COUNT_`, so the bound is exact, not a guess) — the first time any of the 213 3D feature codes have been checked in production for having exactly one provider. `FeatureIMQ` is left as an explicit follow-up TODO at the same site (no `findFeatureNameByCode(FeatureIMQ)` overload exists yet to name a failure by).

FeatureManager had zero dedicated unit tests before this; `test_feature_manager_mechanics.h` adds one. Because registration happens in `FeatureManager`'s constructor, a bare instance is already fully populated, so `compile()` can be exercised directly — no phantom, no I/O, no ROI setup, milliseconds instead of ~100ms × 213. It's added to `check_test_names.py`'s `DIM_AGNOSTIC` list (one `compile()` call covers 2D and 3D together, so it carries no dimension token).

**4. Code-review follow-up**

- `oracle_coverage.csv` had flipped from LF to CRLF in step 1 (`csv.writer`'s default `lineterminator`), turning a 59-cell edit into a 759-line whole-file diff. Normalized back to LF; added `*.csv text eol=lf` to `.gitattributes` so it can't recur regardless of what tool touches the file next.
- 30 registry rows' `notes` column still named the deleted `test_3d_glcm_coverage.h` after step 2's `current_test` cleanup (`--check` only validates structured columns, so it didn't catch free-text drift). Dropped the stale clause.
- `glcm_3d_golden_regen.md` still described the deleted file as live. Corrected, and given a new "grey64 table and the retired Wave-9 sweep" section that's now the durable home for the retirement history, replacing history/phase language ("Wave-9", `// FIX: ...` defect-history comments) that had leaked into `test_3d_glcm_regression.h`, `test_feature_manager_mechanics.h`, and `test_all.cc` — those now carry one-line pointers instead.
- `assert_3d_glcm_feature_grey64_regression` duplicated ~40 lines of `assert_3d_glcm_feature_regression` (phantom load, ROI gather, GLCM calculate); both now share a `run_3d_glcm_feature_pipeline()` helper and differ only in the two `GREYDEPTH` settings and comparison shape (aggregate-then-compare vs. per-angle compare).
- The grey64 table's drift tolerance was a flat rel=1e-6 for all 36 pins — three orders looser than the file's own rel=1e-9 standard, which grants 1e-6 only to the two information measures (`INFOMEAS1`/`INFOMEAS2`, both routing through `fast_log10` and losing bits to compiler-specific float rounding). Reusing that same policy function surfaced a real measured residual on `DIFENTRO_AVE` (rel 3.6e-9 — `DIFENTRO`'s difference-entropy calc hits the same `fast_log10` path). Rather than keep growing a per-feature exception list, both tables now assert at a single flat rel=1e-8 — still headroom over every measured residual — so `glcm_3d_regression_frac_tolerance()` is gone and both call sites just pass the `1.e8` literal.
- `feature_mgr.cpp`'s new 3D loop bound was `(int)Nyxus::Feature2D::_COUNT_`, justified by a three-line comment explaining that `Feature3D::COV` is defined as exactly that value. Switched to `(int)Nyxus::Feature3D::_FIRST_` (`= COV`, `featureset.h:917`) instead — the idiom already used at `featureset.h:1128/1132/1138` — which says what the bound *is* rather than requiring the reader to know a cross-enum numbering fact, and makes the justifying comment unnecessary.

## How this was verified

- `python3 tests/vetting/check_coverage.py --check` — clean.
- `python3 tests/vetting/audit/scan_glcm3d_coverage.py --check` — clean (59 rows).
- `python3 tests/vetting/check_test_names.py` — 0 naming violations.
- Fresh local `runAllTests` (gtest) rebuild after every change.
- Full `runAllTests`: 840/840 passing, including all 111 3D-GLCM + FeatureManager-mechanics tests at the tightened grey64 tolerance.
- Not run: `pytest tests/python/` — no Python extension (`nyxus.backend`) build available in this environment; nothing in this PR touches Python-facing code, bindings, or feature math.


vjaganat90:
- ~~Can we correct the provenance comment to identify the actual PyRadiomics recipe—compat phantom, binCount 20, `GLCM_GREYDEPTH=-20`—rather than the unrelated `-100` regression profile?~~
The comment should be in the oracle test file, perhaps out of scope for this PR

- After removing the automatically enumerated GLCM suites, what now guarantees that every named replacement remains present and registered? Since `scan_glcm3d_coverage.py --check` provides that guarantee but is not run in CI, can it be added to the workflow?

- ~~Can “previously-unvetted features” be replaced with “features vetted under `glcm3d.pyradiomics_bincount20` and regression-only under grey64”? SPEC explicitly requires configuration-specific language and rejects a bare “unvetted” label when a feature is vetted under another recipe.~~

This was a terminology trip-up, the comment correctly claims them as regression.

- Was the wholesale LF→CRLF conversion of `oracle_coverage.csv` intentional? If not, can it be reverted so only the substantive registry changes remain?

darkclad (approved):

PR #442, opened 2026-08-20
test(2d neighbor): re-vet the family, and set its tolerances from measurement

darkclad:
Eight of the nine 2D neighbour rows have read `vetted` since PR #390. This re-vet re-ran both
oracles rather than trusting that label. **The values hold — exactly.** What did not hold up was
everything around them: the assertion band, both generators' self-checks, and where the
`PERCENT_TOUCHING` bound assertions lived.

No feature is promoted or demoted here. The family's coverage was right; its evidence was weaker
than it looked.

## The values, measured for the first time

| oracle                                                    | features × ROIs | measured residual                                                                      | was asserted at        |
| --------------------------------------------------------- | ---------------- | -------------------------------------------------------------------------------------- | ---------------------- |
| CellProfiler 4.2.8 (`MeasureObjectNeighbors`, Adjacent) | 2 × 5           | **exactly 0** — the same double, ten times out of ten                           | 1e-4**absolute** |
| analytic (numpy recomputation of the closed forms)        | 6 × 5           | **1.2e-16** — 29 of 30 bit-identical, one ulp on `ANG_BW_NEIGHBORS_STDDEV` L1 | 1e-4**absolute** |

On `CLOSEST_NEIGHBOR1_ANG`'s value of 258.69, a 1e-4 absolute band is a relative 3.9e-7 — nine
orders of magnitude of slack over a measured zero. The registry's `tolerance` column was **empty on
all nine rows**, and both generators validated at 1e-4 too, so nothing anywhere recorded a measured
number. Both files now assert at `rel=1e-9`.

Eighteen of the thirty analytic goldens are **structural zeros** (`CLOSEST_NEIGHBOR2_*` is 0 below
two in-radius neighbours; the sample standard deviation of one angle is 0). `agrees_gt` computes
`tolerance = golden / frac`, so a zero golden demands **bit-exact equality** — for those eighteen the
assertion is now as strong as it can be, and it passes.

Goldens are re-pinned at full `repr` precision. They had been recorded to ~15 digits
(`2.54950975679639` for `2.5495097567963922`), harmless at 1e-4 and real margin at 1e-9.

## Both generators were validating themselves

Each compared its fresh run against a hardcoded copy of the goldens kept **inside the generator**
(`PINNED`, `CP_VETS`). Editing a golden in either header would have been caught by nothing.

`gen_neighbor_analytic.py` was detached twice over: its comment and printed banner both named
`test_2d_neighbor_regression.h` — a file those goldens had **already been moved out of**. The
CellProfiler generator's `NYXUS_PT` divergence table was a second hardcoded copy, already stale to 6
significant figures.

Both now parse the header they feed, verify every pin, **and report the reverse gap** — a feature the
oracle produces that the header pins nothing for — so a header quietly losing coverage cannot read as
all-clear.

That reverse check earned its keep immediately: the first version of my header parser used a
non-greedy regex, which swallows the closing brace of each label block's **last** entry. On the
multi-line analytic table it happened to work; on the single-line CellProfiler table it silently
dropped `CLOSEST_NEIGHBOR1_DIST` from every ROI. It reported five `UNPINNED` rows instead of a clean
pass. Both generators now count braces. Worth noting the analytic table had been parsing correctly
*only* because of where its commas and newlines fell — right by accident, one reformat from wrong.

## An oracle-suffixed test asserting a feature with no oracle

`test_2d_neighbor_analytic.h` held `test_2d_neighbor_percent_touching_enclosed_analytic`, asserting
`PERCENT_TOUCHING ≤ 100` and `= 100` for the enclosed ROI. That feature's registry row correctly
reads `status=regression` with an empty oracle — CellProfiler measures a different quantity and
diverges on 3 of 5 ROIs.

Those two facts contradict each other. **Every coverage scanner in this tree attributes an oracle
from the test function's name suffix**, so a scan would have credited `PERCENT_TOUCHING` with
`oracle=analytic` on the strength of a bound assertion.

SPEC §4.4 classes these as invariants — "does output obey a required property/bound/relation" — not
oracle comparisons. They move to `test_2d_neighbor_invariant.h` under `_invariant` names. The
registry row is unchanged, because it was already right. The enclosed-ROI check also becomes
`ASSERT_DOUBLE_EQ`: the value is 100 by construction, so any tolerance weakened it.

### The divergence, measured

| ROI | CellProfiler | Nyxus    |                       |
| --- | ------------ | -------- | --------------------- |
| 1   | 100.0000     | 100.0000 | agree (both saturate) |
| 2   | 50.0000      | 66.6667  | **diverge**     |
| 3   | 50.0000      | 66.6667  | **diverge**     |
| 4   | 50.0000      | 50.0000  | agree                 |
| 5   | 50.0000      | 33.3333  | **diverge**     |

Nyxus counts distinct contour pixels 8-adjacent to a neighbour over contour length; CP counts outline
pixels overlapping a `disk(distance+0.5)`-dilated neighbour over perimeter. No CP distance method
reproduces it. A definition difference, not a defect — so it stays pinned rather than compared.

## Also cleaned up

- **Include hygiene, both directions.** `test_2d_neighbor_regression.h` repeated **nine** includes
  that `test_2d_neighbor_common.h` already supplies and the two oracle files correctly omit
  (`<unordered_set>` was not used at all). And `common.h` itself had the opposite problem, which
  matters more because every file in the family includes it: it uses `Fsettings`, `NyxSetting`,
  `ImageMatrix`, `Pixel2` and `PixIntens` directly while including a header for **none** of them,
  reaching all five transitively through `test_main_nyxus.h`. Now explicit, each with a trailing
  comment naming what it supplies — the shape `test_2d_glcm_common.h` and `test_2d_glrlm_common.h`
  already use.
- **Duplicate assertions removed.** That file also pinned `NUM_NEIGHBORS`, `CLOSEST_NEIGHBOR1_DIST`
  and `CLOSEST_NEIGHBOR2_DIST` — all three vetted at the exact tier on the same fixture and config,
  so the snapshot was a second, looser assertion of an already-checked value. Its own header comment
  had said so for some time while the table kept them.
- **History out of test headers** ("FIX #13 … was 33.33", "was 87.5 (7/8) because the approximate
  min_sqdist …", "now vetted at tight tolerance by …"). That belongs in the audit report.
- **`const` tables read with `.at()`.** All three were non-`const` and read with `operator[]`, which
  default-inserts a missing key so the comparison runs against the zero it just created — which
  `agrees_gt` reads as a demand for exact 0.
- **A coverage over-crediting bug in the shared scanner logic.** A python helper's body was bounded
  at the next *helper* rather than the next top-level `def`, so the last helper in a file swallowed
  every test below it and handed every feature name in the file to each of its callers. Here it
  credited a `PERCENT_TOUCHING`-only test with `NUM_NEIGHBORS` and `CLOSEST_NEIGHBOR1_DIST`. Fixed in
  this family's scanner; **the same logic is copied in the other per-family `scan_*` scripts.**
- **Registry hygiene.** All nine rows had an empty `tolerance` and a prose sentence where the
  `config_recipe` id belongs.

## New audit artifacts

| file                                                               | what it holds                                                                                           |
| ------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------- |
| `audit/neighbor_2d_cellprofiler_vetting_report.md`               | the exact-zero result, the loose band, the divergence table                                             |
| `audit/neighbor_2d_analytic_vetting_report.md`                   | the 1.2e-16 result, why the analytic oracle is not circular, the negative control, the parser bug       |
| `audit/neighbor_2d_golden_regen.md`                              | regeneration steps for all three tables + the two convention traps                                      |
| `audit/scan_neighbor_coverage.py` + `neighbor_2d_coverage.csv` | feature→test map;`--check` adds "no oracle-suffixed test for an oracle-less feature". Clean on all 9 |
| `config_recipes.md`, `TOOLS.md`, `not_covered.md`            | the recipe, the conda CellProfiler setup, the findings                                                  |

## How it was verified

| gate                                                                             | result                                                                                                                                       |
| -------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
| gtest, Windows/MSVC                                                              | **849 passed** — net zero change, 5 neighbour cases before and after                                                                  |
| ASan + UBSan gtest, Linux/gcc (`-fsanitize=address,undefined`, RelWithDebInfo) | **849 passed, 0 sanitizer diagnostics**, exit 0; 5 neighbour cases                                                                     |
| `pytest tests/python/`                                                         | 86 passed, 1 skipped; the 7`@pytest.mark.arrow` failures are the tiff-only local build                                                     |
| `gen_neighbor_analytic.py`                                                     | 30 verified, 0 failed, 0 unproducible, 0 unpinned                                                                                            |
| `gen_neighbor_cellprofiler.py`                                                 | 10 verified, 0 failed, 0 unproducible, 0 unpinned, every residual`rel=0`                                                                   |
| `check_test_names.py`                                                          | 0 naming violations                                                                                                                          |
| `check_coverage.py --check`                                                    | clean                                                                                                                                        |
| `scan_neighbor_coverage.py --check`                                            | 9 rows, clean                                                                                                                                |
| range / identity / cross-table checks                                            | 45 pins, 0 violations — including "no feature pinned in two tables", which confirms the duplicate assertions were removed rather than moved |

Fresh oracle reruns, not just "tests pass": CellProfiler was run in-process from the conda env and
the analytic oracle recomputed, both at the recipe's config, and every golden in both headers was
re-verified by its generator against those runs.


tests/test_2d_neighbor_analytic.h
```diff
@@ -53,24 +62,33 @@ static ref_vals_map_by_label<double> neighbor_2d_analytic_ref_vals_by_label{
...
-	ASSERT_NEAR(roiData.at(label).fvals[static_cast<int>(feature)][0],
-		neighbor_2d_analytic_ref_vals_by_label[label][feature_name], 1e-4);
+
+	// .at() on a const table: operator[] would default-insert a missing key and compare against
+	// the zero it just created, which agrees_gt reads as a demand for exact 0
+	ASSERT_TRUE(neighbor_2d_analytic_ref_vals_by_label.count(label) > 0) << label;
+	const auto& golden = neighbor_2d_analytic_ref_vals_by_label.at(label);
+	ASSERT_TRUE(golden.count(feature_name) > 0) << feature_name;
+
+	ASSERT_TRUE(agrees_gt(roiData.at(label).fvals[static_cast<int>(feature)][0],
```
vjaganat90: SPEC §7 defines `exact` as an absolute `1e-9` band, while `agrees_gt(..., 1e9)` is relative and permits about `2.6e-7` at the largest angle. Please use `ASSERT_NEAR(..., 1e-9)` here and align the generator and registry to `abs=1e-9`.

tests/test_2d_neighbor_cellprofiler.h
```diff
@@ -1,40 +1,56 @@
...
-	ASSERT_NEAR(roiData.at(label).fvals[static_cast<int>(feature)][0],
-		neighbor_2d_cellprofiler_ref_vals_by_label[label][feature_name], 1e-4);
+
+	// .at() on a const table: operator[] would default-insert a missing key and compare against
+	// the zero it just created, which agrees_gt reads as a demand for exact 0
+	ASSERT_TRUE(neighbor_2d_cellprofiler_ref_vals_by_label.count(label) > 0) << label;
+	const auto& golden = neighbor_2d_cellprofiler_ref_vals_by_label.at(label);
+	ASSERT_TRUE(golden.count(feature_name) > 0) << feature_name;
+
+	ASSERT_TRUE(agrees_gt(roiData.at(label).fvals[static_cast<int>(feature)][0],
```
vjaganat90: Same issue here: the measured zero residual supports the exact tier, but that tier is `abs=1e-9`, not `rel=1e-9`. Please use `ASSERT_NEAR(..., 1e-9)` and align the generator and registry.

tests/vetting/oracle_coverage.csv
```diff
@@ -134,15 +134,15 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
+2D,NUM_NEIGHBORS,neighbor,vetted,cellprofiler,agreed,neighbor.scene2d_radius1,rel=1e-9,test_2d_morphology_regression.h;test_2d_neighbor_cellprofiler.h;test_2d_neighbor_invariant.py,test_2d_neighbor_cellprofiler.h,,,audit,"vetted vs CellProfiler 4.2.8 MeasureObjectNeighbors (Adjacent method, neighbors_are_objects) on the neighborhood2d_scene_labels fixture at PIXELDISTANCE=1, recipe neighbor.scene2d_radius1; agreement is EXACTLY 0 -- CP reproduces Nyxus bit for bit on all five ROIs -- asserted per ROI at rel=1e-9 in test_2d_neighbor_cellprofiler.h. Re-measured 2026-08-20: the previous assertion was ASSERT_NEAR at 1e-4 absolute with no tolerance recorded here. gen_neighbor_cellprofiler.py re-verifies every pin in the header it feeds; tests/vetting/audit/neighbor_2d_cellprofiler_vetting_report.md"
```
vjaganat90: After rebasing onto current `main`, these rows still need assertion-level `test_name` and `benchmark` fields, plus `tests/vetting/matrix/neighbor.md` and a benchmark entry for this scene. The oracle mapping itself looks right; this is the remaining SPEC §3/§5.1/§6.3 wiring.

tests/vetting/oracles/gen_neighbor_analytic.py
```diff
@@ -1,42 +1,45 @@
...
-env=nyxus_mirp (conda), numpy 1.26/2.x; generator=tests/vetting/oracles/
-gen_neighbor_analytic.py. Run offline; CI never invokes it.
+"""OFFLINE analytic oracle for the 2D neighbour second-distance and angle features.
+
+    python tests/vetting/oracles/gen_neighbor_analytic.py        (from the repository root)
+
+Prints the paste-ready goldens AND re-verifies every golden pinned in test_2d_neighbor_analytic.h,
+exiting non-zero on any mismatch or on any pin this generator cannot produce.
+
+Recipe `neighbor.scene2d_radius1`: the `neighborhood2d_scene_labels` fixture from tests/test_data.h,
```
vjaganat90, line 8: `SCENE` is a handwritten second copy of `neighborhood2d_scene_labels`; the same copy also lives in `gen_neighbor_cellprofiler.py`. This does not invalidate either oracle, but a `test_data.h` fixture edit leaves both generators reproducing the old scene, so please parse or share the checked-in fixture for regeneration.

tests/test_2d_neighbor_common.h
```diff
@@ -4,22 +4,28 @@
...
-#include "test_main_nyxus.h"
+#include "../src/nyx/feature_settings.h"        // Fsettings, NyxSetting
+#include "../src/nyx/roi_cache.h"               // LR, init_label_record_3, update_label_record_3
+#include "../src/nyx/features/basic_morphology.h"  // BasicMorphologyFeatures
+#include "../src/nyx/features/contour.h"        // ContourFeature
+#include "../src/nyx/features/image_matrix.h"   // ImageMatrix
+#include "../src/nyx/features/neighbors.h"      // NeighborsFeature
+#include "../src/nyx/features/pixel.h"          // Pixel2, PixIntens
+#include "test_data.h"                          // neighborhood2d_scene_labels
+#include "test_main_nyxus.h"                    // the Nyxus namespace the loaders live in
```
vjaganat90: `test_2d_neighbor_common.h` uses nothing from this header; instead the assertion headers receive `agrees_gt` transitively through it. Please remove it here and include `test_main_nyxus.h` only where `agrees_gt` is called (`<string>` directly in the invariant header).

tests/test_2d_neighbor_analytic.h
darkclad: Fixed

tests/test_2d_neighbor_cellprofiler.h
darkclad: Fixed

tests/vetting/oracle_coverage.csv

tests/vetting/oracles/gen_neighbor_analytic.py
darkclad: Fixed

tests/test_2d_neighbor_common.h
darkclad: Fixed

PR #443, opened 2026-08-20
test(2d ngtdm): vet the family against mirp, per slice and averaged

darkclad:
All five 2D NGTDM rows have claimed `oracle=pyradiomics` with
`target_test=test_2d_ngtdm_pyradiomics.h` since the tracker was imported — **a file that does not
exist** — with an **empty `tolerance` column** and a prose sentence where the `config_recipe` id
belongs. The only test covering them asserted the published consensus at ±1%.

Running independent oracles for the first time found Nyxus reproducing **both** mirp and PyRadiomics
on every feature and every slice. Nothing is promoted or demoted; the claim was right, and is now
backed, measured, and pinned at the precision the agreement supports.

## Three implementations, one answer

| comparison                                     | worst residual                 |
| ---------------------------------------------- | ------------------------------ |
| Nyxus vs mirp 2.6.0                            | **3.2e-16**              |
| Nyxus vs PyRadiomics 3.0.1                     | **2.4e-16**              |
| mirp vs PyRadiomics                            | **1.6e-16**              |
| Nyxus vs the published IBSI consensus (3 s.f.) | 4.1e-3, on`NGTDM_COARSENESS` |

All five features move to `rel=1e-9`, the SPEC §7 exact tier, asserted **per slice and on the
four-slice mean**, with the published-consensus file alongside at `rel=1e-2`. Everything left is
floating-point summation order. This family has no entropy term, so the `fast_log10` approximation
that costs 2D GLDM and 2D GLSZM their exact tier does not arise here — all five reach it.

**PyRadiomics — the oracle the rows actually named — was run rather than quietly replaced.** It
agrees with mirp to 1.6e-16, so the original claim was correct in substance and only ever lacked a
file. Only mirp's values are pinned; a second table identical to the first to 1.6e-16 is redundancy,
not coverage.

## Two tables, because a mean cannot vet the parts

**Negative control.** Adding +0.01 to the pinned `NGTDM_CONTRAST_z1` and −0.01 to `_z2` leaves the
mean untouched:

```
[ RUN      ] TEST_NYXUS.TEST_2D_NGTDM_CONTRAST_IBSI
[       OK ] TEST_NYXUS.TEST_2D_NGTDM_CONTRAST_IBSI (0 ms)      <- mean only: passes
[ RUN      ] TEST_NYXUS.TEST_2D_NGTDM_CONTRAST_MIRP
	abs of (actual=1.63628 - groundtruth=1.64628)=0.01 > tolerance=1.64628e-09
  ... NGTDM_CONTRAST_z1
[  FAILED  ] TEST_NYXUS.TEST_2D_NGTDM_CONTRAST_MIRP (0 ms)      <- per slice: fails, and names it
```

The generator catches it independently: `23 verified, 2 failed`.

## The static nobody was tracking

`NGTDMFeature::n_levels` is a `static int` shared by every test in the binary.
`test_2d_ngtdm_regression.h` set it to **100** and never restored it; no other NGTDM test set it at
all.

**Not a live defect** — `ngtdm.cpp` forces the grey-binning info to 0 in IBSI mode, so the IBSI and
mirp assertions are immune. That immunity was load-bearing and unstated, so it is now **measured and
asserted**: the IBSI-mode values are bit-identical with the static at 0 and at 100
(`ASSERT_DOUBLE_EQ`, not a tolerance — the setting is either read or it is not).

Outside IBSI mode it is decisive:

| config                          | `NGTDM_CONTRAST`, four-slice mean |
| ------------------------------- | ----------------------------------- |
| IBSI mode                       | 0.9252630                           |
| default mode,`n_levels = 100` | 3169.9291                           |
| default mode,`n_levels = 0`   | **6634.5048**                 |

So a future non-IBSI NGTDM test placed after the regression one would silently inherit a grey count
nobody chose for it, off by a factor of two with nothing to say why. The shared fixture now takes the
value as a **parameter** and restores whatever it found, and `test_2d_ngtdm_mechanics.h` asserts both
properties.

**`GLRLMFeature::n_levels` is assigned the same way in `test_2d_glrlm_regression.h:62`** and should
get the same treatment when that family is next touched.

## A setting that looks meaningful and is not

The previous tests set `PIXELDISTANCE=5`. `ngtdm.cpp` **never reads it** — the neighbourhood is fixed
at the d=1 8-neighbourhood the IBSI definition uses, which is why these values match the published
consensus at all. The recipe no longer sets a pixel distance, and says why.

## Also cleaned up

- **A shared fixture header.** `test_2d_ngtdm_common.h` holds the settings recipe, the per-slice
  featurisation and the mean-against-a-table assertion; both existing files had carried their own
  copy of four hand-unrolled load/calculate/save sequences.
- **Include hygiene**, both directions: every include in the fixture header is named for the symbol
  it supplies, and the files including it repeat none — the direct `roi_cache.h`, `features/pixel.h`
  and `environment.h` includes are gone from both.
- **The `_ibsi` table was checked for digit count** — all five entries carry three significant
  figures, so it is a transcription of the published consensus, not a run pasted under a
  published-value name.
- **Range, completeness, digit-count and cross-table checks over all 35 pins**, run mechanically:
  non-negativity, `NGTDM_COARSENESS` in (0,1], every feature pinned in every table, each IBSI pin
  within `rel=1e-2` of the mirp mean, each mirp mean equal to the arithmetic mean of its own four
  per-slice pins (`rel=1e-12`), and — the one specific to this family — that no default-mode
  regression pin merely duplicates its IBSI-mode value. All pass.

## New audit artifacts

| file                                                         | what it holds                                                                                                                                                                                                                                   |
| ------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `tests/vetting/oracles/gen_ngtdm_mirp.py`                  | offline generator; verifies**both** pinned tables (5 means + 20 per-slice), reports values it produces that the header pins nothing for, exits non-zero. Current run: 25 verified, 0 failed, 0 unproducible, 0 unpinned, all at `rel=0` |
| `audit/ngtdm_2d_mirp_vetting_report.md`                    | the three-way agreement, negative control, the static and its measured effect, the registry errors                                                                                                                                              |
| `audit/ngtdm_2d_ibsi_vetting_report.md`                    | the published table entry by entry, and why both oracle files stay                                                                                                                                                                              |
| `audit/ngtdm_2d_golden_regen.md`                           | regeneration steps for all three tables + both tools' config mappings                                                                                                                                                                           |
| `audit/scan_ngtdm_coverage.py` + `ngtdm_2d_coverage.csv` | feature→test map scanned from the sources; clean on all 5 rows                                                                                                                                                                                 |
| `config_recipes.md`, `not_covered.md`                    | the recipe and the findings                                                                                                                                                                                                                     |

## How it was verified

| gate                                                             | result                                                                                             |
| ---------------------------------------------------------------- | -------------------------------------------------------------------------------------------------- |
| gtest, Windows/MSVC                                              | **856 passed**, 849 before — exactly the 7 cases added (5 mirp + 2 mechanics)               |
| ASan + UBSan gtest, Linux/gcc (`-fsanitize=address,undefined`) | **856 passed, 0 sanitizer diagnostics**, exit 0; all 17 NGTDM cases including both mechanics |
| `pytest tests/python/`                                         | 86 passed, 1 skipped; the 7`@pytest.mark.arrow` failures are the tiff-only local build           |
| `gen_ngtdm_mirp.py`                                            | 25 verified, 0 failed, 0 unproducible, 0 unpinned                                                  |
| PyRadiomics corroboration                                        | 1.6e-16 against the pinned mirp goldens                                                            |
| range / completeness / digit-count / cross-table                 | 35 pins, 0 violations                                                                              |
| `check_test_names.py`                                          | 0 naming violations                                                                                |
| `check_coverage.py --check`                                    | clean                                                                                              |
| `scan_ngtdm_coverage.py --check`                               | 5 rows, clean                                                                                      |

Fresh oracle reruns, not just "tests pass": both tools were run at the recipe's config and every
golden in the header was re-verified by its generator against those runs.


tests/test_2d_ngtdm_mirp.h
```diff
@@ -0,0 +1,104 @@
...
+#include "test_2d_ngtdm_common.h"   // gtest, <string>, <vector>, the fixture and the mean helper
+#include "test_ref_vals.h"          // ref_vals_map
+
+// mirp goldens for the 2D NGTDM family (SPEC 6.4 provenance).
+// tool=mirp 2.6.0; env=nyxus_mirp (conda); recipe=ngtdm.ibsi_phantom_2d, i.e. the four IBSI
+// digital-phantom slices from test_data.h featurised one at a time, with by_slice=True and
+// base_discretisation_method="none" (the phantom is already discrete 1..6), which is what Nyxus
+// computes in IBSI mode; generator=tests/vetting/oracles/gen_ngtdm_mirp.py.
+//
+// All five features are bit-identical to mirp on every slice (3.2e-16 worst case, floating-point
```
vjaganat90: These values agree, but they are not bit-identical: the audit records nonzero Nyxus–MIRP residuals up to `3.2e-16`. Please describe this as floating-point-roundoff agreement and implement SPEC §7 exact as absolute `1e-9` (`ASSERT_NEAR`), not relative `agrees_gt(..., 1e9)`; align the generator and registry too.

tests/test_2d_ngtdm_mirp.h
```diff
@@ -0,0 +1,104 @@
...
+    {
+        const std::string key = feature_name + "_z" + std::to_string (z + 1);
+        SCOPED_TRACE ("mirp " + key);
+        ASSERT_TRUE (ngtdm_2d_mirp_slice_ref_vals.count(key) > 0) << key;
+        ASSERT_TRUE (agrees_gt (per_slice[z], ngtdm_2d_mirp_slice_ref_vals.at(key),
+                                ngtdm_2d_mirp_frac_tolerance)) << key;
+    }
+
+    // then the four-slice mean, the quantity IBSI publishes
+    assert_ngtdm_feature_against_golden_values (feature_, feature_name, ngtdm_2d_mirp_ref_vals,
```
vjaganat90: `per_slice` above is the vector this test just asserted, but this helper featurises all four slices again before taking the mean. Pass the asserted vector into the mean helper so both checks cover the same computation, as the current GLDZM test does.

tests/vetting/oracle_coverage.csv
```diff
@@ -301,11 +301,11 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
...m_2d,rel=1e-9,test_2d_ngtdm_ibsi.h;test_2d_ngtdm_mirp.h;test_2d_ngtdm_regression.h;test_2d_ngtdm_mechanics.h,,,,audit,"vetted vs mirp 2.6.0 feature extraction on the four IBSI phantom slices (by_slice=True, base_discretisation_method=none), recipe ngtdm.ibsi_phantom_2d; agreement 3.2e-16 worst case, asserted on each slice AND on the four-slice mean at rel=1e-9 in test_2d_ngtdm_mirp.h. PyRadiomics 3.0.1 -- the oracle this row used to name, against a target_test file that never existed -- was run on the same fixture and recipe and agrees with mirp to 1.6e-16, so the claim is corroborated rather than dropped; only mirp's values are pinned, a second identical table being redundancy and not coverage. The same quantities are pinned against the published IBSI consensus in test_2d_ngtdm_ibsi.h at rel=1e-2, the 3-significant-figure precision those values are published to. Nyxus' default mode bins to a fixed grey count, which no reference reproduces, and is drift-pinned only, in test_2d_ngtdm_regression.h. Re-measured 2026-08-20: this row previously carried an EMPTY tolerance. gen_ngtdm_mirp.py re-verifies every pin in the header it feeds; tests/vetting/audit/ngtdm_2d_mirp_vetting_report.md"
```
vjaganat90: This is not asking for another oracle; it is asking not to describe two configurations with one row. This row names `ngtdm.ibsi_phantom_2d` (`IBSI=true`, identity binning), while `test_2d_ngtdm_regression.h` runs `IBSI=false` with `NGTDMFeature::n_levels=100`, and `test_2d_ngtdm_mechanics.h` checks setting behavior rather than a reference value. Keep the vetted row tied to the MIRP/IBSI gtests, define a separate recipe such as `ngtdm.default_fbn100` with five `status=regression` rows for those snapshots, and keep mechanics out of both oracle assertions.

tests/vetting/config_recipes.md
```diff
@@ -158,3 +158,30 @@ chosen reference tool (SPEC 5). Oracle tests reference a recipe by id; this file
   `test_3d_glcm_dump_regression()`.
 - A different binCount from `glcm3d.pyradiomics_bincount20`, so the two benchmarks are not
   comparable to each other by construction.
+
+## ngtdm.ibsi_phantom_2d
```
vjaganat90: Please complete this family migration with `tests/vetting/matrix/ngtdm.md`. After rebasing onto current `main`, populate the registry’s assertion-level `test_name` fields and use the already-defined `ibsi_digital_phantom` benchmark for the MIRP/IBSI rows.

vjaganat90 (commented):
Overall looks good. Once you rebase and fix minor issues this is good to go.

tests/test_2d_ngtdm_mirp.h
darkclad: Fixed

tests/test_2d_ngtdm_mirp.h
darkclad: Fixed

tests/vetting/oracle_coverage.csv

tests/vetting/config_recipes.md
darkclad: Fixed

PR #444, opened 2026-08-20
test(2d radial): run the CellProfiler oracle the registry names, and record that it does not vet

darkclad:
Re-vets the 2D radial intensity distribution — `FRAC_AT_D`, `MEAN_FRAC`, `RADIAL_CV` — against the
oracle the registry has been naming for it since the tracker was seeded.

**The oracle does not vet the family, and this PR says so with the numbers.** These three rows are
the first in the series that read `status=regression` with an *empty* `oracle` column: honestly
unvetted rather than falsely labelled. The job was to find out whether CellProfiler
`MeasureObjectIntensityDistribution` could cover them. It cannot: **21 of the 24 (feature × bin)
values disagree by more than 1%**, and the 3 that agree are bins both tools leave empty. Nyxus
computes a different quantity under each of the three CellProfiler names it uses, and CellProfiler's
own help strings are pasted verbatim into `radial_distribution.h` above functions that do something
else.

No feature values change here. The three rows stay `regression`, now with the reason recorded, the
band measured, and a runnable generator that keeps the verdict honest in both directions.

---

## What was cleaned up

The family was one header holding the fixture, three golden vectors and the assertions, banded at
`ASSERT_NEAR(..., 1e-9)` **absolute** against goldens spanning 0 to 53 — `rel=2e-11` on the largest
value and vacuous on the zeros — with the registry `tolerance` column empty on all three rows.

It is now the four SPEC §2 kinds:

| file                            | what it holds                                         |
| ------------------------------- | ----------------------------------------------------- |
| `test_2d_radial_common.h`     | the fixture, shared by the three below                |
| `test_2d_radial_regression.h` | the 24 pins, at`agrees_gt(..., 1e9)` = `rel=1e-9` |
| `test_2d_radial_invariant.h`  | 4 properties that hold for any ROI                    |
| `test_2d_radial_mechanics.h`  | 3 guards on the wiring the 24 pins depend on          |

The band comes from measurement, not habit: a fresh Windows build reproduces all 24 values
**bit-for-bit**, checked by temporarily replacing the assertion with `actual[i] == golden_values[i]`
and rebuilding. `rel=1e-9` is now also in the registry's `tolerance` column.

`rel=1e-9` additionally makes an empty bin's exact `0.0` an *exact* comparison — a zero ground truth
gives `agrees_gt` a zero tolerance — where a `1e-9` absolute window let anything small through.

## What was demoted, and why

Nothing. Nothing was ever promoted. What changed is that `candidate_oracle` no longer names a tool
nobody had run, and `flag` moves from `oracle-identified` to `convention-mismatch` — the token the
tree already uses for "an oracle was tried and no external numeric reference agrees".

The evidence, in full in `tests/vetting/audit/radial_2d_cellprofiler_vetting_report.md`:

| step                 | CellProfiler                                        | Nyxus                                                 |
| -------------------- | --------------------------------------------------- | ----------------------------------------------------- |
| centre               | pixel of maximum distance-to-edge                   | pixel minimising (max − min) distance to the contour |
| distance from centre | geodesic inside the mask                            | straight-line Euclidean                               |
| radial coordinate    | `d_c / (d_c + d_edge)`, **per pixel**       | `d_c / r_max`, one global `r_max`                 |
| bin index            | `int(nd × n_bins)`                               | `int(nd × (n_bins − 1))`                          |
| `FracAtD`          | bin intensity ÷ ROI intensity                      | **bin pixel count ÷ ROI pixel count**          |
| `MeanFrac`         | dimensionless, ≈ 1                                 | **the bin's raw mean intensity**                |
| `RadialCV`         | CV of 8 wedge**means**, non-empty wedges only | CV of 8 wedge**sums**, all 8                    |

`MEAN_FRAC` is the clearest single reading: CellProfiler returns 0.6–1.3 on these bins because the
quantity is normalised by the ROI mean; Nyxus returns 22–53 because it *is* that mean, in absolute
units. `FRAC_AT_D` never reads the image at all — it is a function of the mask alone.

## Two findings that are not about radial

Both are recorded, measured, and deliberately not fixed: each changes public feature values.

**1. The traced contour is returned one pixel off in both axes.**
`ContourFeature::buildRegularContour` copies the ROI into a padded image at `px.x - base_x + 1`,
traces there, and converts back with `p.x += base_x` — the bounding-box origin is restored, the
one-pixel pad never is. On this ROI, **7 of the 18 contour pixels are not ROI pixels**, and **all 18
land back on a ROI pixel after shifting by (−1,−1)**. Both are asserted in
`test_2d_radial_mechanics.h`, so the correction cannot land silently.

Nothing has caught it because no test in the tree asserts a contour *coordinate* —
`test_2d_contour_analytic.h` asserts only how many contours come back. A uniform translation cancels
wherever the contour is only measured against itself (`PERIMETER`, Feret, the fractal divider walk,
the neighbour features), so what it breaks is the callers that mix contour coordinates with *pixel*
coordinates: this family, `ROI_RADIUS_{MEAN,MAX,MEDIAN}`, `DIAMETER_{INSCRIBING,CIRCUMSCRIBING}_CIRCLE`,
and every `WEIGHTED_*` moment. The `DIAMETER_*_CIRCLE` registry rows already carry a note about "the
centroid-1 offset" making a symmetric circle come out asymmetric; this is a candidate explanation for
it, and re-measuring those once the shift is gone is the first thing the fix branch should do.

**2. `Pixel2::find_center` and `Pixel2::max_sqdist` return non-extremal answers here.**
Both are coarse-to-fine searches that assume an ordered contour, and `merge_multicontour` concatenates
the outer contour with the hole's. Against a full linear scan of the same contour:

| quantity             | searched | exact |
| -------------------- | -------- | ----- |
| centre pixel         | (3,4)    | (4,4) |
| max squared distance | 10       | 13    |

The normalising radius is 23% short of the largest distance present, so 4 of the 26 pixels sit at
`r/r_max > 1` and are clamped. Combined with the `num_bins - 1` scaling, the last bin ends up holding
7 of the 26 pixels — more than any of the 7 real rings. `exact_min_sqdist()` already exists beside
`min_sqdist()` for this reason and is used by the neighbour features; there is no `exact_max_sqdist`.

## New audit artifacts

- `tests/vetting/oracles/gen_radial_cellprofiler.py` — runs the real module. It **holds no copy of
  anything it checks**: the fixture is parsed out of `test_data.h`, the pinned vectors out of the
  regression header and the centre and radius out of the mechanics header, by brace counting rather
  than a non-greedy regex. It verifies all 24 pins against a written-down model of the
  implementation, checks the reverse direction (a feature the model produces that nothing pins), and
  rebuilds CellProfiler's three measurements from the published definitions in numpy/scipy with no
  `centrosome` in the path — agreeing with the module to **6.47e-08**, which is CellProfiler storing
  the image as float32. It exits non-zero if a pin stops reproducing, if the rebuild stops
  reproducing the module, **or if CellProfiler ever starts agreeing**, because that would mean the
  three rows became promotable.
- `tests/vetting/audit/radial_2d_cellprofiler_vetting_report.md` — the run, both tables, the six
  divergences with their source lines, and why this fixture could not vet the family even after a fix.
- `tests/vetting/audit/radial_2d_golden_regen.md` — how to re-pin, how to re-run CellProfiler, the
  name mapping, and the three conditions that all have to hold before promotion is possible.
- `tests/vetting/audit/scan_radial_coverage.py` + `radial_2d_coverage.csv` — the coverage table
  derived from the sources rather than written by hand. Its `ORACLE_SUFFIX` is empty by design, so no
  function in this family can contribute an oracle token however it is named.
- `tests/vetting/TOOLS.md` — the headless-CellProfiler-on-Windows recipe, with the three things that
  stop the run before any module executes.
- `tests/vetting/MIGRATION.md` — the oracle-selection row that predicted this family would be
  "near-exact" against CellProfiler because Nyxus "copied verbatim from CP". What was copied is the
  three names and the three help strings; the row now says what the run found. That prediction is why
  the family was queued as the most promotable one left, so leaving it would send the next reader the
  same way.

**There is no second oracle to corroborate this one.** `TOOLS.md` lists `wndcharm` under
"Gabor / Tamura / radial", but its radial features are Radon-transform coefficients — a different
quantity — and it is a Python 2.7 source build. No other tool in the §4 catalog implements
`FracAtD` / `MeanFrac` / `RadialCV`, so CellProfiler's rejection settles the family rather than
opening a second comparison.

## Something the fixture cannot do

CellProfiler's centre is the pixel of maximum distance-to-edge. On `shape2d_morphology_mask` that
maximum is **attained by 8 of the 26 pixels**, so which pixel `scipy.ndimage.maximum_position`
returns depends on the label image's shape — CellProfiler's own answer moved when the padding around
the ROI changed while this was being written. The generator prints the tie set on every run.

A ROI with a unique distance-to-edge maximum, and with enough pixels that 8 radial bins are not
mostly empty, is a precondition for this family ever vetting against CellProfiler, independently of
the six divergences. 26 pixels over 8 bins leaves 3 of them empty on one side or the other.

## How it was verified

Not "tests pass":

| gate                                | result                                                                                                                                            |
| ----------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
| gtest, Windows                      | **856**, up from 849 by exactly the 7 cases added                                                                                           |
| ASan + UBSan gtest, Linux           | **856 pass, 0 diagnostics**, exit 0                                                                                                         |
| `pytest tests/python/`            | 86 passed, 1 skipped; the 7`@pytest.mark.arrow` failures are the tiff-only local build                                                          |
| `gen_radial_cellprofiler.py`      | all 24 pins reproduced, reverse check clean, range/identity checks over the pinned literals clean, rebuild agrees to 6.47e-08, divergence present |
| `check_test_names.py`             | clean                                                                                                                                             |
| `check_coverage.py --check`       | clean                                                                                                                                             |
| `scan_radial_coverage.py --check` | clean                                                                                                                                             |

The ASan run script verifies **five md5 checksums plus the registered case count** before it will
build, so the run is provably this tree and not a leftover of a previous one. The five it reported
match the files in this diff.

**Negative controls, both run and both recorded in the report.** For the per-bin regression:
swapping the `FRAC_AT_D` bin-3 and bin-6 goldens leaves the sum exactly 1 and every bin count a whole
number, so all four invariants still pass while the per-bin assertion fails naming `FRAC_AT_D[3]` —
the aggregate is blind and the per-element assertion is not. For the invariants: `FRAC_AT_D[0] += 0.5`, `MEAN_FRAC[2] = 1000`, `MEAN_FRAC[2] += 1` and `RADIAL_CV[2] = 3` each fire their own invariant
and nothing else.


tests/vetting/oracles/gen_radial_cellprofiler.py
```diff
@@ -0,0 +1,444 @@
...
+    for f in FEATURES:
+        for i in range(NBINS):
+            r = rel(cp[f][i], pins[f][i])
+            if r > 1e-2:
+                diverging += 1
+            print("  %-10s %3d %22.17g %22.17g %10.3g" % (f, i, pins[f][i], cp[f][i], r))
+
+    print("\n  %d of the %d (feature x bin) values disagree by more than 1%%."
+          % (diverging, 3 * NBINS))
+    if diverging == 0:
```
vjaganat90: Promotion is decided per **feature × config**, not once for the whole family. This single counter fails only when all 24 cells agree. For example, all eight `FRAC_AT_D` bins could move within the oracle tolerance while one `MEAN_FRAC` bin still differs; this script would exit 0 and leave a now-promotable `FRAC_AT_D` row marked regression. Track agreement separately for each feature and fail when every bin of any one feature meets its applicable SPEC band. The 1% cutoff can remain a descriptive report threshold, but it should not be the promotion gate.

tests/vetting/oracle_coverage.csv
```diff
@@ -311,11 +311,11 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
-2D,FRAC_AT_D,radial,regression,,na,"Not mode-specific. Any disagreement is a definition, coordinate-frame, spacing, aggregation, or implementation issue rather than Nyxus IBSI/radiomics mode.",,test_2d_radial_regression.h,test_2d_radial_regression.h,cellprofiler:MeasureObjectIntensityDistribution (RadialDistribution_*),oracle-identified,tracker,,,
+2D,FRAC_AT_D,radial,regression,,disagreed,radial.shape2d_native,rel=1e-9,test_2d_radial_regression.h;test_2d_radial_invariant.h;test_2d_radial_mechanics.h,,"cellprofiler:MeasureObjectIntensityDistribution (RadialDistribution_*) -- run 2026-08-20, does not vet",convention-mismatch,in-tree,"Not CellProfiler-vettable: CP's FracAtD is the fraction of the ROI's intensity in a bin, Nyxus' is the fraction of its pixel count and never reads the image. Full run, numbers and the six divergences in tests/vetting/audit/radial_2d_cellprofiler_vetting_report.md; generator tests/vetting/oracles/gen_radial_cellprofiler.py. 8-bin vector, every bin asserted separately at rel=1e-9 against a measured residual of 0; invariants in test_2d_radial_invariant.h, wiring in test_2d_radial_mechanics.h.",,
```
vjaganat90: The rebase added the assertion-level `test_name` and `benchmark` columns, but all three migrated radial rows leave both empty. The exact regression case is already known (`TEST_NYXUS.TEST_2D_RADIAL_DISTRIBUTION_REGRESSION`), and the 8×8 `shape2d_morphology_*` fixture should receive a benchmark entry in `benchmarks.md`. Please populate these on all three rows. If the invariant cases are intentionally credited through `current_test`, list their exact gtest names as well; a semicolon-separated file list is not a substitute for identifying the assertions.

tests/test_2d_radial_invariant.h
```diff
@@ -0,0 +1,111 @@
...
+	ASSERT_EQ(frac.size(), (size_t)RadialDistributionFeature::num_bins) << "FRAC_AT_D";
+
+	double sum = 0.0;
+	for (size_t i = 0; i < frac.size(); i++)
+	{
+		ASSERT_GE(frac[i], 0.0) << "FRAC_AT_D[" << i << "]";
+		ASSERT_LE(frac[i], 1.0) << "FRAC_AT_D[" << i << "]";
+		sum += frac[i];
+
+		const double count = frac[i] * double(roidata.raw_pixels.size());
```
vjaganat90: Only part of this file is definition-independent. The `[0,1]`/sum-to-one checks and the non-negative CV bound are genuine invariants, but requiring `FRAC_AT_D × pixel_count` to be integral encodes Nyxus's current pixel-count convention. The raw-intensity range and total-intensity reconstruction checks for `MEAN_FRAC` below likewise encode the exact semantics this PR shows differ from CellProfiler. Keep those as regression/characterization checks or remove them; calling them invariants would make a correct semantic change fail a test advertised as universally required. Also, the header cites nonexistent “SPEC 4.4”; the taxonomy is SPEC §2.

tests/test_2d_radial_mechanics.h
```diff
@@ -0,0 +1,140 @@
...
+{
+	LR roidata(101);
+	build_radial_2d_roi(roidata);
+
+	std::vector<Pixel2> K;
+	roidata.merge_multicontour(K);
+
+	int idxO = Pixel2::find_center(roidata.raw_pixels, K);
+	const Pixel2& pxO = roidata.raw_pixels[idxO];
+	ASSERT_EQ(pxO.x, 3);
```
vjaganat90: This suite deliberately freezes the shifted contour, non-extremal centre, and short radius that the report identifies as defects. The 24-value regression already catches their public output impact. If these checks are retained for diagnosis, describe them explicitly as known-defect characterization, link the tracked corrective work, and do not credit this mechanics file as feature coverage in the three registry rows. Otherwise the test suite turns `(3,4)` and radius² `10` into acceptance criteria that a correct fix must violate.

tests/vetting/audit/radial_2d_cellprofiler_vetting_report.md
```diff
@@ -0,0 +1,228 @@
...
+`ROI_RADIUS_MEDIAN` demotion records the same approximation as a second defect underneath its units
+bug (`morphology_2d_skimage_vetting_report.md`).
+
+**3. The bin index is scaled by `num_bins - 1`.**
+`radial_distribution.cpp:81` computes `int(rat * (n-1))` and clamps to `n-1`. With `n = 8` that is 7
+rings of width 1/7 plus a last bin nothing reaches unless `r >= r_max`. Combined with defect 2 the
+last bin ends up holding 7 of the 26 pixels — 3 sitting exactly on `r_max` and the 4 beyond it —
+more than any of the 7 real rings. CellProfiler scales by `bin_count` and gets 8 equal rings.
+
+**4. `FRAC_AT_D` returns a pixel-count fraction.**
```
vjaganat90, line 193: This report establishes that the production header describes CellProfiler's quantities while the functions return different ones, but the PR leaves `src/nyx/features/radial_distribution.h` unchanged. Updating those comments is behavior-free and directly part of recording the result honestly: document `FRAC_AT_D` as a pixel-count fraction, `MEAN_FRAC` as the raw bin mean, and `RADIAL_CV` as the CV of wedge sums including empty wedges (or explicitly mark the intended semantics unresolved). The source should not continue advertising the rejected oracle definitions after this audit lands.

tests/vetting/oracles/gen_radial_cellprofiler.py
```diff
@@ -0,0 +1,444 @@
...
+and the centre pixel and normalising radius out of tests/test_2d_radial_mechanics.h,
+so "ALL CHECKS PASSED" cannot mean the script agrees with its own copy.
+
+Environment: a dedicated CellProfiler env is required (cellprofiler 4.2.8 +
+cellprofiler-core + centrosome, headless, python 3.9). CI never invokes it --
+CellProfiler is not a runtime dependency. On Windows the interpreter's working
+directory must be on the SAME DRIVE as the cellprofiler package or importing the
+module raises ValueError from os.path.relpath; see tests/vetting/TOOLS.md.
+
+Provenance: tool=cellprofiler, version=4.2.8 (module package) / cellprofiler-core
```
vjaganat90, line 27: The script records CellProfiler 4.2.8 here but never verifies or prints the installed package versions. A later environment can therefore produce a run still presented as 4.2.8. Read the actual `cellprofiler`/`cellprofiler-core` versions via package metadata and either report them in the output or reject an unexpected version so the audit provenance remains truthful.

vjaganat90 (commented):
The direct CellProfiler result is useful and the central verdict is correct: none of the three current Nyxus vectors is CellProfiler-vetted. The remaining issues are about keeping that verdict SPEC-accurate: promotion must be detected per feature × config, the migrated registry rows must identify their actual assertions and fixture, and tests labeled as invariants/mechanics must not turn the _implementation differences and known defects documented here into correctness requirements_. Details inline.

tests/vetting/oracles/gen_radial_cellprofiler.py
darkclad: Fixed

tests/vetting/oracle_coverage.csv

tests/test_2d_radial_invariant.h
darkclad: Fixed

tests/test_2d_radial_mechanics.h
darkclad: Fixed

tests/vetting/audit/radial_2d_cellprofiler_vetting_report.md
darkclad: Fixed

tests/vetting/oracles/gen_radial_cellprofiler.py
darkclad: Fixed

PR #445, opened 2026-08-21
test(3d morphology): add MATLAB oracle for three volume features

vjaganat90:
## Summary

- add `tests/vetting/oracles/gen_morphology3d_matlab.m`, retrieving `ut_mask57.nii` through the [`PolusAI/nyxus` main fixture permalink](https://github.com/PolusAI/nyxus/blob/main/tests/data/nifti/phantoms/ut_mask57.nii) and calling MATLAB built-ins `niftiread` and `regionprops3`
- restore `test_3d_morphology_matlab.h` with three assertions matching the existing MIRP coverage: `3VOXEL_VOLUME`, `3VOLUME_CONVEXHULL`, and the current implementation of `3MESH_VOLUME`
- add separate MATLAB registry rows for those feature/config pairs
- share the existing MIRP tolerances across the MATLAB and MIRP assertions: `rel=1e-3` for voxel volume and 5% for both hull-valued features
- pin the MATLAB provenance and config recipe to R2026a
- make the coverage rollup count unique features when multiple oracle rows cover one feature

## MATLAB–MIRP oracle comparison

| Nyxus assertion | MATLAB direct property | MATLAB | MIRP reference quantity | MIRP | MATLAB vs MIRP |
|---|---|---:|---|---:|---:|
| `3VOXEL_VOLUME` | `regionprops3.Volume` | 274432 | `morph_vol_approx` | 274432 | exact |
| `3VOLUME_CONVEXHULL` | `regionprops3.ConvexVolume` | 497824 | `morph_volume / morph_vol_dens_conv_hull` | 496958.32 | 0.17% |
| `3MESH_VOLUME` | `regionprops3.ConvexVolume` | 497824 | same MIRP convex-hull quantity | 496958.32 | 0.17% |

MATLAB produces two values, not three. Both tools provide separate assertions for the hull-valued output currently returned for `3MESH_VOLUME`. That follows the existing implementation, [`fval_MESH_VOLUME = fval_VOLUME_CONVEXHULL`](https://github.com/PolusAI/nyxus/blob/main/src/nyx/features/3d_surface.cpp#L485); neither assertion validates it as a true mesh volume.

MATLAB remains an offline golden generator, not a CI runtime dependency.


darkclad:
Comments on files not present in the PR, or on file lines that I cannot comment on
tests/vetting/check_coverage.py:63-75	§3.1's "vetted by ≥2 oracles" line is missing, and this PR is what makes it non-zero
tests/vetting/SPEC.md:152	still says R2025b; the "legacy tables" paragraph no longer describes reality
tests/vetting/MIGRATION.md:315-316	describes a file that no longer looks like that
tests/test_3d_morphology_regression.h:20-21	"only executable assertion is the MATLAB one" — now two

vjaganat90:
Addressed:

- The `>=2` rollup is intentionally not added; `>=1` remains the §3.1 headline/acceptance metric.
- The stale R2025b SPEC wording is already fixed on the rebased branch.
- `MIGRATION.md` §5.16 intentionally records the Wave 8 state; the document explicitly preserves earlier wave descriptions instead of rewriting them to match the current tree (§5.37).
- The regression-header comment is now fixed to name both MATLAB and MIRP oracle assertions.

tests/test_3d_morphology_common.h
```diff
@@ -16,10 +16,22 @@
...
+static constexpr double morphology_3d_hull_volume_ref_tol_pct = 5.0;
+
+static double morphology_3d_volume_ref_tol_pct (const std::string& fname)
+{
+    // Callers first require fname to exist in their three-entry volume golden table.
+    return fname == "3VOXEL_VOLUME"
+        ? morphology_3d_voxel_volume_ref_tol_pct
+        : morphology_3d_hull_volume_ref_tol_pct;
+}
+
```
sameeul: this is way to verbose, can we just a simple one liner, since there is no point of having those global variables
.Variables declared at  line 22 and 23 are not going to used anywhere and so should be scoped only inside that function. 

On further reading, the better solution is to bring `static const ref_vals_map<double> morphology_3d_mirp_volume_ref_tols` here and rename it to `static const ref_vals_map<double> morphology_3d_mirp_matlab_volume_ref_tols` and use it.

tests/test_3d_morphology_mirp.h
```diff
@@ -54,62 +53,51 @@ static const ref_vals_map<double> morphology_3d_mirp_pca_ref_vals
...
+// The assertion judges the alias against the quantity it actually computes and does not validate it
+// as a true mesh volume. If 3MESH_VOLUME becomes a real mesh integral, this golden must be revisited.
 static const ref_vals_map<double> morphology_3d_mirp_volume_ref_vals
 {
     { "3VOXEL_VOLUME", 274432.0 },               // morph_vol_approx
     { "3VOLUME_CONVEXHULL", 496958.3201121965 }, // morph_volume / morph_vol_dens_conv_hull
     { "3MESH_VOLUME", 496958.3201121965 }        // the same hull volume: Nyxus aliases MESH_VOLUME to it
 };
 
-// Bands as the percent relative_absdiff may reach, measured per feature (SPEC 7).
```
sameeul, line 88: see the previos comment

tests/test_3d_morphology_common.h
vjaganat90: Done. Removed the two globals and reduced this to one shared helper. Both MIRP and MATLAB call that same helper, so the per-feature bands remain identical.

tests/test_3d_morphology_mirp.h
vjaganat90: Addressed by the same change; this assertion now reads the shared MIRP/MATLAB tolerance helper.

PR #448, opened 2026-08-24
test(3d ngldm): retire the coverage sweep, which was a bit-for-bit duplicate

darkclad:
Follow-up to #436 (3D NGLDM, merged), doing for that family what #446 did for 3D morphology and #441
for 3D GLCM: the generic `TEST_P` coverage sweep is retired once everything it checked has a named
home. Based on `upstream/main` at `9703fe3`, the merge of #436.

## This is the cleanest case in the series: the table was a literal duplicate

`test_3d_ngldm_coverage.h` instantiated the two shared suites over the family's 19 features. Both
halves were already covered:

| what the sweep did                                            | where it lives now                                                                                                                                              |
| ------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `NGLDM_WITH_3P_EMBEDDED_GT`                                 | **instantiated zero cases** — no NGLDM table appears in `externally_vetted_3d_feature_names()`, because the family has no oracle-backed feature at all |
| `ngldm_3d_regression_coverage_ref_vals`, 19 pins            | `ngldm_3d_regression_ref_vals` — **the same 19 keys with the same 19 doubles**, verified by parsing both tables and comparing, not by eye              |
| "the feature name resolves and the code matches"              | every named test:`find_3D_FeatureByString` plus an assert on the returned code                                                                                |
| "every registered`Feature3D` code has exactly one provider" | `FeatureManager::check_11_correspondence()`, in production since #441                                                                                         |

Both tables were taken at the same recipe — the sweep runs `GREYDEPTH=64`, `IBSI=false`, and so does
`test_3d_ngldm_regression.h`. Unlike 3D GLCM's grey64 table there was no second configuration to
preserve, and unlike 3D morphology nothing had to be ported: #436 had already regenerated the
regression table to full precision, and all 19 features already had a named `*_regression` test and a
`TEST()` registration.

**The migration cost exactly one `add_keys()` line**, which is what #446 rebuilt that guard to make
true. Negative control: delete `3NGLDM_DCV` from the table and `TEST_3D_FEATURE_COVERAGE_COUNTS`
fails with *"3NGLDM_DCV (family ngldm) has no oracle and no drift guard"*, naming the feature.

## The generator could not fail, and now can

`revet.txt` §9 says a generator must re-verify the artifact it feeds and exit non-zero on a mismatch —
and warns to *check it actually opens the file*. `gen_ngldm3d_mirp.py` pointed its verification at
`test_3d_ngldm_mirp.h`, **a header that has never existed** (that dangling name is the very defect
#436 closed in the registry). So it always took the "nothing to verify" branch and returned 0, with
the entire pin-checking body below it unreachable. Its `ALL CHECKS PASSED` meant only that the script
had run.

Nothing in the tree *pins* these MIRP values — the family is regression-only and the oracle is
deliberately unasserted — so the artifact the generator actually feeds is the comparison table in
`ngldm_3d_mirp_vetting_report.md`. It now verifies that: every MIRP value quoted there against a
fresh run at `rel<=1e-5` (the six significant figures the report quotes), plus the reverse direction
for a feature MIRP produces that the report omits, exiting non-zero on either. A missing report is
now a failure rather than a pass.

Run against MIRP 2.6.0: **17 verified, 0 failed, 0 unproducible, 0 unquoted.**

## The over-count is measurable from the pins, and it is not the bounding box

`f_GLNU /= Ns` and `f_GLNUN /= (Ns*Ns)` in `3d_ngldm.cpp`, so `GLNU/GLNUN` is exactly `Ns` — the
number of voxels that contributed a matrix entry. `DCNU/DCNUN` is the same quantity by the same
construction. Both ratios agree:

```
GLNU/GLNUN = 511360.000000
DCNU/DCNUN = 511360.000000     rel 2.3e-16
```

**Ns = 511,360**, and 511,360 = 68×80×94 is exactly the *interior* of the 82×96×70 bounding box the
report measures at 551,040 — the one-voxel shell, whose 26-neighbourhood would leave the box, is
skipped. Two independent routes to the same box, so this is a measurement rather than a coincidence.

That refines #436's finding rather than contradicting it — the matrix is still built over far more
than the ROI — but it changes a number the report leans on. The over-count is
**511,360 / 274,432 = 1.8633×**, not 2.008×, so `DCNU`'s 2.09× is *the nearest of the seventeen to
the voxel-count ratio and still 12% above it*, not "almost exactly the bounding-box-to-ROI ratio".
Corrected in the report and the regen doc.

## Hygiene, from the `revet.txt` §9 audit

- **The family's fixture used four headers it did not include.** This family has no `_common.h`, so
  `test_3d_ngldm_regression.h` *is* the fixture, and it used `Fsettings`/`NyxSetting` (20 sites),
  `SlideProps`/`scan_slide_props`, the four `globals.h` ROI-gathering functions, and `agrees_gt`
  with none of `feature_settings.h`, `slideprops.h`, `globals.h` or `test_main_nyxus.h` included —
  every one transitive. The same finding #446 made against 3D morphology's `_common.h`; #436's
  include commit fixed the std headers and stopped there. All four now direct, `<iostream>` dropped
  because `test_main_nyxus.h` supplies it.
- **The reference table is `static const` and read through `.at()`.** #436 guarded the
  default-insert hazard by checking `find()` earlier in the same function while the read stayed
  `operator[]` — which works, but leaves the hazard one edit away. `const` + `.at()` removes it by
  construction, and matches the rule from `code-review-818.md` finding 6.
- **History language trimmed** from the tolerance comment ("the previous 10% band sat on two- and
  three-significant-figure pins"), which is already in the regen doc, per `revet` §6.
- The failing assertion now prints the feature name and the full-precision actual, as 3D
  morphology's does.

## Verification

- **gtest, MSVC Release/Ninja, DCMTK on, GPU/Arrow/Z5 off** — `upstream/main` baseline **846/846**;
  this branch **827/827**. Delta −19, exactly the retired `NGLDM_UNVETTED_LOCAL_REGRESSION` cases
  (the embedded half was empty). No tests added, because #436's 19 named tests plus its dump already
  covered the family.
- **ASan + UBSan, Linux gcc `RelWithDebInfo -O1`** — **823/823 pass**, zero sanitizer reports.
  823 = 827 − 4, the `TEST_2D_HU_LOADER_DICOM_*` cases compiled out under `NOEXTRAS=ON`, compared
  against this branch's own Windows count rather than a remembered offset.
- **`pytest tests/python/`, whole directory** — 88 passed, 1 skipped, 7 failed; all 7 are the
  `@pytest.mark.arrow` cases that need an Arrow-enabled build, which this one is not.
- **Mechanical checks over all 19 pins** — the `GLNU/GLNUN` = `DCNU/DCNUN` normalisation identity
  (rel 2.3e-16); `DCP`, `DCENE`, `LDE`, `LGLCE` in [0,1]; `HDE`, `HGLCE` ≥ 1; `DCENT`, `GLV`, `DCV`,
  `GLM`, `DCM` ≥ 0; and low-≤-high on each emphasis axis. All pass.
- **Negative control** on the guard, as above.
- `gen_ngldm3d_mirp.py` against MIRP 2.6.0: 17/17, exit 0 — and it can now return 1, which it could
  not before.
- `check_coverage.py --check`, `check_test_names.py`, and all six family scanners clean;
  re-running `scan_ngldm3d_coverage.py` reproduces its CSV byte-for-byte.


tests/vetting/audit/ngldm_3d_golden_regen.md
```diff
@@ -23,11 +23,51 @@ ROI bounding box instead of the ROI, and a 24-shift neighbourhood where 3D Cheby
...
 Recipe `ngldm3d.mirp_fbn64`. Nothing in the tree asserts against it today; it exists so the
 divergence stays reproducible and so the promotion can be re-run against a fixed implementation.
 
+**It verifies the report, because that is the artifact it feeds.** The generator used to point its
+re-verification at `test_3d_ngldm_mirp.h`, a header that has never existed, so it always took the
+"nothing to verify" branch and exited 0 — its `ALL CHECKS PASSED` meant only that the script ran. It
+now parses the comparison table in `ngldm_3d_mirp_vetting_report.md`, checks every MIRP value quoted
+there against a fresh run at `rel<=1e-5` (the six significant figures the report quotes), checks the
+reverse direction for a feature MIRP produces that the report omits, and exits non-zero on either.
+Last run: 17 verified, 0 failed, 0 unproducible, 0 unquoted.
```
vjaganat90, line 63: The existing paragraph immediately below this new section (lines 75–77) still describes the old behavior: it says the generator checks `test_3d_ngldm_mirp.h`. Please update it to say the generator verifies the MIRP values in `ngldm_3d_mirp_vetting_report.md`, so these regeneration instructions do not contradict themselves.

tests/vetting/oracles/gen_ngldm3d_mirp.py
```diff
@@ -146,32 +161,38 @@ def main():
...
+    if not os.path.exists(REPORT):
+        print(f"\n# {os.path.basename(REPORT)} is missing -- there is nothing this run can be "
+              f"checked against, which is a failure, not a pass")
+        return 1
 
-    pins = parse_pins(open(TEST_H, encoding="utf-8", errors="replace").read(),
-                      "ngldm_3d_mirp_ref_vals")
-    print(f"\n# verifying {len(pins)} pinned goldens against this run")
+    quoted = parse_report(open(REPORT, encoding="utf-8", errors="replace").read())
+    print(f"\n# verifying the {len(quoted)} MIRP values quoted in "
```
vjaganat90, line 179: This verification can run under any installed MIRP, but line 157 always labels the run as MIRP 2.6.0. Please print the installed version—the `gen_morphology3d_mirp.py` pattern already handles `__version__`/package metadata—so the generated provenance remains truthful.

tests/test_3d_ngldm_regression.h
```diff
@@ -2,14 +2,17 @@
...
-#include "../src/nyx/roi_cache.h"
-#include "../src/nyx/features/3d_ngldm.h"
-#include "../src/nyx/helpers/fsystem.h"
-#include "test_ref_vals.h"   // ref_vals_map, and the <string> / <vector> it already includes
+#include "../src/nyx/environment.h"           // Environment
+#include "../src/nyx/feature_settings.h"      // Fsettings, NyxSetting
+#include "../src/nyx/featureset.h"            // Nyxus::Feature3D
+#include "../src/nyx/globals.h"               // clear_slide_rois, gatherRoisMetrics_3D, scanTrivialRois_3D, allocateTrivialRoisBuffers_3D
+#include "../src/nyx/roi_cache.h"             // LR
+#include "../src/nyx/slideprops.h"            // SlideProps, scan_slide_props
```
vjaganat90, line 11: `feature_settings.h`, `globals.h`, and `slideprops.h` are redundant here: retained headers already provide them (`environment.h`/`3d_ngldm.h`, `test_main_nyxus.h`, and `roi_cache.h`, respectively). Please keep this include block minimal.

vjaganat90 (commented):
The sweep retirement itself is sound: the deleted table exactly matches the 19 named regression pins, the oracle suite instantiated zero cases, and the completeness guard retains feature-name coverage. In the PR description, please change ‘two independent routes’ to ‘two algebraic consistency checks’—both ratios recover the same shared `Ns`—and ‘the generator could not fail’ to ‘the generator could not detect a stale report,’ since extraction/runtime failures were already possible.

tests/vetting/audit/ngldm_3d_golden_regen.md
darkclad: Fixed

tests/vetting/oracles/gen_ngldm3d_mirp.py
darkclad: Fixed

tests/test_3d_ngldm_regression.h
darkclad: Fixed

PR #450, opened 2026-08-24
test(3d firstorder): clean up PR 449 oracle provenance

vjaganat90:
## Summary

PR #449 did the important foundation work: it made the 3D first-order family runnable, preserved
the PyRadiomics comparison, and exposed how much of the old table had never been exercised. This
follow-up keeps that useful work and tightens the one place where the resulting claim was stronger
than its evidence.

The revised PR:

- replaces the reconstructed GNU Octave path with a checked-in MATLAB R2026a generator;
- vets 29 features from native MATLAB statistics, with only the feature-defining arithmetic applied
  to built-in results;
- retains all 17 native PyRadiomics assertions from #449;
- moves the two MATLAB semantic mismatches to explicit regression tests;
- folds in #459 by retiring the redundant first-order coverage sweep;
- corrects the registry, recipe, benchmark, matrix, audit, and coverage documents to match the
  assertions that actually run.

No production feature code changes.

## Why this follow-up is needed

The issue with #449 is not that Octave produced numerically poor answers. The 29 shared values agree
with the MATLAB R2026a values to floating-point precision. The issue is what generated some of the
remaining values.

The Octave generator manually:

- formed an inclusive P10/P90 subset and calculated robust statistics over it;
- rebuilt equal-width histogram bins and evaluated entropy and uniformity formulas;
- calculated Nyxus mean deviation about the median while separately showing that native
  `mad(x,1)` has different semantics.

Those are independent reimplementations, not native MATLAB/Octave feature calls. They may be useful
diagnostic calculations, but they should not be registered as direct `oracle=matlab` evidence under
SPEC. This PR removes that ambiguity instead of arguing around it.

## Exact oracle boundary

| | Count |
|---|---:|
| Nyxus 3D first-order features | **36** |
| Vetted by at least one oracle | **33** |
| Regression-only | **3** |
| MATLAB R2026a feature/config assertions | **29** |
| PyRadiomics 3.0.1 feature/config assertions | **17** |
| Features covered by both tools | **13** |

### MATLAB R2026a

The checked-in generator uses the repository fixture at `PolusAI/nyxus/main`, reproduces the
documented float-NIfTI loader domain as fixture setup, and then calls MATLAB built-ins.

Twenty-two features map directly to `sum`, `iqr`, `kurtosis`, `max`, `mean`, `mad`, `median`, `min`,
`mode`, `prctile`, `range`, `rms`, `skewness`, `std`, or `var`.

The remaining seven apply only the feature definition to built-in results:

- `std / mean` for COV;
- `std / sqrt(n)` for standard error;
- `kurtosis - 3` for excess kurtosis;
- QCOD from the two MATLAB quartiles;
- PIU from MATLAB `min` and `max`;
- standardized fifth and sixth MATLAB central moments.

There is no hand-built histogram, percentile estimator, robust window, entropy, or uniformity
implementation in the MATLAB generator.

### PyRadiomics retained from #449

PyRadiomics remains the direct native oracle for 17 features. Four are intentionally absent from
the MATLAB table and are covered only by PyRadiomics:

- `3ENERGY`
- `3ENTROPY`
- `3ROBUST_MEAN_ABSOLUTE_DEVIATION`
- `3UNIFORMITY`

This is the strongest part of #449, and this PR preserves it.

### Regression-only boundary

Three features make no external correctness claim:

| Feature | Why |
|---|---|
| `3COVERED_IMAGE_INTENSITY_RANGE` | Nyxus-specific slide/ROI ratio with no matching external feature; its current cross-domain bound violation remains documented. |
| `3MEDIAN_ABSOLUTE_DEVIATION` | MATLAB `mad(x,1)` takes the median absolute deviation, while Nyxus averages deviations about the median. |
| `3ROBUST_MEAN` | MATLAB `trimmean` trims samples by rank, while Nyxus retains values inside histogram-derived P10/P90 thresholds. |

These are named regression pins in `test_3d_firstorder_regression.h`. They guard behavior without
manufacturing an oracle claim.

## Tolerances

The MATLAB assertion uses the existing SPEC tiers:

- `rel=1e-3` for same-definition statistics on the same integer voxel vector;
- `rel=1e-2` for the percentile group, where MATLAB evaluates sample percentiles and Nyxus uses its
  fixed 100-bin interpolated CDF.

The largest measured MATLAB residual is `2.30e-3` on `3P01`, comfortably inside the shared
one-percent percentile band. Values within the declared relative tolerance agree; the test does not
demand bit identity from different percentile estimators.

The PyRadiomics assertions retain their measured bands from #449: `rel=1e-9` for same-definition
values, `rel=1e-2` for percentile-derived values, and `rel=1e-3` for the population-versus-sample
variance comparison.

## SPEC and repository cleanup

- Checks in the licensed MATLAB R2026a generator and full-precision pins.
- Removes the obsolete Python Octave wrapper and the stale 35-feature Octave audit.
- Records every feature × config × oracle assertion separately in `oracle_coverage.csv`.
- Adds the missing `matrix/firstorder3d.md` disposition.
- Updates recipes, benchmarks, audit coverage, the generated rollup, and the provenance exception
  in SPEC.
- Keeps oracle values beside their owning assertions and regression values in the regression header.
- Retires `test_3d_firstorder_coverage.h`; its duplicate parameterized pass added no independent
  evidence.
- Removes the now-empty shared unvetted `TEST_P`, its dead baseline registry, and the three empty
  family instantiations; named regression tests now exclusively own unvetted drift guards.
- Keeps the global 3D accounting guard strict and points it at the owning MATLAB, PyRadiomics, and
  regression tables.

**Why are shared and non-first-order coverage files in this diff?** Retiring
`test_3d_firstorder_coverage.h` removes the last non-empty instantiation of the shared unvetted
`TEST_P`. `test_3d_coverage_common.h` owns that suite, while the GLDM, GLSZM, and NGTDM coverage
headers retained only empty instantiations of it. Removing this dead machinery prevents GoogleTest
from reporting an orphaned suite; no GLDM/GLSZM/NGTDM assertions, goldens, or oracle behavior change.

## Review scope

The branch is rebased directly onto the merge of #449 and contains two commits:

1. replace reconstructed Octave coverage with native MATLAB assertions and fold in #459;
2. correct the 3D first-order vetting records and audit trail.

This is a focused evidence-quality cleanup of #449: it preserves the useful oracle work, removes
overclaims, and leaves the family at an honest **33/36 features vetted by at least one external
oracle**.


sameeul:
#463

PR #453, opened 2026-08-25
fix(3d ngtdm): default the neighbourhood radius so the family is not NaN

darkclad:
# 3D NGTDM — vetting pass

All five 3D NGTDM features were already recorded as `vetted` against PyRadiomics, and a fresh run
says they are: Nyxus reproduces PyRadiomics 3.0.1 **bit for bit** on the fixture the test header
names — identical doubles, not agreement to 1e-15. So no row changes status.

What did not hold up was everything around that claim: the band the values were asserted at, the
recipe they recorded, the matrix they are all derived from, the drift-guard file that had never been
compiled, and the fact that **the family returns NaN in the configuration a default run reaches**.

|                                                 | before                                           | after                                            |
| ----------------------------------------------- | ------------------------------------------------ | ------------------------------------------------ |
| oracle band                                     | `agrees_gt(..., 10.)` = ±10%                  | `rel=1e-9`                                     |
| matrix-correctness band                         | `agrees_gt(..., 1)` = **±100%**         | `rel=1e-9`                                     |
| pytest band                                     | `rtol=1e-1, atol=1e-2`                         | `rtol=1e-9, atol=0`                            |
| NGTDM matrix on the phantom                     | not asserted                                     | 18 values pinned, per grey level                 |
| `test_3d_ngtdm_regression.h`                  | 5 functions, 0 registrations, never`#include`d | wired in, re-pinned at`%.17g`                  |
| `NGTDM_RADIUS` with no `--3ngtdm/radius`    | **0 → all five features NaN**             | 1, held by a mechanics guard                     |
| `config_recipe` / `tolerance` on the 5 rows | empty                                            | `ngtdm3d.pyradiomics_binwidth1` / `rel=1e-9` |

---

## The one source change: 3D NGTDM is NaN by default

`Environment::compile_feature_settings()` zero-fills each family's settings vector and then writes
back the entries whose zero value is degenerate. It covers `GLCM_OFFSET`, `GLCM_GREYDEPTH` and
`GLCM_NUMANG`. It did not cover `NGTDM_RADIUS`.

`NGTDM_RADIUS` is the Chebyshev radius of the neighbourhood. At 0 the scan in `gather_zones()` runs
exactly one iteration — the voxel of consideration, which it skips — so no voxel is recorded as
having a neighbour, the matrix stays empty, `Nvc = 0` and every `P[i]` is `0/0`. Measured against
the settings vector the environment actually compiles:

```
[PROBE] default NGTDM_RADIUS=0 NGTDM_GREYDEPTH=0
[PROBE] fcode 699 = -nan(ind)      3NGTDM_COARSENESS
[PROBE] fcode 700 = -nan(ind)      3NGTDM_CONTRAST
[PROBE] fcode 701 = -nan(ind)      3NGTDM_BUSYNESS
[PROBE] fcode 702 = -nan(ind)      3NGTDM_COMPLEXITY
[PROBE] fcode 703 = -nan(ind)      3NGTDM_STRENGTH
```

Nothing caught it because every assertion in the tree sets the radius itself — the C++ tests build
their own `Fsettings`, and the pytest case calls `set_metaparam("3ngtdm/radius=1")`. One line,
beside the GLCM default it mirrors, plus `test_3d_ngtdm_default_radius_mechanics`, which compiles
the real settings vector and requires the five values to be finite. 2D NGTDM does not read this
setting, so no 2D value moves.

**This is why the PR is titled `fix`.** It is the only behaviour change on the branch and the only
thing here a user would notice; everything else is `test`, `chore` and `docs`. See "Changelog" below.

---

## What was cleaned up

**One shared fixture header.** Both files opened with the same eight includes and the same
twenty-line mock 3D workflow. `test_3d_ngtdm_common.h` now holds the phantom accessor, the settings
recipe and that workflow; both files include it and repeat none of its includes. The audit found
that `agrees_gt` was never included by either file — they compiled because `test_all.cc` includes
`test_main_nyxus.h` earlier in the translation unit — and the same was true of `gatherRoisMetrics_3D`,
`scanTrivialRois_3D`, `allocateTrivialRoisBuffers_3D` and `clear_slide_rois` from `globals.h`.
`raw_nifti.h` was included and unused.

**Provenance that described a different run.** The header recorded

```
pyradiomics mri.nii.gz liver.nii.gz --param settings1.yaml
```

against `ut_inten.nii` as both image and mask — two files the test does not load — and the sentence
"(100 grey levels, offset 1, and asymmetric cooc matrix)", which describes GLCM settings this family
has no concept of. The fixture, the recipe and the reason the two tools' grey levels coincide are
recorded instead; the history lives in the audit report.

**A `#if 0` copy of an older assertion body** in the regression file, removed.

## What was demoted

**Nothing.** All five rows keep `status=vetted`, `oracle=pyradiomics`. Every feature in the family
has an oracle both before and after.

## What was found

### The matrix under the five scalars was never asserted

Coarseness, Contrast, Busyness, Complexity and Strength are all contractions of one
`(i, n_i, p_i, s_i)` table over the grey levels present — eighteen numbers on this fixture. Two
errors in that table can cancel inside any one of the five sums, so five scalar assertions cannot
tell a correct matrix from a compensating pair. PyRadiomics builds `P_ngtdm` *before* any feature
formula runs, so intercepting it costs one attribute read and reimplements nothing:

| `i` | `n_i` | `p_i`              | `s_i`            |
| ----: | ------: | -------------------- | ------------------ |
|     1 |      32 | 0.6666666666666666   | 47.05118411000764  |
|     2 |       6 | 0.125                | 0.8266145619086798 |
|     3 |       2 | 0.041666666666666664 | 2.358288770053476  |
|     4 |       4 | 0.08333333333333333  | 9.10880296174414   |
|     5 |       1 | 0.020833333333333332 | 3.1923076923076925 |
|     6 |       3 | 0.0625               | 12.916289592760181 |

The assertion takes those levels **out of the same featurisation the five scalar assertions run** —
`extract_3d_ngtdm` hands the ROI's voxel cube back beside `fvals`. The one step it reproduces is
`calculate()`'s zero-min correction (`+1` on every level when the minimum is 0), which at
`NGTDM_GREYDEPTH=0` is the whole of its preamble, and the minimum is asserted rather than assumed.

**Two negative controls**, one per property:

| planted                                                                                                                  | result                                                                                                              |
| ------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------- |
| `s_5` golden moved by rel 1e-8 (sized to exceed `rel=1e-9` while level 5's 2%/4% share keeps the features inside it) | matrix test**FAILED** naming grey level 5; all five scalar tests passed; the generator exited 1 independently |
| one voxel of the returned cube perturbed, every golden left alone                                                        | matrix test**FAILED** naming grey level 1 (`48.1176` vs `47.0512`); all five scalar tests passed          |

The second is the one that matters for the shape. An earlier version of this assertion worked from a
48-entry hand-written literal of the phantom's levels, and no change to what the loader produces
could have reached it — SPEC §5.2's fixture-encodes-the-model failure, and the same defect the
reviewer raised on #438 item 2 where a mean was recomputed from a second featurisation of the same
slices. Both perturbations were reverted.

### `agrees_gt(..., 1)` is not a loose band, it is no band

`agrees_gt`'s third argument is a **divisor**: `tolerance = golden / frac`. `1` makes the tolerance
equal to the golden, so `agrees_gt(P[0], 0.375, 1)` passes for anything in `[0, 0.75]`. Eight values
of the `P` and `S` vectors were asserted that way — the vectors the whole family is derived from.

Their goldens were three-significant-figure copies of PyRadiomics' documentation, and one is
visibly rounded: `s_3` was pinned at `3.03` where the exact value is `91/30 = 3.033333333333333`.
An 1.1e-3 relative error, which `rel=1e-9` rejects and `1` never could.

Two details of that hand-worked example, recorded because both tools agree on them and the
docstring may not lead a reader to expect either: grey level 4 is absent from the image and **both
tools drop empty levels**, so the table has four rows and not the docstring's five; and the
docstring's own table and its worked arithmetic disagree on `s_3` (`2.63` vs `3.03`). The run agrees
with the arithmetic.

### PyRadiomics cannot load this fixture through its public API

The mask is label 57 in all 48 voxels with no background, and `imageoperations.getMask()` raises

```
ValueError: No labels found in this mask (i.e. nothing is segmented)!
```

whenever `numpy.unique(mask)` has a single entry. `RadiomicsFeatureExtractor.execute()` goes through
it, and so does the `pyradiomics` CLI — so the invocation the header recorded has never been
runnable against this fixture. The generator constructs `RadiomicsNGTDM` directly, which reaches the
same feature code without the loader check.

Because the oracle had to be driven through a non-public entry point, the generator carries a second
reference: `reference_ngtdm()` builds the NGTDM from the IBSI definition in exact rational
arithmetic with no `radiomics` import in its path, and refuses to print anything if the two disagree
on levels or counts.

| comparison                                                               | worst relative difference |
| ------------------------------------------------------------------------ | ------------------------- |
| PyRadiomics vs the independent reference, 6 levels × (`p_i`, `s_i`) | 2.7e-16                   |
| PyRadiomics vs the independent reference, 5 features                     | 0                         |
| **Nyxus vs PyRadiomics, 5 features**                               | **0**               |

The limitation is recorded in `TOOLS.md` and in the benchmark entry, since it applies to any future
oracle run against a background-free mask.

### The drift guards had never been compiled — and would have compared against NaN

`test_3d_ngtdm_regression.h` held five functions, a five-entry golden table and a `#if 0` block, and
no translation unit included it; its five `TEST()` registrations did not exist either. Its settings
vector set `GREYDEPTH` and nothing NGTDM-specific, so `NGTDM_RADIUS` would have been 0 and all five
assertions would have compared NaN against their pins on first execution.

It now runs at `ngtdm3d.regression_ut_phantom` (`GREYDEPTH=64`, `NGTDM_GREYDEPTH=64`,
`NGTDM_RADIUS=1`) on `bench_ut57_3d`, at `rel=1e-9`, with a `test_3d_ngtdm_dump_regression()` to
regenerate it. Its old pins turn out to have been 3–4% stale and **all five would have passed their
own ±10% band** — the orphan's real defect was its settings, not its numbers. That generalises, and
`not_covered.md` §B.1 now says so.

### Range, identity and cross-table checks

Run mechanically over all 35 pins by the generator, not spot-checked. Positivity; `n_i ≥ 1` (both
tools drop empty levels, so no row may be empty); `s_i ≥ 0`; ascending levels; `p_i == n_i/N_v,p`;
`Σ n_i == N_v,p`; `Σ p_i == 1`; and the cross-table one — **the five feature pins recomputed from
the matrix pins alone**, which is what a copy-paste into one table and not the other cannot survive.
All pass. As on 2D GLDZM, these catch a rotted pin and not a wrong definition: the ±10% and ±100%
bands this PR replaces would have passed every one of them.

### The coverage sweep stays, and here is why

`test_3d_ngtdm_coverage.h` still instantiates both parameterized suites, and its oracle half calls
the same helper the five named tests call — so five of its cases duplicate five named assertions, in
a family that is fully oracle-backed. That is the condition under which four other 3D families have
retired their instantiations (open question 7).

Not retired here, and not only because that question is open:
`Test3DFeature_WITH_3P_EMBEDDED_GT`'s body is three assertions and only the third is the duplicate.
The second — `assert_3d_feature_is_registered_and_computable` — checks the feature is registered in
the FeatureManager and computable through the generic path, which none of the five named tests do,
since they drive the feature class directly. Retiring the sweep without porting that check would lose
it, so this family's retirement is a port rather than a delete, and `test_3d_coverage_common.h` is
shared with every other 3D branch.

## New audit artifacts

| file                                                           | what it is                                                                                                  |
| -------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------- |
| `tests/vetting/oracles/gen_ngtdm3d_pyradiomics.py`           | the oracle run, the independent reference, pin verification, range/identity and cross-table checks          |
| `tests/vetting/audit/scan_ngtdm3d_coverage.py`               | regenerates the coverage CSV from the test sources;`--check` runs the acceptance check                    |
| `tests/vetting/audit/ngtdm_3d_coverage.csv`                  | feature → test map, scanned rather than written                                                            |
| `tests/vetting/audit/ngtdm_3d_pyradiomics_vetting_report.md` | measurements, the NaN defect, the negative control, registry corrections                                    |
| `tests/vetting/audit/ngtdm_3d_golden_regen.md`               | runnable steps for all four tables, and the convention differences to account for                           |
| `tests/vetting/matrix/ngtdm3d.md`                            | the config-point matrix — five cells, one VALID, one degenerate, two unexercised with what each would need |
| `tests/test_3d_ngtdm_common.h`                               | shared fixture, settings recipe and mock workflow                                                           |
| `tests/test_3d_ngtdm_mechanics.h`                            | the guard on the radius default                                                                             |

`config_recipes.md` gains `ngtdm3d.pyradiomics_binwidth1` and `ngtdm3d.regression_ut_phantom`;
`benchmarks.md` gains `bench_compat_ngtdm_3d`; `TOOLS.md` gains the loader gotcha; `not_covered.md`
§B.1 drops from seven orphaned files to four and from 112 dead functions to 66.

## How it was verified

Not "tests pass" — a fresh build, a fresh oracle run, and the generator against the header it feeds.

| gate                                              | result                                                                                                                                                                                                                                                                                                                                                            |
| ------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| gtest, Windows (MSVC, Release, tiff-only)         | **879 pass** on this branch, +9 from its base's 870 — exactly the cases added (2 matrix + 5 regression + 2 dumps + 1 mechanics, less the renamed one). **881 pass / 1 skip on the post-rebase tree**, the extra two being #437's                                                                                                                     |
| ASan+UBSan gtest, Linux (`ubtest1`, gcc 13)     | **881 cases, 880 pass, 1 skip, 0 sanitizer diagnostics**, rc=0 — run against the POST-REBASE tree, not this branch's pre-rebase bytes; the skip is upstream's `TEST_2D_GABOR_GPU_RUNS_MECHANICS` on a CPU-only build. The count matches the Windows run because neither side builds DICOM here, and all six source checksums re-read clean after the run |
| `pytest tests/python/` (whole directory)        | **88 passed, 1 skipped**, 7 `@pytest.mark.arrow` failures from the tiff-only local build                                                                                                                                                                                                                                                                  |
| `gen_ngtdm3d_pyradiomics.py` against the header | 15 pins + 2 matrix tables verified, reverse-clean, range/identity/cross-table clean —`ALL CHECKS PASSED`                                                                                                                                                                                                                                                       |
| `check_test_names.py --check`                   | 0 violations                                                                                                                                                                                                                                                                                                                                                      |
| `check_coverage.py --check`                     | clean                                                                                                                                                                                                                                                                                                                                                             |
| `scan_ngtdm3d_coverage.py --check`              | `checked 5 rows: clean`                                                                                                                                                                                                                                                                                                                                         |

The generator was also run against a deliberately corrupted header and exits 1 naming the level, so
its "ALL CHECKS PASSED" is a result rather than a default.

## Changelog

Commit types on this branch: one `fix`, two `test`, one `chore`, one `docs`. `release-please` shows
`fix` and `docs` and hides `test` and `chore`, and only the `fix` moves the version (a patch,
pre-1.0).

- **Squash-merge** — the PR title becomes the subject, so the changelog gets the `fix` line. That is
  the right one to keep: it is the only user-visible change here.
- **Merge-commit** — the changelog gets the `fix` line and the `docs(vetting)` line; the two `test`
  and the `chore` are hidden by configuration.


tests/test_3d_ngtdm_mechanics.h
```diff
@@ -0,0 +1,32 @@
...
+// and the family is undefined at 0: gather_zones() then visits only the centre voxel, skips it, and
+// no voxel is recorded as having a neighbour, so the matrix stays empty and every feature divides by
+// zero. This asserts that a run which sets no --3ngtdm/radius still reaches the feature with a
+// radius that produces numbers, which is the same guarantee compile_feature_settings() already gives
+// GLCM_OFFSET a few lines up, for the identical reason.
+void test_3d_ngtdm_default_radius_mechanics()
+{
+	Environment e;
+	e.compile_feature_settings();
+	ASSERT_GE (STNGS_NGTDM_RADIUS (e.fsett_D3_NGTDM), 1);
```
vjaganat90: The behavior introduced by this PR is specifically **default radius = 1**, but `ASSERT_GE(..., 1)` allows it to drift silently to 2 or higher. Those radii are finite yet produce different feature values, so the loop below would still pass. Please pin `ASSERT_EQ(..., 1)` so this guard actually protects the new default contract.

tests/vetting/matrix/ngtdm3d.md
```diff
@@ -0,0 +1,61 @@
...
+|---|---|---|---|---|
+| 0 (no binning) | false | 1 | **VALID(pyradiomics)** | `ngtdm3d.pyradiomics_binwidth1` — all five features on `bench_compat_ngtdm_3d`, plus the 18-entry matrix, at `rel=1e-9`; measured residual **0** |
+| 64 (MATLAB) | false | 1 | **VALID-prod-only** → regression | `ngtdm3d.regression_ut_phantom`; five pins in `test_3d_ngtdm_regression.h` on `bench_ut57_3d` at `rel=1e-9` |
+| any | true | ≥1 | **not exercised** | IBSI mode builds the `0..max` ladder, which includes levels with `n_i = 0`; PyRadiomics deletes empty levels, so the two are not directly comparable and no assertion runs it |
+| any | any | **0** | **INVALID — degenerate** | the neighbourhood is empty, no voxel is recorded as having a neighbour, `Nvc = 0` and every feature is `0/0`. This was the default until `compile_feature_settings()` was given `NGTDM_RADIUS = 1`; `test_3d_ngtdm_default_radius_mechanics` now holds that closed |
+| any | any | ≥2 | **VALID in principle, not exercised** | the setting is honoured (`--3ngtdm/radius`), but no reference tool in the matrix exposes a distance other than 1 for NGTDM, so nothing could be compared against |
```
vjaganat90: This says no reference exposes a distance above 1, but lines 59–61 correctly note that PyRadiomics supports `distances=[2]` and that it was already verified. Under SPEC §5.1 that makes radius 2 a real oracle-mappable config point. Please either add the radius-2 oracle assertion/recipe or document a measured incompatibility that prevents comparison; the current two claims cannot both stand.

src/nyx/env_features.cpp
```diff
@@ -729,6 +729,12 @@ void Environment::compile_feature_settings()
 			s[(int)NyxSetting::GLCM_OFFSET].ival = 1;
 			s[(int)NyxSetting::GLCM_GREYDEPTH].ival = get_coarse_gray_depth();
 			s[(int)NyxSetting::GLCM_NUMANG].ival = (int)GLCMFeature::angles.size();
+			// NGTDM-specific default (otherwise zero-initialised). NGTDM_RADIUS=0 empties the
+			// neighbourhood - the scan visits only the voxel of consideration and skips it - so no
+			// voxel is recorded as having a neighbour, the matrix stays empty and every 3D NGTDM
+			// feature divides by zero. Default the Chebyshev neighbourhood radius to 1 (IBSI
+			// delta=1), which is what --3ngtdm/radius overrides.
```
vjaganat90: Small but user-facing terminology fix: `3ngtdm/radius` is exposed through `Nyxus3D.set_metaparam(...)`, not as a `--3ngtdm/radius` CLI option. Please use the actual API spelling here and in the repeated audit/matrix text so readers are not directed to a nonexistent flag.

tests/vetting/oracle_coverage.csv
```diff
@@ -707,11 +707,11 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
... imageoperations.getMask() rejects -- so the generator constructs RadiomicsNGTDM directly and cross-checks it against an independent numpy NGTDM (worst rel 2.7e-16). test_3d_ngtdm_regression.h pins Nyxus' own values on the ut_ phantom at recipe ngtdm3d.regression_ut_phantom and claims no oracle; test_3d_ngtdm_coverage.h is a completeness sweep. gen_ngtdm3d_pyradiomics.py A SECOND CONFIG POINT, ngtdm3d.pyradiomics_binwidth1_r2: the same phantom and binning at NGTDM_RADIUS=2 against PyRadiomics distances=[1, 2], asserted by test_3d_ngtdm_<feature>_r2_pyradiomics and test_3d_ngtdm_matrix_r2_pyradiomics, again at rel=1e-9 with residual 0. The radius is the axis a run gets wrong by leaving it unset, and one config point cannot separate honouring it from ignoring it; at radius 2 every feature moves by a quarter or more. PyRadiomics' distances names Chebyshev shells rather than a radius, so the match for radius 2 is [1, 2] and not [2] -- distances_semantics_check() in the generator measures both readings every run. The phantom is 4x4x3, so radius 3 already spans it from every voxel and supports no further config point.",TEST_NYXUS.TEST_3D_NGTDM_COARSENESS_PYRADIOMICS,bench_compat_ngtdm_3d
```
vjaganat90: SPEC §3 — and this PR’s own `scan_ngtdm3d_coverage.py:11` — requires one registry row per `(feature × config × oracle)` assertion. These five rows still identify only the radius-1 recipe/cases while the new radius-2 assertions are buried in `notes`; please add five `ngtdm3d.pyradiomics_binwidth1_r2` rows naming the corresponding `*_R2_PYRADIOMICS` cases.

tests/vetting/audit/ngtdm_3d_pyradiomics_vetting_report.md
```diff
@@ -0,0 +1,466 @@
...
+
+Per SPEC and revet §6, the headers state current-state facts. Removed: the `#if 0` copy of an older
+regression body, the provenance block naming `ut_inten.nii` as both image and mask, and the sentence
+`(100 grey levels, offset 1, and asymmetric cooc matrix)` — a GLCM description, in the NGTDM file,
+describing settings this family has no concept of. The recipe, the fixture and the two tools' binning
+are recorded instead, and this report carries the rest.
+
+---
+
+## The generic 3D coverage sweep stays
```
vjaganat90: Since #463 is now merged, this section is no longer true: main deleted `test_3d_ngtdm_coverage.h` and the shared `Test3DFeature_WITH_3P_EMBEDDED_GT` harness. Please drop/update this argument on rebase and remove the dead coverage-header references from the NGTDM registry rows; this PR should not resurrect the retired sweep.

tests/vetting/oracle_coverage.csv
```diff
@@ -720,11 +720,16 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
...on the 4x4x3 NGTDM phantom (label 57), recipe ngtdm3d.pyradiomics_binwidth1. Both sides discretise the phantom's levels 0..5 to 1..6 -- PyRadiomics by binWidth=1, Nyxus by the zero-min correction at NGTDM_GREYDEPTH=0 -- so no binning convention separates them and the residual is 0 to the last bit; asserted at rel=1e-9. Backed by the FIVE SCALARS AND THE MATRIX THEY CONTRACT: all five features are sums over one (i, n_i, p_i, s_i) table of six grey levels, so test_3d_ngtdm_matrix_pyradiomics pins that table too, per level, against PyRadiomics' own P_ngtdm array. A planted 1e-8 error in one s_i fails the matrix assertion naming its level while all five scalars still pass. PyRadiomics' public extractor cannot load this phantom at all -- its mask is label 57 in all 48 voxels with no background, which imageoperations.getMask() rejects -- so the generator constructs RadiomicsNGTDM directly and cross-checks it against an independent numpy NGTDM (worst rel 2.7e-16). test_3d_ngtdm_regression.h pins Nyxus' own values on the ut_ phantom at recipe ngtdm3d.regression_ut_phantom and claims no oracle. gen_ngtdm3d_pyradiomics.py",TEST_NYXUS.TEST_3D_NGTDM_COARSENESS_PYRADIOMICS,bench_compat_ngtdm_3d
```
vjaganat90: These radius-1 oracle rows still fold in `test_3d_ngtdm_regression.h`, but that header asserts a different recipe: `NGTDM_GREYDEPTH=64` on the UT phantom, with Nyxus snapshots and no oracle. Please give those five `*_REGRESSION` assertions separate `status=regression`, empty-oracle rows for `ngtdm3d.regression_ut_phantom`, and remove the regression header from these PyRadiomics rows.

tests/test_3d_ngtdm_pyradiomics.h
```diff
@@ -1,208 +1,315 @@
...
+	std::vector<PixIntens> I (U.begin(), U.end());
+	std::sort (I.begin(), I.end());
+	ASSERT_EQ (I.size(), expected.size());
+
+	std::vector<std::pair<PixIntens, double>> Zones;
+	D3_NGTDM_feature::gather_zones (Zones, D, radius, 0/*zeroI: no level is background here*/);
+
+	std::vector<int> N;
+	std::vector<double> P, S;
+	double Nvp = D3_NGTDM_feature::calc_NGTDM (N, P, S, Zones, I);
```
vjaganat90: This verifies a separately reconstructed `N/P/S`: it calls `gather_zones()` and `calc_NGTDM()` after the `calculate()` run, but never observes the matrix state that `calculate()` actually used. Please expose/assert that production state, or narrow the matrix-vetting language to an independent reconstruction using the production helpers.

tests/test_3d_ngtdm_mechanics.h
darkclad: Fixed

tests/vetting/matrix/ngtdm3d.md
darkclad: Fixed

src/nyx/env_features.cpp
darkclad: Fixed

tests/vetting/oracle_coverage.csv

tests/vetting/audit/ngtdm_3d_pyradiomics_vetting_report.md
darkclad: Fixed

tests/vetting/oracle_coverage.csv

tests/test_3d_ngtdm_pyradiomics.h
darkclad: Fixed

vjaganat90:
LGTM!

PR #454, opened 2026-08-25
test(3d glszm): vet the family against PyRadiomics and pin the matrix behind it

darkclad:
## What this is

The 3D GLSZM entry in the family-vetting series. The registry already called all sixteen features `vetted` against PyRadiomics, with **`config_recipe` and `tolerance` blank on every row** â€” the same pair of blanks that hid 3D NGTDM's Â±10% band. This pass runs the oracle the rows name, records what it measures, and tightens what the claim rests on.

**The one `src/` change is 12 lines of read-only accessors, and no behaviour moves.** `3d_glszm.h` gains `get_P()`, `get_I()`, `get_Ng()`, `get_Ns()`, `get_Nz()`, `get_Np()` â€” `const` views of the size-zone matrix, the grey levels its rows are indexed by, and the four dimensions it was allocated from. No member is added, no logic is touched, nothing outside the tests calls them. They exist because the matrix assertion has to read the object `calculate()` filled rather than a copy rebuilt beside it; see *The matrix is pinned* below, which is the review finding that made them necessary.

## What was measured

All sixteen features reproduce PyRadiomics 3.0.1. Fifteen agree to **within a few ULP** â€” the worst relative residual over them is 1.19e-15 (`3GLSZM_GLV`), which is float summation order and nothing else. `3GLSZM_ZE` sits at 1.9e-4, which is Nyxus' `fast_log10` approximation.

So no row changed status and nothing was demoted. The pass is about what surrounded that claim:

|  | before | after |
|---|---|---|
| oracle band | `agrees_gt(..., 10.)` = Â±10% on all 16 | `rel=1e-9` on 15, `rel=1e-3` on `3GLSZM_ZE` |
| pytest band | `rtol=1e-1, atol=1e-2` | `rtol=1e-9, atol=0` (`3GLSZM_ZE` `rtol=1e-3`) |
| the size-zone matrix on the phantom | not asserted at all | 186 cells, `Ng`/`Ns`/`Nz`/`Np`, and a count proving no cell outside them is populated |
| the 4Ã—4Ã—3 matrix check | hand-computed values under a `_pyradiomics` name | the same cells, produced by a PyRadiomics run the generator repeats |
| `test_3d_glszm_regression.h` | 16 functions, 0 registrations, never `#include`d | wired in, re-pinned at `%.17g` |
| the `IBSI=true` branch | executed by no assertion in the tree | vetted, `glszm3d.pyradiomics_ibsi_gapped`, 16/16 plus its matrix |
| the family's settings at a default run | never exercised | `TEST_3D_GLSZM_DEFAULT_GREYDEPTH_MECHANICS` |
| a constant-intensity ROI | undescribed | recorded as a divergence, pinned at `glszm3d.regression_constant_roi` |
| registry rows | 16, `config_recipe` and `tolerance` empty on all of them | **80**, one per assertion (SPEC Â§3), every one carrying a recipe, a band and a benchmark |

`agrees_gt`'s third argument is a **divisor**: `10.` is Â±10%, `1e9` is `rel=1e-9`. The old band accepted anything within a tenth of the golden, on every feature of the family.

The 80 rows are five configurations of sixteen features: two oracle recipes (`glszm3d.pyradiomics_bincount20`, `glszm3d.pyradiomics_ibsi_gapped`) and three drift guards (`glszm3d.regression_ut_phantom`, `â€¦_nobinning`, `â€¦_constant_roi`).

### `3GLSZM_ZE` keeps a looser band, and the reason is named at the site

`calc_sums_of_P()` accumulates the family's only logarithm as `fast_log10(p/sum_p + EPS) / LOG10_2` where PyRadiomics uses `numpy.log2`. Measured residual 1.9e-4; band `rel=1e-3`, SPEC Â§7's documented-residual tier. This is the same fast path 2D GLCM (`rel=5e-3`) and 2D GLDM (`rel=2.5e-3`) already band for â€” a deliberate convention of this codebase, banded rather than reported. It is not a free pass: the other fifteen features contract the same matrix at `rel=1e-9`, so an error in the matrix fails fifteen assertions plus 186 matrix cells.

## The matrix is pinned, not only the sixteen scalars

Every feature of this family is a sum over one `P(i, j)` size-zone matrix, so sixteen scalar assertions cannot separate a correct matrix from two errors inside it that cancel. `TEST_3D_GLSZM_MATRIX_PYRADIOMICS` pins the matrix itself against PyRadiomics' `P_glszm`, the table it builds before any feature formula runs: on this phantom **20 Ã— 634, 186 non-empty cells, 860 zones over 4 800 voxels**, plus a count that no cell outside those 186 is populated.

**It reads the feature object, not a copy.** The first version rebuilt `P` beside `calculate()` from `gather_size_zones()`, so a defect in the row mapping, in `Ng`/`Ns`, in the allocation or in the fill loop left it green â€” review caught this, and it is why the six accessors above exist. Negative control **J** is the proof: making the production fill loop skip zones of size 5, with `gather_size_zones()` untouched, now fails the matrix assertion, and could not have failed the earlier one.

**A PyRadiomics column index is not a zone size.** `_calculateCoefficients()` deletes every column no zone occupies and keeps the surviving sizes in `jvector` â€” 634 columns become 46. The first draft of the generator read the size off the index, and the cross-table check caught it before anything was pinned: recomputing the sixteen definitions from the mislabelled matrix reproduced the nine features that do not weight by `j` and missed the other seven by up to 94%. That gotcha is now in `TOOLS.md`, because it applies to any family whose matrix is intercepted this way.

**The 4Ã—4Ã—3 check now claims the oracle it names.** It asserted hand-computed expectations under a `_pyradiomics` suffix â€” the function name is the oracle claim, so the name credited a tool the assertion had never run. Fixed by making the claim true rather than by renaming: the generator runs `RadiomicsGLSZM` on the same literal at `binWidth=1` and reproduces all eight of its cells. The hand-worked values were correct.

**And the connectivity fixture now measures connectivity.** Its predecessor had one populated slice between two empty ones, so every `dz != 0` neighbour of every voxel was background: 26-, 18- and 2D 8-connectivity all produced the same nine zones there. `bench_cube4x4x3_zcross` has three populated slices, and the four readings give 9, 10, 13 and 13.

## The orphan, and what wiring it in cost

`test_3d_glszm_regression.h` was 16 functions and 16 pins that had never compiled. It built a settings vector that set `GREYDEPTH=64` and **left `GLSZM_GREYDEPTH` at 0**, so it had never executed at the settings it carried. Measured before assuming: at those settings **all sixteen of its assertions would have failed on their first run**, by between 1.6Ã— and 1.5e6Ã—. Its pins reproduce at no configuration in the tree â€” 12 of 16 miss at the settled `=64` and 13 of 16 at the `=100` its own header comment claimed â€” so they are a snapshot of an older state of the code, not of a lost recipe.

Re-pinned from a fresh `%.17g` run at `glszm3d.regression_ut_phantom`: `GREYDEPTH=64`, `IBSI=false`, `GLSZM_GREYDEPTH=64`, the configuration the retired 3D coverage sweep ran every 3D family at on this same phantom.

Unlike 3D NGTDM's radius, `GLSZM_GREYDEPTH=0` is **not** degenerate â€” it is this family's documented "no binning" default and it produces finite values. `TEST_3D_GLSZM_DEFAULT_GREYDEPTH_MECHANICS` asserts exactly that, which is why this PR proposes no behaviour change.

## A constant-intensity ROI is discarded, and should not be

`calculate()` opens with `if (r.aux_min == r.aux_max) { invalidate (STNGS_NAN(s)); return; }` under a comment reading "intercept blank ROIs (equal intensity)". Those are two different conditions. A constant-intensity ROI is fully populated: one grey level, one 26-connected zone, a single populated cell, sixteen finite features over it â€” on a 2Ã—2Ã—2 block, `SAE = 1/64`, `LAE = 64`, `ZE = 0`, `GLV = 0`, `ZP = 1/8`. Nyxus returns the soft-NaN sentinel for all sixteen.

The guard is doing real work at radiomics binning, where `to_grayscale_radiomix` divides by `(max - min)`; it is unconditional, so it discards the two schemes that would have answered too. Negative control **O** makes one voxel of the block different so the intercept does not fire, and the family computes `SAE = 0.51020408163265307` â€” the arithmetic is intact, the guard is too wide.

Narrowing it changes a public feature's output on a reachable input, so it is `src/` work on its own branch. This PR pins the current behaviour instead, with the sentinel set to a distinctive `-98765` so a zero-filled feature buffer cannot satisfy the assertion, and gives it a recipe of its own plus sixteen `status=regression` rows on `bench_cube2_constant`. The day the intercept is narrowed, sixteen named assertions fail and say so.

## The scanner also walks backwards now

`scan_glszm3d_coverage.py --check` started every check from a registry row, so the one thing it could not report was a row that had never been written â€” and review found exactly that: `test_3d_glszm_constant_roi_regression` asserts all sixteen features at a seventeenth configuration, and nothing named it.

- `--check` now also walks tests â†’ registry: a gtest case whose function asserts a feature value must be named by a row. Mechanics and invariant cases are exempt (no such row exists anywhere in the registry) and the `_dump_` regenerators fall out on their own, since they print rather than compare.
- That needed the coverage rule widened. A function which range-loops `D3_GLSZM_feature::featureset` asserts the whole family while naming no feature, so no pattern in the file saw it; it is now credited with all sixteen, which is also what `glszm_3d_coverage.csv` should have been saying.
- **Proven non-vacuous**: delete the sixteen new rows and `--check` names the case and exits 1; restore them and it is clean.

The forward checks tightened in the same review round: `current_test` must be *exactly* the file the row's `test_name` is defined in â€” a superset passed the old one-way rule â€” plus a kind match, one recipe per case, and uniqueness of (feature, recipe, oracle).

## What is in the diff

New:

- `tests/test_3d_glszm_common.h` â€” the phantom declarations, the settings recipe, the three literal fixtures and one mock workflow the oracle, regression and mechanics assertions share. Fixtures only, no reference data (SPEC Â§6.3.1).
- `tests/test_3d_glszm_mechanics.h` â€” the default-settings guard and the `IBSI=true` â‰¡ no-binning equivalence.
- `tests/vetting/oracles/gen_glszm3d_pyradiomics.py` â€” opens the header it feeds, re-verifies every pin in it, and gates its exit code on the bound, identity, cross-table and matrix checks too.
- `tests/vetting/audit/scan_glszm3d_coverage.py` + `glszm_3d_coverage.csv` â€” the featureâ†’test map, read out of the sources; `--check` also runs the family-plan acceptance check, both directions.
- `tests/vetting/audit/glszm_3d_pyradiomics_vetting_report.md`, `tests/vetting/audit/glszm_3d_golden_regen.md`, `tests/vetting/matrix/glszm3d.md`.

Changed: `src/nyx/features/3d_glszm.h` (the six accessors), the two test headers, `tests/test_all.cc`, `tests/python/test_nyxus.py`, and the registry plus `config_recipes.md`, `benchmarks.md`, `TOOLS.md` and `not_covered.md`.

## How it was verified

Re-run on the current head, after the rebase onto `main`:

- **gtest, Windows**: **845 cases, 844 pass, 1 skipped** (`TEST_2D_GABOR_GPU_RUNS_MECHANICS`, CPU-only build). The base measured the same way is **821** â€” the delta is exactly the 24 registrations this branch adds net of one rename.
- **`pytest tests/python/`**: 88 passed, 1 skipped, 7 failed â€” all 7 are `USE_ARROW=OFF` in the local build, unrelated to this branch.
- **The generator against the header**: `32 verified, 0 failed, 0 unproducible, 0 unpinned, 0 out of range or identity-violating, 0 cross-table mismatch(es), 0 matrix mismatch(es)` â€” 186 + 8 + 4 matrix cells reproduce. `ALL CHECKS PASSED`.
- **Checkers**: `check_test_names.py --check` 0 violations; `check_coverage.py --check` exits 0 and `--write` moves nothing; `scan_glszm3d_coverage.py --check` clean at 80 rows and idempotent.
- **Twelve negative controls** stand, each built and run then reverted, all failing as designed â€” six perturbing a **golden** and six perturbing the **input**, including one that changes a single voxel of the extracted cube and one that plants a defect in the production fill loop. Controlling only the golden proves the assertion compares, not that it compares the right thing; both halves are in the report, along with the three the review pass retired because the code they perturbed no longer exists.
- **ASan + UBSan, Linux**: clean on the pre-rebase tree. **Not yet re-run on the current head** â€” the rebase and the two commits after it are test, registry and doc changes plus the six `const` accessors, but the gate should run again before merge.

## Types that reach the changelog

None. Every commit is `test`, `chore` or `docs`, all configured hidden in release-please, so a merge-commit produces no changelog entry and no version bump. The PR title is `test` for the same reason â€” there is no behaviour change to announce, and the `src/` change is `const` accessors no shipping code calls.

## Recorded, not fixed here

`src/` findings from reading the family, filed rather than acted on:

- **The `aux_min == aux_max` intercept discards constant-intensity ROIs** (above). Pinned and recorded as a divergence in `matrix/glszm3d.md`; narrowing it is a behaviour change on a reachable input.
- `env_features.cpp:697` maps `typeid(D3_GLSZM_feature)` a second time where it means `D3_GLRLM_feature`. It is unreachable today, and for a worse reason: the map's only consumer, `phase3.cpp:114`, passes `typeid(f)` on a `FeatureMethod*`, so every oversized-ROI feature already gets `f_settings_[0]` regardless.
- Two dead locals in `calculate()`: a shadowed `std::unordered_set<PixIntens> U` and an unused `int i = 0`.


tests/test_3d_glszm_pyradiomics.h
```diff
@@ -1,289 +1,522 @@
...
+	{ 20, 2, 2 },
+};
 
-    // (6) get values
+// A 4x4x3 volume with one populated slice between two empty ones, small enough that its zones can be
+// read off by eye and every one of them is a 2D 8-connected component of the middle slice. It is the
+// direct check that the 26 offsets gather_size_zones() walks agree with PyRadiomics' neighbourhood:
+// the phantom above would hide a connectivity difference in a matrix nobody can check by hand.
+//
+// The volume is the fixture here, not a copy of one -- there is no file to read it from, and the
```
vjaganat90: This fixture has only one populated z-slice, so 26-connectivity and 2D 8-connectivity produce the same zones; removing every z-neighbor from production would still pass. Put matching voxels on adjacent slices (ideally connected only vertically/diagonally) and regenerate the oracle matrix before calling this a 3D connectivity check.

tests/vetting/matrix/glszm3d.md
```diff
@@ -0,0 +1,80 @@
...
+the voxel count of a connected component, a property of the volume rather than a setting. 3D GLSZM
+therefore has no distance, angle or alpha axis at all, unlike 3D GLCM, 3D GLRLM or 3D NGLDM.
+
+| ibsi | binning | levels | verdict | recipe / oracle |
+|---|---|---|---|---|
+| False | radiomics (`GLSZM_GREYDEPTH < 0`) | 20 bins | VALID | `glszm3d.pyradiomics_bincount20` — pyradiomics, 16/16 |
+| False | MATLAB (`GLSZM_GREYDEPTH > 0`) | 64 | VALID-prod-only | drift guard, `glszm3d.regression_ut_phantom` — no oracle |
+| False | none (`GLSZM_GREYDEPTH == 0`) | the raw levels | REACHABLE, NOT VETTED | the default a run with no `--3glszm/greydepth` gets; asserted finite only, in `test_3d_glszm_mechanics.h` |
+| True | none (forced) | the raw levels | NOT MEASURED | see below |
+| any | any | ROI with `aux_min == aux_max` | INVALID | degenerate — a blank ROI has no zones; `calculate()` returns NaN before building a matrix, guarded in code rather than by an oracle |
```
vjaganat90: `aux_min == aux_max` is a nonempty constant-intensity ROI, not a blank ROI; it has a valid one-zone GLSZM. Please record the current NaN as a reachable production divergence/defect (with a regression guard if retained), not `INVALID`.

tests/test_3d_glszm_pyradiomics.h
vjaganat90: **Blocker:** z=0 and z=2 are entirely background, so for every foreground voxel all 18 neighbours with `dz != 0` are zero and can never join its zone. This matrix is therefore identical under full 26-connectivity and a purely 2D 8-neighbour implementation; add adjacent populated slices with vertical, z-edge, and z-corner-only connections, then regenerate the PyRadiomics pins.

tests/vetting/matrix/glszm3d.md
vjaganat90: **Blocker:** `aux_min == aux_max` is a nonempty constant-intensity ROI, not a blank ROI; it has a valid one-zone GLSZM. Please record the current NaN as a reachable production divergence/defect (with a regression guard if retained), not `INVALID`.

vjaganat90 (changes requested):

tests/test_3d_glszm_pyradiomics.h
```diff
@@ -1,289 +1,522 @@
...
+	std::vector<std::pair<PixIntens, int>> zones;
+	D3_GLSZM_feature::gather_size_zones (zones, binned, 0/*zeroI at radiomics and no binning*/);
+	ASSERT_EQ ((int)zones.size(), expect_nz);
+
+	int ns = 0;
+	for (const auto& z : zones)
+		ns = (std::max) (ns, z.second);
+	ASSERT_EQ (ns, expect_ns);
+
+	SimpleMatrix<int> P;
```
vjaganat90: **Blocker:** This never checks the matrix built by `D3_GLSZM_feature::calculate()`: it calls `gather_size_zones()` and fills a separate local `P`. A defect in the production row mapping, `Ng`/`Ns`, allocation, or fill loop can therefore leave the advertised matrix pin green; expose/assert production `P`, or narrow the claim to independent zone-list reconstruction.

tests/vetting/matrix/glszm3d.md
```diff
@@ -0,0 +1,80 @@
...
+fixture whose zones can be counted by hand (see below). Zone *size* likewise has no parameter: it is
+the voxel count of a connected component, a property of the volume rather than a setting. 3D GLSZM
+therefore has no distance, angle or alpha axis at all, unlike 3D GLCM, 3D GLRLM or 3D NGLDM.
+
+| ibsi | binning | levels | verdict | recipe / oracle |
+|---|---|---|---|---|
+| False | radiomics (`GLSZM_GREYDEPTH < 0`) | 20 bins | VALID | `glszm3d.pyradiomics_bincount20` — pyradiomics, 16/16 |
+| False | MATLAB (`GLSZM_GREYDEPTH > 0`) | 64 | VALID-prod-only | drift guard, `glszm3d.regression_ut_phantom` — no oracle |
+| False | none (`GLSZM_GREYDEPTH == 0`) | the raw levels | REACHABLE, NOT VETTED | the default a run with no `--3glszm/greydepth` gets; asserted finite only, in `test_3d_glszm_mechanics.h` |
+| True | none (forced) | the raw levels | NOT MEASURED | see below |
```
vjaganat90: **Blocker:** Both unguarded rows are reachable production configs. Under SPEC §5.1 the default `GLSZM_GREYDEPTH=0` needs numeric regression pins (finiteness is mechanics only), and the distinct `IBSI=true` fill/indexing branch needs a measured gapped-level oracle or regression; `NOT MEASURED` is not a final matrix disposition.

tests/vetting/oracle_coverage.csv
```diff
@@ -707,22 +707,22 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
...us takes the only logarithm in the family through fast_log10() -- a documented fast path, banded rather than reported. Backed by the SIXTEEN SCALARS AND THE MATRIX THEY CONTRACT: all sixteen are sums over one P(i,j) size-zone matrix, so test_3d_glszm_matrix_pyradiomics pins that matrix too -- all 186 non-empty cells of the 20x634 table, its Ng/Ns/Nz/Np, and a count proving no cell outside them is populated -- against PyRadiomics' own P_glszm array, read through its jvector because PyRadiomics deletes the columns for zone sizes nothing occupies. It reads the cube the same extraction produced, so perturbing one voxel of it fails the matrix assertion; a planted cell-count error fails it naming the level and size, while all sixteen scalars still pass. test_3d_glszm_smallmatrix_pyradiomics runs the same assertion on a 4x4x3 volume whose nine zones can be counted by eye, which is the direct check of the 26-neighbourhood. test_3d_glszm_regression.h pins Nyxus' own values on the ut_ phantom at recipe glszm3d.regression_ut_phantom and claims no oracle; test_3d_glszm_coverage.h is a completeness sweep. gen_glszm3d_pyradiomics.py",TEST_NYXUS.TEST_3D_GLSZM_SAE_PYRADIOMICS,bench_compat_liver_3d
```
vjaganat90: **Blocker:** These rows conflate two assertions: their recipe and `test_name` identify the `-20` PyRadiomics oracle, while `current_test` also claims the separate `+64` regression. SPEC §3 requires one row per assertion; add 16 `status=regression`, empty-oracle rows for `glszm3d.regression_ut_phantom` with the corresponding `*_REGRESSION` test names.

tests/vetting/audit/scan_glszm3d_coverage.py
```diff
@@ -0,0 +1,283 @@
...
+            out.append(f"{f}: current_test names {stale}, which covers nothing for it")
+
+        # The row describes one assertion, so its two identifiers must agree: the file named in
+        # current_test is the file the case named in test_name is defined in.
+        name = r.get("test_name", "")
+        if not name:
+            out.append(f"{f}: no test_name, so current_test names an assertion nothing identifies")
+        elif name not in cases:
+            out.append(f"{f}: test_name {name} resolves to no registered case in test_all.cc")
+        elif cases[name] not in claimed:
```
vjaganat90: **Should fix:** This check is only one-way: it passes when `current_test` contains the correct oracle file plus unrelated regression/sweep files, which is exactly the current conflation. Since this scanner says it enforces one assertion per row, require the claimed assertion file to match the `test_name` file exactly (or validate row kind/config explicitly) so `--check` catches this.

tests/test_3d_glszm_pyradiomics.h
darkclad: Fixed

tests/vetting/matrix/glszm3d.md
darkclad: Fixed

tests/test_3d_glszm_pyradiomics.h
darkclad: Fixed

tests/vetting/matrix/glszm3d.md
darkclad: Fixed

tests/vetting/oracle_coverage.csv

tests/vetting/audit/scan_glszm3d_coverage.py
darkclad: Fixed

tests/vetting/oracle_coverage.csv
vjaganat90: **Blocker:** The +64 rows are now correctly split, but `test_3d_glszm_constant_roi_regression` adds 16 assertions for `glszm3d.regression_constant_roi` and the registry has no rows for that config. Please add one regression row per feature naming `TEST_NYXUS.TEST_3D_GLSZM_CONSTANT_ROI_REGRESSION`; the scanner currently cannot catch the omission because it only validates rows that already exist.

vjaganat90 (approved):
LGTM!

PR #456, opened 2026-08-26
test(3d gldm): vet the family against PyRadiomics and pin the dependence matrix behind it

darkclad:
Continues the family-vetting series. Rebased onto `upstream/main` at `834e59da`.

**The `src/` change is eleven lines**, all in `src/nyx/features/3d_gldm.h`: five read-only accessors
(`get_P`, `get_I`, `get_Ng`, `get_Nd`, `get_Nz`) so an assertion can read the dependence matrix
`calculate()` built instead of a copy rebuilt beside it. `3d_gldm.cpp` is untouched — `git diff
upstream/main -- src/nyx/features/3d_gldm.cpp` is empty, and the ASan gate checksums both files to
say so. This is the shape `D3_GLSZM_feature` already carries on `main` since `a1090c0d`. No public
feature value moves.

## The family's values were already right; the evidence around them was not

A fresh PyRadiomics 3.0.1 run at the recipe the header's provenance block claims reproduces **all
fourteen pinned goldens at `rel = 0`**, so those goldens were genuinely that tool's output rather
than a snapshot wearing an oracle name. Nothing is demoted here. What did not hold up was the band,
the orphaned snapshot file, two defects inside it, and two production configurations nothing ran.

| | before | after |
|---|---|---|
| oracle band | `agrees_gt(..., 10.)` = ±10% on all 14 | `ASSERT_NEAR` at **`abs=1e-9`** on 13, `abs=4e-3` on `3GLDM_DE` |
| the dependence matrix | not asserted at all | the table the run built: 163 cells, `Ng`/`Nd`/`Nz`, the allocation, the row mapping, + a 4×4×3 neighbourhood check |
| `test_3d_gldm_regression.h` | 14 functions, 0 registrations, never `#include`d | wired in, re-pinned at `%.17g`, one wrong assertion fixed |
| the compiled default (`GLDM_GREYDEPTH=0`) | reachable in production, pinned nowhere | 14 snapshot assertions of its own |
| a nonempty constant-intensity ROI | undescribed | measured against PyRadiomics, dispositioned as a defect, guarded |
| pytest band | `rtol=1e-1, atol=1e-2` (vacuous on one feature) | `rtol=0, atol=1e-9`, matching the C++ |
| registry | 14 rows, `config_recipe` and `tolerance` blank on every one | 56 rows, one per assertion, recipes and bands filled |
| generator, matrix doc, audit artifacts | none existed | all present |

## SPEC §7's exact tier, and why it is honest here

Measured Nyxus against PyRadiomics at the recipe: across the thirteen features that take no
logarithm the worst residual is **7.11e-15 absolute** (`3GLDM_SDHGLE`) and **8.79e-16 relative**,
eight of them exactly 0. The band is an **absolute** `1e-9` asserted with `ASSERT_NEAR` — SPEC §7 row
one verbatim, not a relative check wearing the same number.

The tier applies for the reason the SPEC gives it, no estimator disagreement, and all three things
that could have separated the implementations are asserted rather than assumed:

- `GLDM_GREYDEPTH=-20` and PyRadiomics' `binCount=20` produce the same twenty levels;
- both sides walk the full 26-voxel neighbourhood at Chebyshev distance 1 with the cutoff `alpha=0`
  — PyRadiomics through its `distances=[1]` / `gldm_a=0` defaults, Nyxus through a fixed `shifts[]`
  table and `pi == neig_pi`, so neither exposes a knob the other lacks;
- both start the dependence count at 1, so a voxel with no dependent neighbour lands in the same
  column and there is no index shift to correct for.

**`3GLDM_DE` takes `abs=4e-3`.** It is the family's only sum over logarithms and Nyxus takes it
through `fast_log10()`; the measured residual is **1.7512e-3**, and the band is that doubled and
rounded up to one significant figure. That is this family's own measurement — the 2D twin's identical
`calc_DE()` lands at `1.3e-3` *relative*, roughly five times wider, so carrying that number across
would have produced a band far looser than anything measured here. A documented fast path, banded
with the cause named at the site.

## The matrix under the scalars, read off the run that produced it

All fourteen features are contractions of one `P(i, j)` dependence matrix, so a compensating pair of
errors inside it survives every scalar assertion.

`TEST_3D_GLDM_MATRIX_PYRADIOMICS` asserts the table `D3_GLDM_feature::calculate()` left behind, taken
off the same feature object the fourteen scalar assertions run through. It pins `Ng`, `Nd` and `Nz`
as `calculate()` reports them, `P`'s width and height **as it allocated them**, each of 163 cells at
the row `I` maps its grey level to, and then the non-empty count, the matrix's own zone total and the
deepest occupied column read off `P` rather than off the pins. `Np=4800` is counted **off the cube
rather than out of the matrix**, which is what makes `Nz == Np` evidence that no voxel outside the
ROI enters the matrix instead of a tautology.

`Nd` is pinned twice on purpose: the largest dependence carrying a zone, and separately what
`calculate()` leaves in `Nd`, which is that same number only when the binning is not IBSI — at
`GLDM_GREYDEPTH=0` the `if (greyInfo) Nd = max_Nd` trim is skipped and `Nd` stays at the full
neighbourhood width. Pinning one would have hidden the trim.

Beside it, and separated by `EXPECT_NO_FATAL_FAILURE` so neither arm suppresses the other's verdict,
the same pinned table is held to a 26-offset walk written out in the test file, sharing no code with
the feature class — because a pin regenerated from production is not an oracle.

`TEST_3D_GLDM_SMALLMATRIX_PYRADIOMICS` runs the same pair on a 4×4×3 volume whose two populated
slices are **identical**, forcing every dependence to `2·(in-slice matches) + 2`: a neighbourhood
missing the z offsets would halve them, one missing the two strictly-vertical offsets would make them
odd. It runs `calculate()` on that volume rather than tallying a hand-built cube, so the same four
places are under it.

`gen_gldm3d_pyradiomics.py` re-verifies every pin it feeds, carries the reverse check, and
**rebuilds the matrix from the definition independently of PyRadiomics' C extension** — the one check
that can catch the tool and the written definition disagreeing, which no cross-table check can see.
All 163 cells agree three ways: the tool's array, the definition written out, and the fourteen
published scalars recomputed from the pinned cells.

## Two defects in a file that had never executed

`test_3d_gldm_regression.h` was reached by no `#include`, so none of its fourteen functions had ever
run.

1. **`test_3d_gldm_lgle_regression()` asserted `Feature3D::GLDM_SDE`** under the `_lgle_` name, so
   `3GLDM_LGLE` had no drift guard at all and `3GLDM_SDE` had two.
2. **Its pin was `0.26` — byte-identical to the `3GLDM_SDE` pin beside it** — against a measured
   `0.00073572128237550161`. **A factor of 353.** Both halves of one paste.

The whole table is regenerated at `%.17g` rather than carried forward, because it could not be
reproduced at any configuration. At `GLDM_GREYDEPTH=64` the **grey-level axis** reproduces to four or
five digits while every quantity touching the **dependence axis** does not (`SDE` 1.69×, `LDE` 0.86×,
`DV` 0.93×, and all four cross-terms). That split identifies which axis moved, and says the old table
came from a run whose dependence distribution differed from today's rather than from rounding.

`scan_gldm3d_coverage.py --check` now catches this class directly: it diffs each table's keys against
the functions that read them, those against the `TEST()` registrations, and each registry row's
`test_name` against the assertion it names. Run against the pre-fix file it reports the unread key,
the name/argument mismatch and the duplicated assertion separately.

## Two production configurations nothing ran

**`GLDM_GREYDEPTH=0` is the compiled default, not one option among several.**
`Environment::compile_feature_settings()` names a default for `GLCM_OFFSET`, `GLCM_GREYDEPTH`,
`GLCM_NUMANG` and the `FPIMG_*` group, and for nothing else, so the zero-fill stands and only
`--3gldm/greydepth=N` moves it. Every run that passes no such option computes this family there, and
`IBSI=true` reaches the same path deliberately. It is a genuinely different quantity from the `64`
snapshot rather than a rounding of it — on the same ROI `3GLDM_HGLE` is 4,275,550 against 1957.19 —
so all fourteen are snapshotted at it (`gldm3d.regression_ut_phantom_nobinning`). No oracle claim:
whether PyRadiomics can be made to skip discretisation on a non-integer intensity volume is still
open, and `matrix/gldm3d.md` records what would answer it.

**A nonempty constant-intensity ROI is a measured defect, guarded rather than fixed.**
`calculate()`'s opening `if (r.aux_min == r.aux_max)` says "intercept blank ROIs" but tests the
intensity extrema, so it also takes a fully populated ROI of one intensity — reachable wherever a
segmentation lands on a flat or saturated region. PyRadiomics 3.0.1 computes a full GLDM on such an
ROI (`P_gldm` is 1×4 = `[8, 20, 16, 4]` on a 4×4×3 cube); Nyxus returns the no-value sentinel for all
fourteen. Bypassing the intercept as a negative control, **the arithmetic underneath reproduces
PyRadiomics to the last bit on all seven dependence-axis features**, so the intercept is the whole
defect. It is not fixed here: at radiomics binning the same input divides by a zero bin width, so
narrowing the intercept alone would trade a wrong answer for undefined behaviour, and the identical
intercept sits in five sibling families. `test_3d_gldm_constant_roi_regression()` pins what production
emits today, under a `--noval` sentinel no GLDM feature can take, so the day it is fixed is visible.

## The pytest band was vacuous on one feature

Ten of the fourteen Python assertions ran at `rtol=1e-1, atol=1e-2` against the same goldens. On
`3GLDM_SDLGLE` — golden `0.00244`, band `±0.0102` — the test **accepted zero**. It also widened
`3GLDM_DNN` to ±18% and `3GLDM_SDE` to ±16%. All fourteen now use the C++ band in the same form,
`rtol=0` with `atol` carrying the absolute tier. Because this exercises the same recipe through the
Python API, the two paths agreeing at `1e-9` also says the API reaches the same code.

## Registry: 14 rows become 56

Each assertion gets its own row — one file in `current_test`, an exact `test_name` resolved against
`test_all.cc`, a benchmark id, and `source=in-tree`. Fourteen vetted rows on
`gldm3d.pyradiomics_bincount20`, and fourteen `status=regression` rows on each of
`gldm3d.regression_ut_phantom`, `gldm3d.regression_ut_phantom_nobinning` and
`gldm3d.regression_constant_roi`. The three snapshot configs are **different quantities**, not second
opinions: `GLDM_GREYDEPTH`'s *sign* selects the scheme (positive MATLAB level-count, negative
PyRadiomics bin count, zero IBSI no-binning). This is the row shape #436 established.

`scan_gldm3d_coverage.py` resolves each row's `test_name` through `test_all.cc` to the function it
runs and requires that function to carry an assertion of the row's own feature, at its kind, in a
file its `current_test` names, at its `config_recipe`. `check_coverage.py::validate_test_names` asks
only whether a name is *a* gtest case, so a row could name an unrelated existing one and stay green;
that is how these rows came to name the small-matrix case, and it is what the resolver closes.

`matrix/gldm3d.md` records that the family has **one and a half axes, not two** — `IBSI=true` does
not cross with `GLDM_GREYDEPTH`, it forces it to 0, so it collapses onto an existing row — and gives
each cell a SPEC §5.1 disposition: the `binCount` cell VALID(pyradiomics), the two snapshot cells
VALID-BUT-PRODUCTION-ONLY, and the constant-ROI cell a defect with the measurement behind it.

## Verification

| gate | result |
|---|---|
| gtest, Windows | **879 cases, 878 pass, 1 skip** (`TEST_2D_GABOR_GPU_RUNS_MECHANICS`, GPU off) |
| `--gtest_filter=*3D_GLDM*` | **48/48 pass** — 14 oracle + 2 matrix + 14 + 14 + 1 regression + 3 dumps |
| ASan+UBSan, Linux | **879 cases, 878 pass, 1 skip, 0 sanitizer diagnostics**, gtest rc=0; own tree and log, `hostname` printed, source md5s of all six files verified before the build and again after the run |
| `pytest tests/python/` | **88 passed, 1 skipped**; the 7 `@pytest.mark.arrow` failures are the tiff-only local build |
| generator against its header | `ALL CHECKS PASSED`, exit 0 — 14 scalars, 163 + 9 cells, 7 constants, 12 bounds, 8 identities, 0 unpinned |
| the checkers | `check_coverage.py --check`, `check_test_names.py --check`, `scan_gldm3d_coverage.py --check` all clean; the scanner is byte-identical on re-run |

**Eighteen negative controls, all failing as designed**, tabled in
`tests/vetting/audit/gldm_3d_pyradiomics_vetting_report.md`. Three perturb the **input** rather than a
golden, because a perturbed golden proves an assertion compares while only a perturbed input proves it
compares the run under test. One tightens `3GLDM_DE` to the exact tier and fails, proving that band is
load-bearing rather than decorative. Four perturb `src/` to measure what the matrix assertion sees
that nothing else does, and two of those are the point:

| perturbation | fails | everything else |
|---|---|---|
| `P.allocate(Nd + 1, Ng + 1)` → `Nd + 3` | both matrix cases | **46 of 48 GLDM cases stay green** |
| the `if (greyInfo) Nd = max_Nd` trim disabled | the phantom matrix case | **47 of 48 stay green** |

Neither changes a single feature value — `calc_*()` reads rows `1..Ng` and columns `1..Nd`, so extra
allocated columns are never summed and an untrimmed `Nd` only widens a loop over cells that are zero.
Both were invisible to every one of the 48 assertions until the matrix was read off the run.

## One thing this does not close, stated rather than claimed

`calculate()` indexes a zone's row as `inten - 1` under `IBSI` and as the level's position in `I`
otherwise. **Both matrix fixtures carry contiguous grey levels** — the phantom's `binCount` binning
gives 1..20, the small volume's raw levels are 1..4 — so on either one the two expressions are the
same number, and swapping them leaves both matrix cases green. What catches it is the MATLAB-binning
snapshot, whose binned levels are gapped: all fourteen `_REGRESSION` assertions fail. So the branch is
under assertion in this suite, just not by the assertion that reads `P`.

Closing it there needs a gapped-level fixture with a `binWidth=1` oracle — the shape
`glszm3d.pyradiomics_ibsi_gapped` uses next door, and the same measurement `matrix/gldm3d.md` lists as
what would move the no-binning cell to VALID. That is a new oracle recipe and a new config point
rather than a fix to this assertion, so it is recorded as the named limitation in the vetting report.
Happy to land it here instead if you would rather it not be a follow-up.


tests/test_3d_gldm_pyradiomics.h
```diff
@@ -1,187 +1,458 @@
...
+	{ 3, 4, 4 },
+	{ 3, 6, 2 },
+	{ 4, 2, 4 },
+	{ 4, 6, 6 }
+};
 
-void test_3d_gldm_gln_pyradiomics() { 
-    assert_3d_gldm_feature_pyradiomics(Nyxus::Feature3D::GLDM_GLN, "3GLDM_GLN"); 
-}
+// The 26 offsets calculate() walks, in {dz, dy, dx}. Spelled out here rather than reached for in the
```
vjaganat90, line 285: Spelling out the shifts here makes this an independent definition check, but it never observes the matrix built by `D3_GLDM_feature::calculate()`—`got` is constructed entirely in this helper. A production change to `shifts`, `nd`, or the fill logic can leave this green, so either expose/assert production `P` or narrow the test/PR/docs claim.

tests/vetting/matrix/gldm3d.md
```diff
@@ -0,0 +1,90 @@
...
+different quantity, not a rounding of the others — on `bench_ut57_3d` the same ROI gives:
+
+| feature | `GLDM_GREYDEPTH=64` | `GLDM_GREYDEPTH=0` |
+|---|---|---|
+| `3GLDM_HGLE` | 1957.1946 | 4275550.79 |
+| `3GLDM_GLV` | 153.0947 | 341996.30 |
+| `3GLDM_SDE` | 0.15361 | 0.86996 |
+| `3GLDM_DE` | 8.40347 | 11.48909 |
+
+Nothing in the tree asserts it. It is recorded here rather than closed because pinning it would be a
```
vjaganat90: `GLDM_GREYDEPTH=0` is the compiled production default, so SPEC §5.1 makes this VALID-BUT-PRODUCTION-ONLY → regression, not “reachable, unmeasured.” The +64 snapshot cannot guard it—the values above differ by orders of magnitude—so please add a default/no-binning regression or a mapped oracle.

tests/vetting/audit/scan_gldm3d_coverage.py
```diff
@@ -0,0 +1,325 @@
...
+                out.append(f"{fn}() passes {feat} but its name says 3GLDM_{want}")
+            if fn not in registered:
+                out.append(f"{fn}() reads a pin but no TEST() in test_all.cc calls it")
+        seen = list(readers.values())
+        for dup in sorted({v for v in seen if seen.count(v) > 1}):
+            out.append(f"{table}: {dup} is asserted by more than one function")
+    return out
+
+
+def disagreements(rows, asserted, oracles, regression, where):
```
vjaganat90, line 328: `disagreements()` never examines `test_name`; `registered_calls()` only verifies table-reader functions are registered. A row can therefore name an unrelated existing GTest and still pass both checks. Please resolve each row’s `test_name` to the matching feature/kind/file, or narrow the PR’s exact-mapping claim.

tests/vetting/matrix/gldm3d.md
```diff
@@ -0,0 +1,93 @@
...
+There is no GPU axis either: `D3_GLDM_feature::calculate()` has no `STNGS_USEGPU` branch.
+
+## Config points
+
+| `GLDM_GREYDEPTH` | `IBSI` | benchmark | verdict | recipe / oracle |
+|---|---|---|---|---|
+| `-20` (binCount 20) | false | `bench_compat_liver_3d` | **VALID(pyradiomics)** | `gldm3d.pyradiomics_bincount20`; all 14 features asserted in `test_3d_gldm_pyradiomics.h` at `abs=1e-9`, `3GLDM_DE` at `abs=4e-3` |
+| `64` (MATLAB levels) | false | `bench_ut57_3d` | **VALID-prod-only** → regression | `gldm3d.regression_ut_phantom`; all 14 pinned in `test_3d_gldm_regression.h` at `rel=1e-9`; no oracle claim |
+| `0` | false | — | **reachable, unmeasured** | see below |
+| any | true | — | **reachable, unmeasured** | forced to `greyInfo = 0`, i.e. the same code path as the row above, reached a different way |
```
vjaganat90: **Blocker:** This matrix omits nonempty constant-intensity ROIs. `D3_GLDM_feature::calculate()` returns NaN immediately when `aux_min == aux_max`, but such an ROI has a valid one-grey-level dependence matrix that PyRadiomics can compute; add the cell with an oracle/regression defect guard and registry disposition rather than leaving this reachable production case invisible.

tests/vetting/oracle_coverage.csv
```diff
@@ -656,20 +656,34 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
...ck that the neighbourhood includes the z offsets. The generator rebuilds the matrix from the definition independently of PyRadiomics' C extension and requires all 163 cells to agree. test_3d_gldm_regression.h pins Nyxus' own values on the ut phantom at recipe gldm3d.regression_ut_phantom and claims no oracle. Evidence: tests/vetting/audit/gldm_3d_pyradiomics_vetting_report.md, generator tests/vetting/oracles/gen_gldm3d_pyradiomics.py. 3GLDM_DE is the exception at abs=4e-3: it is the family's only sum over logarithms and Nyxus takes it through fast_log10(), a float-precision approximation divided by a ten-digit LOG10_2, where PyRadiomics uses numpy.log2. Measured residual 1.7512e-3 absolute / 2.65e-4 relative (Nyxus 6.6031219389041276 against PyRadiomics 6.6048731874541904); the band is that doubled and rounded up to one significant figure. A documented fast path, banded rather than reported. Note the 2D twin's identical calc_DE() measures 1.3e-3 RELATIVE, five times wider, so this figure is this family's own and not carried across.",TEST_NYXUS.TEST_3D_GLDM_DE_PYRADIOMICS;TEST_NYXUS.TEST_3D_GLDM_MATRIX_PYRADIOMICS;TEST_NYXUS.TEST_3D_GLDM_SMALLMATRIX_PYRADIOMICS,bench_compat_liver_3d
```
vjaganat90: **Blocker:** These per-feature rows claim `TEST_3D_GLDM_SMALLMATRIX_PYRADIOMICS`, but that case asserts only matrix cells on a different literal/no-binning (`binWidth=1`) fixture; it is not a feature assertion on `gldm3d.pyradiomics_bincount20` / `bench_compat_liver_3d`. Keep each row tied to its matching scalar case, and record the matrix corroboration separately rather than conflating fixtures, configs, and assertion kinds in `test_name`.

tests/test_3d_gldm_pyradiomics.h
darkclad: Fixed

tests/vetting/matrix/gldm3d.md
darkclad: Fixed

tests/vetting/audit/scan_gldm3d_coverage.py
darkclad: Fixed

tests/vetting/matrix/gldm3d.md
darkclad: Fixed

tests/vetting/oracle_coverage.csv

tests/test_3d_gldm_pyradiomics.h
vjaganat90: **Still unresolved:** `tally_3d_gldm_matrix_production()` still calls `gather_dependence_zones()` and builds its own map, while `extract_3d_gldm()` discards the feature object. This exercises production traversal, but it still does not inspect the `P`/row-mapping/allocation/fill state produced by `D3_GLDM_feature::calculate()`.

darkclad: Fixed

PR #460, opened 2026-08-27
test(3d gldm): retire the coverage sweep, whose cases duplicate the n…

darkclad:
> **Merge after #456.** Not a preference — see "Why the order matters" below; the ordering is what
> keeps a measured hole closed, and merging this first opens it for as long as the two are apart.

Continues the per-family retirement of the 3D coverage sweep — #441 (glcm), #446 (morphology),
#448 (ngldm), #452 (glrlm), #455 (gldzm), all merged. This is the sixth, and the first on a family
with **no baseline table**.

## What changes

Three paths, no `src/` file among them:

- `tests/test_3d_gldm_coverage.h` deleted (21 lines: two `INSTANTIATE_TEST_SUITE_P` calls, no
  `register_coverage_baseline()`)
- its `#include` drops out of `test_all.cc`
- the deleted filename drops from `current_test` on the family's 14 registry rows, as eb89d06 did
  for glrlm's 32

**No `src/` change at all**, so nothing here can affect the CLI or `import nyxus`.

## Nothing had to be ported, and it cost zero `add_keys()` lines

The previous five retirements each cost one line in `individually_pinned_3d_feature_names()`. This
one costs none, and that is a property of the family rather than an oversight:

All 14 public `3GLDM_*` features are keys of `gldm_3d_pyradiomics_ref_vals`, so
`externally_vetted_3d_feature_names()` contains all 14 and every one classifies into the
`WITH_3P_EMBEDDED_GT` half. `GLDM_UNVETTED_LOCAL_REGRESSION` instantiated **zero** cases — confirmed
empirically, gtest does not even list the suite. The SPEC-1 completeness loop in
`TEST_3D_FEATURE_COVERAGE_COUNTS` runs over `feature_3d_cases(false)` and therefore never sees a
gldm feature, so the family needs no entry in the pinned set and takes no dependency on
`gldm_3d_regression_ref_vals`.

## The 14 cases it did instantiate were duplicates

`assert_oracle_backed_agreement()` dispatches gldm to
`assert_3d_gldm_feature_pyradiomics(c.code, c.name)` — the identical helper the 14 named
`test_3d_gldm_*_pyradiomics` tests call with the same arguments, all 14 registered in `test_all.cc`.
Same shape #452 found for glrlm. **No oracle coverage is removed.**

For the record on a stacked tree: #456 rewrites that helper from `agrees_gt(..., 10.)` = ±10% to
`ASSERT_NEAR` at an absolute 1e-9. So after #456 these are 14 cases that had just been tightened by
eight orders and still passed, at a measured worst residual of 7.11e-15.

## Why the order matters

This is the limitation open question 7 already states — *"deleting a named test while leaving its
table entry passes the C++ guard; only a per-family scanner catches that, and four of the nine
families still do not have one."* gldm is one of the four. What follows is that cost measured rather
than argued.

**The sweep was a real guard, not only a duplicate.** `feature_3d_cases_for_family("gldm", true)`
enumerates from `D3_GLDM_feature::featureset` filtered by `externally_vetted_3d_feature_names()` —
independent of which named tests exist or are registered. So it asserted all 14 features whether or
not their named tests did.

**Negative control on this branch:** delete `test_3d_gldm_de_pyradiomics()` and its `TEST()`, rebuild.

| check | result |
|---|---|
| gtest | 881 tests, 880 pass, **exit 0** |
| `check_coverage.py --check` | **exit 0** |
| `check_test_names.py --check` | **0 violations** |
| 3D GLDM cases registered | **13, not 14** |
| anything asserting `3GLDM_DE` | **nothing** — the three remaining matches are 2D |
| what the registry still claims | `status=vetted, oracle=pyradiomics` |

A feature loses its only assertion and the entire gate set stays green. The registry cannot catch it
because all 14 rows have `test_name`, `benchmark`, `tolerance` and `config_recipe` empty on `main`,
and `validate_test_names()` only inspects rows where `test_name` is non-empty.

**#456 closes it, and that is verified too.** Its 28 rows carry `test_name` on **all 28**, and it
adds `scan_gldm3d_coverage.py`. Either alone is sufficient: populating `test_name` on the one broken
row above and re-running gives

```
ERROR: 3GLDM_DE: test_name 'TEST_NYXUS.TEST_3D_GLDM_DE_PYRADIOMICS' is not a gtest case in
test_all.cc - name it SUITE.CASE as declared there            (exit 1)
```

**A second reason the order is cheap:** #456's registry rewrite already contains **zero** mentions of
`test_3d_gldm_coverage.h`, so this PR's 14-row hunk becomes a no-op on rebase and the change reduces
to the file deletion plus one include line.

## What is removed, stated rather than implied

`assert_3d_feature_is_registered_and_computable`. Its name/code half (`find_3D_FeatureByString` plus
an assert on the returned code) is done by the named tests, and its one-provider-per-code half by
`FeatureManager::check_11_correspondence()` via `test_feature_manager_mechanics.h`. Its last two
assertions — that the feature's slot in the **global** value table is populated and finite, i.e. that
`save_value()` writes it — survive nowhere, here as in every retirement in this series.

Its practical reach was always narrower than it sounds: `make_3d_coverage_settings()` hand-sets every
grey depth and `NGTDM_RADIUS`, so the check only ever covered the calculators under a correctly
configured `Fsettings`, never the defaults `compile_feature_settings()` produces — which is why it
did not catch the 3D NGTDM all-NaN bug that #453 fixes.

## Verification

- **gtest, Windows/MSVC:** 882 tests, 881 pass, 1 pre-existing GPU skip. **896 → 882 = exactly 14**,
  both ends measured by building `main`'s own files, not inferred. Build exit code checked via an
  explicit end marker before either count was believed.
- **ASan/UBSan, Linux:** 881 pass, 1 skip, `WRAPPER_EXIT=0`, **0 diagnostics**. Per-branch tree,
  `tests/` and `src/` both synced and `touch`ed. All four changed paths' md5s pinned in the log
  header and verified equal to the local tree; `CMAKE_HOME_DIRECTORY` asserted to point at that tree;
  `CMAKE_BUILD_TYPE` empty so `-O1 -g1` really applied; `libasan.so.8` confirmed linked. Linux 882 ==
  Windows 882, compared against this branch's own Windows count.
- **`pytest tests/python/`, whole directory:** 88 passed, 1 skipped, 7 failed — all 7 the
  `@pytest.mark.arrow` tests on a tiff-only build.
- **`check_test_names.py --check`** 0 violations; **`check_coverage.py --check`** clean.
  `scan_gldm3d_coverage.py` does not exist on this base — N/A here, and it arrives with #456.
- **Free-text grep, not just the typed columns** (#441 review items 2 and 3): `grep -r` for the
  deleted filename across `tests/`, including the unvalidated `notes` column — **0 hits**.

## Known and deliberate

- **The retirement section for `gldm_3d_golden_regen.md` is not in this PR.** That file exists only
  on #456's branch; appending would mean importing unmerged work or forking a file another branch
  owns. It goes in at the rebase. No placeholder, per #441 review item 4 — a forward reference to a
  doc that does not exist yet is the dangling-name hazard this series has already been burned by.
- **Conflicts with #450**, which now modifies `test_3d_gldm_coverage.h` (comment rewrite, and it
  drops the `_UNVETTED_LOCAL_REGRESSION` instantiation — behaviourally a no-op, since that half
  instantiates zero cases). Modify/delete; resolution is to keep the delete.
- **The completeness guard's failure message** is assembled as
  `<family>_3d_regression_coverage_ref_vals` and so names a deleted table for any retired family.
  Pre-existing, now shipped for five families, untouched here — it wants one cross-family follow-up,
  not a sixth partial edit to a contended file.
- **No registry row changes status.** All 14 stay `vetted` / `pyradiomics`; only the `current_test`
  column loses the deleted filename.


vjaganat90 (commented):
This is obsolete due to [463](https://github.com/PolusAI/nyxus/pull/463).  [456](https://github.com/PolusAI/nyxus/pull/456) needs to carefully merge now.

PR #464, opened 2026-08-31
fix(intensity): report first-order features in the slide's own intensity domain

darkclad:
# Report first-order intensity features in the slide's own intensity domain

Closes the CT/Hounsfield defect reported against `0.11.0.post1.dev76`: a CT NIfTI read **with no
flag set** came back in an arbitrary offset domain rather than in Hounsfield units.

89 files, +3,598 / −755. Ten commits: the defect, six fixes answering five rounds of review, one
`test(intensity)` commit covering the fourth round's fixes, and two empty `ci:` commits that
re-triggered the wheel builds after transient registry and artifact-service outages. The fixes are
all `fix(intensity)` and the rest carry types release-please leaves out of the CHANGELOG, so it takes
one entry whether this is squashed or merge-committed. Several of the review findings are regressions the earlier commits introduced —
they are called out by name under *What review caught* rather than folded in silently.

## The defect

`NiftiLoader::unhounsfield` read a minimum off the buffer it was loading and shifted the whole
volume by `-min` whenever that minimum was negative — which every CT's is. Nothing recorded the
shift and nothing undid it, so `MEAN`, `MIN`, `MAX`, `MEDIAN`, the percentiles and `ENERGY` were
displaced by an image-dependent amount. `--preserve-hu` shifted by a *different* base (the
within-mask minimum), so the two settings disagreed with each other as well as with the truth, and
only the intensity-histogram family inverted anything.

Shift-invariant features (`RANGE`, `VARIANCE`, `STDDEV`) were correct in both modes, which is what
made it easy to miss: the whole feature set was *self*-consistent, just displaced.

Measured on the reporter's own image and mask, with no flag:

| feature | before | after | truth (PyRadiomics / SimpleITK) |
|---|---:|---:|---|
| `3MEAN` | 2371.901625 | **-607.098375** | -607.098375 |
| `3MIN` | 1966 | **-1013** | -1013 |
| `3MAX` | 3810 | **831** | 831 |
| `3MEDIAN` | 2320 | **-659** | -659 |
| `3RANGE` | 1844 | 1844 | 1844 |
| `3VARIANCE_BIASED` | 30067.203284 | 30067.203284 | 30067.203284 |
| `3ENERGY` | 3.314212e+13 | **2.335867e+12** | 2.335867e+12 |

`preserve_hu=True` now produces the same numbers instead of differing by 1966.

## The fix

The load-time map is **recorded per slide** and **undone on the way out**.

- `SlideProps` carries `inten_scale` / `inten_offset` / `inten_map` / `inten_top_grey`, filled by
  `Nyxus::record_intensity_domain_map` at the end of `scan_slide_props`. Four maps: the offset map
  (shift by the floored **all-pixel** minimum, only when the slide holds a negative value), the
  min-max quantization of a real-valued slide into `[0, --fpimgdr]`, the stored map (a DICOM or
  integer NIfTI slide keeps its stored integers, and the header rescale joins the inverse), and
  native.
- Both intensity families apply the inverse in `report_in_source_domain`: affine for the location
  statistics, the scale alone for the dispersions, nothing for the standardized moments and the
  histogram shape, and a recompute from the mapped parts for `COV`, `QCOD` and `UNIFORMITY_PIU`.
  `ENERGY` is reconstructed from the two sums already on hand, so there is no second pass over the
  pixel cloud. The identity map returns early, which leaves every ordinary non-negative integer
  image byte-identical.
- No loader reads a minimum of its own any more — `ImageLoader::open` hands each one the recorded
  offset, and derives `quantize` / `dr` / `fpmin` / `fpmax` once above the format branch so no
  backend can drift from another. The offset is driven by the **all-pixel** minimum rather than the
  within-mask one: offsetting by a within-mask minimum would clamp every off-mask pixel below it and
  move the ROI's own minimum onto grey level 0, which the grey binning reads as background.
- `SlideProps::to_grey_level` is the forward map, mirroring each loader branch for branch — hard
  clamp and rescale on the quantized branch, clamp below only on the offset branch, grey level 0 for
  a non-finite sample. The whole-slide workflows set the vROI's grey range through it, so the ROI
  range the intensity and every texture family bin against is the one the loader actually stored.

## Also in scope, because they are the same defect from other loaders

- Signed-integer TIFF no longer clamps negatives to 0 in either the tile or the strip loader; both
  take the offset map. The strip loader had no real-valued handling at all and now shares the same
  map as the tile loader.
- NIfTI `scl_slope`/`scl_inter` and DICOM `RescaleSlope`/`RescaleIntercept` reach the reported
  values unconditionally rather than only under `--preserve-hu`, so a scanner writing a non-identity
  rescale is no longer wrong by default. Without the flag the rescale is carried in the recorded
  inverse over the stored integers (the stored map); under it the loader applies the rescale and
  takes the offset map. Mask loaders are built with the rescale explicitly off, since their pixels
  are labels.
- A real-valued slide's min-max quantization is inverted too, so its features come back in its own
  float range instead of in quantization steps — what the IH family already did.
- **OME-Zarr was wrong independently of the CT defect.** Both readers instantiated a `uint32_t`
  destination and narrowed every sample into it, so a signed store wrapped its negatives and a
  real-valued one lost its fraction — and the prescan measured the slide's extrema on those same
  converted values, so no offset derived from them could have undone it. Zarr now behaves like every
  other backend: `RawOmezarrLoader` holds its tile as `double` and returns the sample as the file
  states it, `NyxusOmeZarrLoader` applies the same offset-or-quantization map `grayscale_tiff.h`
  does, and the format-keyed early return is gone. An unsigned store is byte-identical to before,
  and a mask is opened with offset 0 and no quantization.
- `COVERED_IMAGE_INTENSITY_RANGE` scales its ROI range, closing the quantized-vs-source half of the
  domain mismatch. The truncation half stays and is documented in `not_covered.md` §F.

## `--preserve-hu` changes meaning

It now selects the offset map for a **real-valued** slide, and for a DICOM or integer NIfTI slide,
which otherwise takes the stored map. Other integer slides take the offset map regardless. The
rescale reaches the reported values either way, so the flag is no longer what stands between a CT
and its Hounsfield units. `README.md`,
`docs/source/supported_formats.rst`, `docs/source/cmdline_and_examples.rst` and the three `nyxus.py`
docstrings say so.

**The name was kept deliberately**, and it is now a poor one — a CT user reading `--preserve-hu`
would reasonably conclude they must pass it, and they must not. Renaming was considered and
declined for this PR; every doc surface therefore leads with the explicit negative ("NOT needed for
CT — Hounsfield units are preserved and reported either way, despite the name") rather than merely
describing the correct behaviour and leaving the inference to the reader. The flag has never
appeared in a tagged release (last tag `0.11.0`, Feb 2026; the flag landed on `main` in July 2026),
so a rename remains available later at no cost to any released API.

## Values that move

Every first-order intensity feature on a slide that holds a negative pixel or is real-valued, and
every one on a signed or real-valued OME-Zarr store. This is not a silent bugfix from a user's point
of view — the numbers change, and they change to the correct ones.

Two smaller sets move for reasons of their own:

- **An ROI sitting entirely at the slide's minimum.** The blank-ROI guard read
  `aux_min == 0 && aux_max == 0`, and the offset map puts the slide's all-pixel minimum on grey
  level 0 — so an all-air CT ROI at −1013 HU could not be told from an ROI with no pixels and was
  turned away before the histogram ran. `UNIFORMITY` came back 0 rather than the 1 a single-valued
  ROI has, `ENTROPY` and the percentiles were defaults rather than computations, and the per-bin
  `HISTOGRAM` was empty, while the affine inverse made the location statistics look plausible. Such
  an ROI is now described. `QCOD` and `UNIFORMITY_PIU` are 0 rather than NaN on it.
- **A real-valued slide holding a non-finite sample.** One `+Inf` used to take the whole slide with
  it — see *What review caught*.
- **A DICOM or integer NIfTI slide whose rescale is not a slope of 1 with an integer intercept.** On
  the default path its grey levels are now its stored integers, so a fractional rescaled value is
  reported exactly, and a slope below 1 keeps every stored level. A slope-1 integer-intercept CT —
  every CT fixture in the tree, and the reporter's pair — stores the same grey levels as before.

**Every 3D texture golden is unchanged**, and the reason is worth stating precisely, because the
offset map *does* now shift the segmented phantom by 1024. The texture families bin relative to the
ROI's own range — `binW = (max − min) / binCount`, `y = (x − min) / binW + 1` — so a uniform shift
cancels exactly and is invisible to GLCM, GLRLM, GLDM, GLDZM, GLSZM and NGTDM. First-order is not
shift-invariant and did see it: `3ROBUST_MEAN` moved by exactly 1024, and the 29 MATLAB first-order
goldens by exactly −1024 (below).

What would move them is a change to the *narrowing*, which is not uniform. `ut_inten.nii` is float32
with 274,416 of its 1,000,000 voxels fractional, so rounding instead of truncating perturbs a
quarter of the volume by ±1 and nothing cancels it. The narrowing of a real-valued volume like this
one is therefore left exactly as it was: truncating by default, rounding under `--preserve-hu`,
which is the split that has always been in place. See *What review caught* for why that split is now explicit rather than incidental.

## Goldens re-derived

The 3D first-order MATLAB family was pinned in the loader's shifted domain, so all 29 values moved.
Re-derived on GNU Octave 11.3.0 + statistics over the same built-ins the MATLAB generator names,
**with the previous MATLAB values reproduced to 13 significant digits on the old domain first as a
control**. Every location statistic moved by exactly -1024, every dispersion statistic is unchanged,
and `3COV` / `3QCOD` / `3UNIFORMITY_PIU` / `3ROOT_MEAN_SQUARED` were recomputed from their moved
parts. The generator and the audit report carry the new domain.

The comparison changed with them. A relative tolerance is meaningless once a percentile can sit near
zero — `3P01` is 13 on a 2000-wide range, where 2.4 grey levels of binning is 18% of the value and
0.12% of the range — so the binned percentiles are now measured against the voxel range at `2e-3`
(worst residual `1.19e-3`), a zero reference against the range at `1e-3`, and everything else keeps
its relative tier.

## What review caught

Three rounds, one commit each. Two of the seven findings are regressions the first commit
introduced; the rest is either a disagreement between the recorded map and the load-time conversion
that the first commit did not reach, or robustness the map made reachable and nothing had ever had.

### Round 1 — `a1cfdd56`, the recorded map did not match what the loaders store

- **OME-Zarr recorded as `native`**, on the grounds that its tile loader copies voxels untouched. It
  does not. Fixed by making Zarr a first-class backend rather than by narrowing the claim — see
  *Also in scope* above.
- **A constant real-valued slide fell past the quantization branch**, which needs `fpmax > fpmin`,
  and kept the default native map — but the loader has no native path. It read `quantize` off
  `inten_map`, got `false`, and stored `trunc(x - 0)`: a constant 0.5 slide was stored as grey level
  0 and reported as 0, and a constant negative slide lost its sign the same way. The branch now
  records the offset map with the **unfloored** minimum, so the value itself is the offset.
- **The blank-ROI guard could not tell an all-minimum ROI from an empty one** — see *Values that
  move*. All four copies (2D and 3D, in-RAM and out-of-core) now test the pixel cloud for emptiness.
- **`TrivialHistogram::initialize` needed the zero-width range to be safe before that was
  reachable.** The percentile histogram guarded it with an `isnan` check on the bin index; the
  custom-bin loop did not, and `to_grayscale()` divides by the range and casts the resulting NaN to
  a bin index — undefined, and only landing on 0 by accident on the toolchains in use.

### Round 2 — `13460faf`, mirror the loaders' upper clamp, and survive degenerate input

- **`to_grey_level` clamped only below** — the loaders' mirror on the offset branch and not on the
  quantized one, where both tile loaders hard-clamp at each end. `--fpimgmin` / `--fpimgmax` may
  name a window narrower than the slide's own range, so a slide holding more than the window asks
  for mapped past the top grey level the loader can store, and both whole-slide workflows read that
  straight into the vROI's `aux_max` — which sets the histogram and grey-binning range for the
  intensity family *and every texture family*. Silent numeric drift on whole-slide float TIFF and
  OME-Zarr. **A regression from the first commit**, which replaced a correct
  `fp_phys_pivoxels ? target_dyn_range : max` with the unclamped forward map. The scan now records
  `inten_top_grey` and the forward map mirrors each loader branch for branch.
- **`QCOD` and `UNIFORMITY_PIU` divide by sums** — `p75 + p25`, `aux_max + aux_min` — that a
  populated ROI of one grey level at 0 leaves at 0. The old blank-ROI guard returned before either
  was computed, so letting such an ROI through exposed what the guard had also been hiding. Each
  quotient now says what a zero denominator means instead of producing a NaN, at all eight sites:
  2D and 3D, in-RAM and out-of-core, and the two source-domain recomputes.
- **`RawOmezarrLoader::get_uint32_pixel` cast straight from the buffer**, which the previous commit
  had widened from `uint32_t` to `double`. Converting a negative double to an unsigned integer is
  undefined, where the integer narrowing it replaced was merely implementation-defined. **A
  regression from `a1cfdd56`.**
- **A single non-finite sample destroyed the whole slide.** The scan folded every sample into the
  extrema, so one `+Inf` or `-Inf` gave the quantized map an infinite span; every finite pixel then
  mapped to NaN, whose unsigned conversion is undefined, and the ROI came back all zeros — `MIN`,
  `MAX` and `MEAN` of 0 on a slide whose values ran 0.5 to 10.5, with nothing in the output saying
  so. This one predates the first commit. Both scans now measure the extrema over the finite samples
  only, and every load-time map stores a non-finite sample as grey level 0.

### Round 3 — `cb7b70d2`, settle a slide that measured nothing, and close the last unguarded cast

- **A slide with no finite sample in it left both extrema on their sentinels**, which is the only way
  the maximum can end up below the minimum. Left there they reached the recorded offset, and the
  intensity families add that offset back: `MIN`, `MAX` and `MEAN` came out as `DBL_MAX` —
  1.797e308, which is *finite*, so the output sanitizer passed it straight through and the
  `isfinite` assertion written for exactly this case passed on it. They also reached
  `to_grey_level()`, where under `--preserve-hu` the offset does not cancel and the conversion of
  `DBL_MAX` to an unsigned grey level is undefined, and
  `COVERED_IMAGE_INTENSITY_RANGE`'s divisor. `Nyxus::record_scanned_intensity_range` settles such a
  slide to a flat zero range at all three scan exits, which leaves the map at the identity and
  settles all three consumers. The all-pixel minimum is settled separately, since it is measured off
  the whole buffer and survives an empty mask on its own.
- **The unsigned Zarr accessor still could not take a NaN or an over-range sample.** Round 2 added
  the negative clamp; `NaN < 0.0` is false, so the conversion stayed undefined for the one class a
  real-valued store is most likely to hold. It now zeroes a non-finite sample and saturates the
  finite over-range end, so the guarantee belongs to the accessor rather than to an assertion about
  who calls it.
- **`report_in_source_domain` claimed the mapped and identity-map paths agree.** They do not, and
  should not: an ROI of one intensity is degenerate in grey levels and perfectly ordinary in the
  slide's own domain, where `UNIFORMITY_PIU` reads 100 — the honest answer for a uniform ROI. The
  guards exist to stop a NaN, not to make two domains produce one number, and the comment now says
  that.

### Round 4 — `ce35a99c`, `86c1e2de` and `e6b24066`: restore the HU rounding, finish the guards, share one narrowing

- **The `--preserve-hu` narrowing had lost its rounding.** At the merge-base every `std::llround`
  sat inside a `preserve_hu` branch with a plain truncating cast in the `else`:
  `raw_nifti.h::rescale_offset`, the `nyxus_dicom_loader.h` per-pixel loop,
  `grayscale_tiff.h::hu_offset` behind `map_real_intensity`, `SlideProps::uint_friendly_inten`, and
  `workflow_2d_whole.cpp`'s `aux_min`/`aux_max`. Unifying the two branches into one offset map
  carried the *quantized* branch's truncating cast onto it and dropped the rounding with it. That
  matters where the flag is set: `preserve_hu` exists to carry absolute intensities, `inten_scale`
  is 1 there so the inverse cannot recover a dropped fraction, and a fractional sample is exactly
  what a non-integer DICOM `RescaleSlope` or NIfTI `scl_slope` produces — a PET or SUV series. Each
  loader now takes a `round_offset` flag, `ImageLoader::open` passes `p.preserve_hu`, and
  `to_grey_level()` reads it directly so the forward map narrows the way the loaders narrow.
  **Restored only where it was:** rounding the default path as well moves 198 3D cases, for the
  reason given under *Values that move*, and that is a decision about every 3D value in the tree
  rather than a defect to fix in passing.
- **The DICOM and NIfTI maps never got the non-finite guard.** Round 2's commit message enumerates
  "the two TIFF maps, the Zarr map and `to_grey_level`" — the two medical-format maps it also owns
  were missed. A float NIfTI holding NaN, which is how a masked-out background is usually written,
  reached an undefined conversion. Both now store grey level 0, as every other map does.
- **`val_COV` divided by a mean that a zero-valued ROI leaves at zero.** Four of the six sites are
  older than this PR; the two in `report_in_source_domain` are added by it, so this PR introduced
  two new unguarded quotients of exactly the kind it guards two lines below. All six now take the
  same zero-denominator convention as `QCOD` and `UNIFORMITY_PIU`. `e6b24066` adds the test the
  guard lacked, in 2D and 3D, and it was checked to discriminate: with the guards reverted to a bare
  division both zero-valued cases fail on "COV is NaN before the output sanitizer".
- **`86c1e2de` — the narrowing was written out eleven times, and only one copy clamped above.**
  The arithmetic that turns a shifted or rescaled double into a grey level lived once per loader,
  and the copies had drifted: the upper clamp reached only `RawOmezarrLoader::get_uint32_pixel`,
  whose comment called it "this accessor's alone". Clamping only the low end is not a saturating
  cast — converting a double above the destination's maximum is undefined, the same as a negative
  or a NaN — and a 64-bit volume, a real-valued volume under `--preserve-hu`, a large rescale slope
  or a `--fpimgdr` above `UINT32_MAX` reaches it. `Nyxus::grey_level_rounded()` /
  `grey_level_truncated()` in `grey_level_cast.h` now carry it once, taking a non-finite value to 0,
  a negative one to 0, and saturating one above the maximum. `Nyxus::grey_level(y, round)` picks
  between them with the flag `ImageLoader::open` records, which is what keeps it a pure refactor —
  a single rounding narrowing is exactly what moves the 198 cases above. All eleven sites convert,
  including `to_grey_level()`, so the forward map and the loaders share one call. One guard stays
  outside the helper deliberately: the quantized maps check finiteness *before* clamping their
  input to `[fpmin, fpmax]`, since the clamp would otherwise turn `+Inf` into the top grey level.

### Round 5 — keep a header-rescaled slide's stored integers

- **Rescaling before narrowing collapsed a slide with a small slope.** Round 4 left DICOM and integer
  NIfTI rescaled to physical units at load and narrowed to one grey level per unit, with a scale-1
  inverse. A PET series with `RescaleSlope` 0.0005, or a Philips MR volume with `scl_slope` 0.0015,
  stores values up to about 30000 that landed on roughly 0..15 — texture and histogram features
  collapsed, and an ROI near physical 0 fell onto grey level 0, which the binning reads as
  background. **A regression from the first commit**: before this PR the default path stored the raw
  integers. Neither setting could avoid it, since DICOM and NIfTI are never real-valued and the
  quantized map was not on offer.
  Such a slide read without `--preserve-hu` now takes a fourth map, `IntenMap::stored`: the loader
  keeps `u = stored − shift` with the rescale off, so nothing is narrowed, and the recorder carries
  the rescale in the inverse, `intensity = (intercept + slope·shift) + slope·u`. The shift follows
  the offset map's rule in stored units, so a slope-1 integer-intercept CT stores exactly the grey
  levels it did — **no golden moved**. The scan learns the rescale from
  `RawFormatLoader::get_integer_rescale()`; a real-valued NIfTI, a zero, negative or non-finite
  slope, and `--preserve-hu` all keep the offset map.

One question stays open, on its own branch, and is named in the review thread:

- **Whether a real-valued volume's default path should round at all.** A DICOM or integer NIfTI no
  longer narrows by default — Round 5 keeps its stored integers — but a real-valued NIfTI still takes
  the offset map and truncates, which is what it has always done, and nothing in the tree actually
  vets that: the 196 regression cases that cover it are
  self-pins, `gen_firstorder3d_matlab.m` reproduces the truncation by construction
  (`fix(intensity - offset) + offset`), and the PyRadiomics oracles run on `compat_int_mri.nii`,
  int16 with `scl_slope=1`, where rounding and truncation agree exactly. So the default narrowing is
  incumbent and covered, not vetted — and settling it needs an oracle that can see it.

## Tests

40 new gtest cases (934 in the suite) and three new Python mechanics files.

- `test_3d_nifti_offset_map_mechanics.py` — the two properties of the load-time maps the CT fixtures
  cannot show. `ct3d_frac.nii` (8×8×8 int16, `scl_slope=0.5`) makes half its voxels fractional,
  which `ct3d_int16.nii` cannot: at `scl_slope=2` every rescaled value is integral. **Both maps are
  pinned on the same fixture** — MAX −768 / MEAN −896 on the offset map under the flag, where the
  narrowing rounds, and MAX −768.5 / MEAN −896.25 exactly on the stored map without it — so neither
  case can pass vacuously. `ct3d_nan.nii` (float32, NaN / +Inf / −Inf planted) pins the
  non-finite convention either side of the flag, and states plainly what it can and cannot prove:
  `cvttsd2si` happens to yield 0 for a NaN on x86-64, the same grey level the guard stores
  deliberately, so the assertion pins intent and the UBSan run is what discriminates.

- `test_hu_analytic.h` — 24 cases pinning the recorder's four branches, the forward map on each
  branch (including that the offset branch keeps mapping past the slide maximum rather than
  saturating with the quantized one), the range the scan records, non-finite input, and the
  `--fpimg*` rejections. Those rejections are **pinned, not added** — `parse_input` already refused
  an inverted, zero-width, unparseable or non-positive-range window; nothing asserted it. Three of
  the nineteen cover the shared narrowing, since no fixture reaches its upper clamp: the helper at
  its edges, saturation on the offset map with and without `--preserve-hu`, and saturation on the
  quantized map at `--fpimgdr` 1e10 — which also fails if the finiteness check is moved behind that
  map's clamp. Under UBSan with `-fno-sanitize-recover=undefined` the saturation cases can only pass
  if the conversion is actually clamped, so the sanitizer run is what proves the clamp. Five of the
  twenty-four cover the stored map: a PET-like slope of 0.0005 keeping all 30001 levels (beside the
  same slide under `--preserve-hu`, where adjacent levels merge), a slope-1 CT matching the offset
  map's grey levels and inverse with and without a negative value, signed stored values at slope 2
  and a fractional slope of 0.5, the non-negative shift — including `0.3/0.1`, which computes as
  2.9999999999999996 and must not floor to 2 — and the fallbacks to the offset map.
- `test_2d_omezarr_mechanics.h` — 14 cases. `signed.ome.zarr` (32×32 int16, `[-1013, -951]`),
  `float.ome.zarr` (32×32 float32, `[-10.5, 5.0]`) and `nonfinite.ome.zarr` (32×32 float32, row 0
  opening with `NaN`, `+Inf`, `-Inf`, `5e9`, `-7.5`, `3.75`) join the fixtures — uncompressed, a few
  kB each, written directly rather than through bfio so they carry no codec dependency; the recipe is
  in `tests/data/omezarr/README.md`. The cases assert the prescan reads the source value, the tile
  loader stores the mapped grey level, and the unsigned accessor handles every sample class it
  cannot convert, with unsigned negative controls throughout.
- `test_2d_intensity_degenerate_roi_mechanics.h` — the three ratios, `QCOD`, `UNIFORMITY_PIU` and
  `COV`, on a populated ROI of one grey level, read off the feature object **before**
  `force_finite_number`. That is deliberate: the output sanitizer substitutes `ResultOptions::noval()`
  for a NaN, and at the default `noval` of 0.0 no end-to-end test can tell a guarded quotient from an
  unguarded one. Also covers the empty-ROI half of the rewritten guard, which had no test at all.
- `test_3d_intensity_degenerate_roi_mechanics.h` — its 3D twin, the same three ratios. The ROI is
  built by hand rather than loaded, since the 3D grey-level pass reads only `raw_pixels_3D`,
  `aux_min`, `aux_max`, `aux_area` and `slide_idx`. Both files' `COV` assertions were confirmed to
  fail with the guard reverted, while their non-degenerate controls still pass.
- `test_2d_intensity_constant_roi_mechanics.py` — an ROI of one intensity end to end, on a mapped
  slide and on an identity-map slide (the one case nothing recomputes afterwards), through the
  in-RAM and the out-of-core paths.
- `test_2d_nonfinite_pixels_mechanics.py` — `NaN`, `+Inf` and `-Inf` through a float TIFF with a
  clean-slide control, plus an all-non-finite slide segmented and whole-slide, asserting the reported
  values rather than merely their finiteness.
- `test_2d_hu_ct_small_pydicom.py` now asserts absolute Hounsfield units directly against pydicom
  rather than reconstructing them in the test, plus a case for the invariant this defect would have
  failed: for a CT read in Hounsfield units, `MIN` is the ROI's true Hounsfield minimum.
- `test_2d_signed_int16_loader_mechanics.py` asserts the negatives survive rather than that they
  clamp to 0.
- `test_2d_hu_regression.py` / `test_3d_hu_nifti_regression.py` re-pinned to absolute values, and
  each gained a case asserting the flag is no longer what makes the difference.
- `test_3d_nifti_mechanics.h` supplies the offset the way `ImageLoader::open` does, and asserts what
  happens without one.
- `test_2d_hu_mechanics.h` gains a fractional DICOM `RescaleSlope`. Every other DICOM fixture has an
  integral slope, where rounding and truncation store the same grey level, so the DICOM loop's
  narrowing was invisible — yet a PET or SUV series routinely carries a fractional one.
  `ct_frac.dcm` (16×16 uint16, stored `idx`, slope 0.5, intercept −1024) makes half its pixels
  fractional. Two loader cases pin the rescale-then-shift narrowing both ways against the same
  pixels: idx 1, 3 and 255 store 1, 2 and 128 rounding and 0, 1 and 127 truncating, with idx 2 as
  the control where they agree. `TEST_2D_HU_DICOM_FRACTIONAL_SLOPE_LOAD_PATH_MECHANICS` runs the slide
  through the scan, the recorder and `ImageLoader::open`, and reads all 256 pixels back through the
  recorded inverse: exactly `0.5·idx − 1024` on the stored map, rounded to a whole unit under
  `--preserve-hu`. It was derived from `ct_u16.dcm` with pydicom, changing only the slope, the
  intercept, the pixel data and the instance UID.

## Gates

All run against the Round 5 commit unless noted. The vetting-gate row and the COV mutation check
were run against `e6b24066`; Round 5 touches none of the files they read.

| gate | result |
|---|---|
| gtest — Windows, MSVC Release, `USE_Z5=ON`, DICOM on | **933 passed / 1 skipped** (GPU), 0 failed — `e6b24066`'s 927 plus the six cases Round 5 adds, so no golden moved |
| `pytest tests/python/` | **121 passed / 1 skipped**, 0 failed |
| ASan + UBSan — Linux (ubtest1), `-fno-sanitize-recover=undefined`, gtest | **909 passed / 1 skipped**, **0 sanitizer diagnostics**, rc 0 |
| ASan + UBSan — Linux (ubtest1), Python backend, the HU / NIfTI / intensity-domain pytest files | **26 passed**, **0 sanitizer diagnostics**, rc 0 — the stored-map NIfTI case included |
| `check_coverage.py --check` | clean |
| `main`'s vetting gates, on this branch merged with current `main` (`e6b24066`) | `scanlib_selftest`, `run_scanners` (20 of 20 families) and `report_features --check` (932 rows) all clean; the merge is conflict-free |
| reporter's own CT pair | all 7 features match PyRadiomics/SimpleITK to ≤ 1.4e-7 relative, both flag settings; after Round 5 all 52 reported columns are byte-identical to the Round 4 run (int16, `scl_slope` 1, `scl_inter` −1024) |
| mutation check on the `COV` guard (`e6b24066`) | with the grey-level guards reverted, both zero-valued degenerate-ROI cases fail and both non-degenerate controls pass |

The sanitizer build is `NOEXTRAS=ON` / `USE_Z5=OFF`, which is why it runs 24 fewer cases than
Windows: the OME-Zarr, DICOM and Arrow cases are compiled out. So it covers the changes that only a
sanitizer can validate on NIfTI — the non-finite guard, since `cvttsd2si` happens to yield 0 for a
NaN on x86-64, the shared narrowing's upper clamp, and the stored map's NIfTI load path in the Python
run — but not the Zarr or DICOM call sites, which are covered by the Windows gtest build instead.

`run_scanners.py` and `report_features.py` arrived on `main` after this branch was cut, so they
are checked against the merged tree rather than the branch alone.

The one relative error above 1e-10 on the CT pair is `3ENERGY`, and it is the reference, not the
result: the truth value in the table is quoted to seven significant figures.

## Not in scope

Confirmed on the same data but untouched here: `3VOXEL_VOLUME` still reports a voxel **count**
rather than a volume (it ignores voxel spacing), and `3MESH_VOLUME` is still byte-identical to
`3VOLUME_CONVEXHULL`. Both are tracked separately.

The montage prescan does not measure a slide range at all, so `COVERED_IMAGE_INTENSITY_RANGE` is
meaningless on numpy input — before this PR it divided by `-inf`, now by zero, and the output
sanitizer flattens both to the same value. Pre-existing, and the montage-vs-directory test already
skips that feature for the same reason.


src/nyx/slideprops.cpp
```diff
@@ -497,4 +516,57 @@ namespace Nyxus
...
+
+		// A slide handed over in memory (montage / numpy input) never went through a tile
+		// loader, so nothing mapped it.
+		if (p.fname_int.empty())
+			return;
+
+		std::string ext = Nyxus::get_big_extension (p.fname_int);
+
+		// OME-Zarr is carried natively; it has no offset or rescale branch to invert.
+		if (ext == ".zarr" || ext == ".ome.zarr")
```
vjaganat90: **[Blocker] Record the map OME-Zarr actually applies.** This early return is only correct for unsigned-integer Zarr. `ImageLoader` instantiates `NyxusOmeZarrLoader<uint32_t>`, and its `loadTile<FileType>()` copies signed and floating buffers into that `uint32_t` destination; `RawOmezarrLoader` likewise explicitly `static_cast<uint32_t>`s every sample during prescan. Negative values therefore wrap and fractional values truncate, while this branch records the identity inverse and reports those converted grey levels as though they were native source values. Please either make prescan and tile loading preserve/map signed and floating Zarr consistently and record that map here, or reject/narrow those dtypes. A signed or float OME-Zarr fixture asserting source-domain `MIN` and `MEAN` would catch this; the current uint16 fixtures cannot.

src/nyx/slideprops.cpp
```diff
@@ -497,4 +516,57 @@ namespace Nyxus
...
+				fpmax = fpo.max_intensity();
+			}
+			double dr = fpo.target_dyn_range();
+			if (dr > 0.0 && fpmax > fpmin)
+			{
+				p.inten_map = IntenMap::quantized;
+				p.inten_scale = (fpmax - fpmin) / dr;
+				p.inten_offset = fpmin;
+			}
+			return;
```
vjaganat90: **[Blocker] Preserve the source value for a zero-range floating slide.** When `fpmax == fpmin`, this branch returns with the default native map. The TIFF loader still runs the non-quantized path and casts the real sample into `uint32_t`, so a constant `0.5` slide is stored as `0`; the identity inverse then reports `MIN`/`MEAN == 0`, not `0.5`. This is an exact hole in the PR's source-domain contract. Please represent a constant map explicitly (or otherwise carry the constant source value through) and add an end-to-end constant float TIFF test, including a fractional value.

vjaganat90 (changes requested):
Nice direction overall—the per-slide inverse map is a much cleaner model, and the CT/NIfTI regression coverage is strong. I found two cases where the recorded map does not match what the loader actually stores, so the central source-domain guarantee still fails for supported inputs. I think these need to be fixed before merge. Ignoring rebase/conflict state here as requested.

src/nyx/features/intensity.cpp
```diff
@@ -55,6 +55,74 @@ PixelIntensityFeatures::PixelIntensityFeatures() : FeatureMethod("PixelIntensity
...
+
+	// Ratios: same definitions as above, recomputed from the mapped parts. None of them is
+	// invariant, and QCOD and PIU are not even meaningful once intensities can be negative --
+	// which is a property of the measure, not something the domain should hide.
+	val_COV = val_STANDARD_DEVIATION / val_MEAN;
+	val_QCOD = (val_P75 - val_P25) / (val_P75 + val_P25);
+	val_UNIFORMITY_PIU = (1.0 - (val_MAX - val_MIN) / (val_MAX + val_MIN)) * 100.0;
+}
+
+void PixelIntensityFeatures::calculate_grey_levels (LR& r, const Fsettings & fsett, const Dataset & ds)
```
vjaganat90: **[Blocker] Do not treat a populated ROI at the mapped source minimum as blank.** The new offset map deliberately maps the slide's all-pixel minimum to grey level `0`, so a valid constant ROI at that value (for example, an all-air CT ROI at `-1024 HU`) reaches this method with `aux_min == aux_max == 0`. The existing early return below then fires. The inverse map makes some location fields look plausible afterward, but `UNIFORMITY` remains its default `0` instead of `1`, `HISTOGRAM` remains empty, and the general constant-ROI path promised by the comment never runs. Please distinguish an empty/no-data ROI from a populated zero-valued ROI and cover a constant negative source-min ROI in both in-memory and out-of-core paths. The 3D implementation has the same guard in `3d_intensity.cpp`.

vjaganat90 (approved):
LGTM

sameeul:
## Unbounded `double → unsigned` cast + duplicated map arithmetic (6 copies)

This PR reimplements the same "offset map" (`u = clamp_low(x - offset)` cast to an unsigned pixel type) independently in six places, and the missing-upper-bound bug shows up in some copies but not others — evidence the duplication itself is the problem:

| Location | Lower clamp | Upper clamp | `isfinite` check |
|---|---|---|---|
| `grayscale_tiff.h:367-375` `offset_map` (tile loader) | ✅ | ❌ | ✅ |
| `grayscale_tiff.h:699-708` `map_intensity` (strip loader) | ✅ | ❌ | ✅ |
| `omezarr.h:246-255` `map_intensity` | ✅ | ❌ | ✅ |
| `nyxus_dicom_loader.h:265-271` (inlined in the per-pixel loop, not even a named function) | ✅ | ❌ | ❌ |
| `raw_nifti.h:333-343` `rescale_offset` | ✅ | ❌ | ❌ |
| `slideprops.h:93-108` `to_grey_level` | ✅ | ❌ | ✅ |
| `raw_omezarr.h:143-149` `get_uint32_pixel` (added in this same PR) | ✅ | **✅** | ✅ |

Every one of the six offset-map copies casts a `double` to an unsigned integer type after clamping only the low end (`if (y < 0.0) y = 0.0; return (DataType) y;`). If `y` exceeds the target type's max (e.g. `UINT32_MAX` for `uint32_t`, or the narrower range for 8/16-bit grey buffers), the conversion is undefined behavior in C++, not merely a saturating/wrapping cast.

This is reachable through paths this PR itself adds tests for:
- `omezarr.h`'s copy is exercised by the new `signed.ome.zarr` / `float.ome.zarr` fixtures — a finite sample with `x - inten_offset_ > UINT32_MAX` hits it directly.
- `nyxus_dicom_loader.h`'s copy has no `isfinite` guard at all and reads `rescaleSlope_`/`rescaleIntercept_` straight from `DCM_RescaleSlope`/`DCM_RescaleIntercept` tags with no sanity check — an unusual or malformed DICOM file can produce a non-finite or out-of-range `y`.
- `slideprops.h::to_grey_level` sets whole-slide `aux_max` directly (`workflow_2d_whole.cpp`/`workflow_3d_whole.cpp` call it unconditionally on `p.max_preroi_inten`), so a single extreme value on a 32-bit or HU-preserving volume reaches it without needing a rare per-pixel event.

Meanwhile `raw_omezarr.h::get_uint32_pixel`, added in this same PR, *was* given the upper clamp and an `isfinite` check — its comment even explains why the clamp is needed and asserts it's "this accessor's alone... since a tile loader reads intensities through a recorded map rather than raw samples." But the "recorded map" it's contrasting itself with (`omezarr.h::map_intensity`) has exactly the gap this comment says doesn't need covering.

**Suggestion:** factor the offset-map arithmetic (`clamp-low`, cast, plus the `isfinite`→0 convention already shared by all six copies) into one function each format's loader and `SlideProps::to_grey_level` calls, and give it the same `isfinite` + full-range clamp `get_uint32_pixel` has. That fixes the UB everywhere at once and removes the risk of the six copies drifting again the next time this logic changes.

sameeul:
## Rounding silently became truncation on the offset map

Before this PR, every offset-map narrowing used `std::llround(y)` (round-to-nearest); this PR removes that call at all four sites and casts the fractional `double` straight down:

| Location | Before (`main`) | After (this PR) |
|---|---|---|
| `grayscale_tiff.h:375` `offset_map` (tile loader) | `return (DataType) std::llround(y);` | `return (DataType) y;` |
| `grayscale_tiff.h:708` `map_intensity` (strip loader) | `return (DataType) std::llround(y);` | `return (DataType) y;` |
| `raw_nifti.h:343` `rescale_offset` | `nyxbuf[i] = static_cast<til>(std::llround(y));` | `nyxbuf[i] = static_cast<til>(y);` |
| `nyxus_dicom_loader.h:271` (per-pixel loop) | `*(dest+i) = (DataType) std::llround(y);` | `*(dest+i) = (DataType) y;` |
| `slideprops.h:107` `to_grey_level` | `return (unsigned int) std::llround(y);` | `return (unsigned int) y;` |

The offset-map comments now describe this as `u = trunc(x - offset)`, so the truncating intent reads as deliberate rather than accidental — but the PR description doesn't call out the round→truncate change itself, and it has a real cost wherever `y` is fractional:

- **DICOM PET**: `hu = rescaleSlope_ * stored + rescaleIntercept_` — SUV/PET rescale slopes are routinely non-integer (e.g. `~1e-4`–`1`), so `y` is fractional for essentially every PET frame.
- **NIfTI**: `cur_scl_slope_`/`cur_scl_inter_` from the header can be fractional for PET-derived or otherwise rescaled volumes.
- **Real-valued TIFF/Zarr on the offset (non-quantized) branch**: `x` is already a float, so `y` is fractional on effectively every sample.

In each case, truncation biases every stored grey level down by up to ~1 unit relative to the old round-to-nearest behavior, and since `inten_scale` is left at `1.0` on the offset branch, `report_in_source_domain`'s inverse map has no way to recover the lost fraction — the bias is permanent and silent, not just an intermediate rounding step.

This PR's own regression/oracle fixtures don't catch it because they all use exact-integer scale factors (e.g. `scl_slope=2`), where round-to-nearest and truncation agree. Non-CT modalities with fractional slopes — the case this PR's `--preserve-hu` generalization is explicitly meant to now cover unconditionally — are exactly what's untested here.

**Ask**: is the round→truncate switch intentional? If so, worth a sentence in the PR description plus a fractional-slope regression fixture (e.g. a synthetic DICOM/NIfTI with `rescaleSlope`/`scl_slope` = 0.5 or similar) so the behavior is pinned rather than incidental. If not, restoring `std::llround` at these five sites would need no other change.

sameeul:
## While we're here: `val_COV` has the same zero-denominator gap this PR just guarded for QCOD/PIU

Not a regression from this PR — `val_COV = val_STANDARD_DEVIATION / mean_` (`intensity.cpp:193`) is byte-identical to `main`, and it already produced `0/0 = NaN` there for an all-zero-grey-level ROI. But this PR is exactly the place to close it, since the reasoning is already spelled out a few lines above it:

> `intensity.cpp:198-202` — "A populated ROI whose grey levels are all 0 does [have a distribution]... an all-air CT ROI at -1013 HU, say -- reaches here constant and zero, and its histogram, percentiles, entropy and UNIFORMITY are all well defined."

That's the exact scenario that makes `mean_ == 0` here, and it's also *not* CT-specific — it's any ROI on any identity-mapped slide (i.e. any ordinary non-negative-integer image, no HU/rescale involved) whose pixels are uniformly grey level 0. QCOD and UNIFORMITY_PIU got guards for precisely this a few lines down (`intensity.cpp:222`, `248`); COV, sitting right above them, didn't.

Since the guard pattern is already established in this diff, six sites, same fix, same convention (`0.0` on a zero denominator, matching QCOD/PIU/HYPERSKEWNESS/HYPERFLATNESS):

- `intensity.cpp:193` (`calculate_grey_levels`)
- `intensity.cpp:130` (`report_in_source_domain`)
- `intensity.cpp:346` (`osized_calculate_grey_levels`)
- `3d_intensity.cpp:183` (`calculate_grey_levels`)
- `3d_intensity.cpp:118` (`report_in_source_domain`)
- `3d_intensity.cpp:334` (`osized_calculate_grey_levels`)

e.g.:
```cpp
val_COV = mean_ != 0.0 ? val_STANDARD_DEVIATION / mean_ : 0.0;
```
(and `val_MEAN` in place of `mean_` for the two `report_in_source_domain` copies).

Low-cost since the convention, the tests that would need touching, and the trivial/non-trivial parity concern are all already in scope for this diff — bundling it here avoids a seventh near-identical PR later to fix the same thing.

darkclad:
**Fixed** Added the `isfinite` guard to the DICOM and NIfTI maps in ce35a99c, then moved all eleven narrowing sites onto `Nyxus::grey_level_*` in `grey_level_cast.h` in 86c1e2de, which zeroes non-finite values and saturates above the grey type's maximum. The quantized maps keep their finiteness check ahead of the clamp, so `+Inf` still stores grey level 0.

darkclad:
**Diverted** Restored rounding only under `--preserve-hu`, where it originally lived, because rounding every site broke 198 3D test cases (ce35a99c). A DICOM or integer NIfTI read without the flag now keeps its stored integers and carries the rescale in the inverse, so a fractional PET slope is reported exactly rather than rounded or truncated (24533f7a). `ct3d_frac.nii` and `ct_frac.dcm` pin both paths.

darkclad:
**Fixed** Guarded `val_COV` at all six sites with the same zero-denominator convention as QCOD and PIU, including the two `report_in_source_domain` copies this PR added (ce35a99c). e6b24066 adds the missing 2D and 3D degenerate-ROI assertions, which fail when the guard is reverted.

src/nyx/slideprops.cpp
darkclad: **Fixed** OME-Zarr now takes the recorded offset or quantization map like every other backend, and the raw loader holds samples as double, so the prescan sees negatives and fractions (a1cfdd56). `signed.ome.zarr` and `float.ome.zarr` pin both, and `get_uint32_pixel` narrows through the shared saturating helper (86c1e2de).

src/nyx/slideprops.cpp
darkclad: **Fixed** A constant real-valued slide records the offset map at its unfloored minimum, so a constant 0.5 reads back as 0.5 instead of 0 (a1cfdd56). An all-non-finite slide, which reaches the same branch, is settled to a flat zero range and no longer reports `DBL_MAX` (cb7b70d2).

src/nyx/features/intensity.cpp
darkclad: **Fixed** All four blank-ROI guards now test the pixel cloud for emptiness, and `TrivialHistogram` puts a zero-width range in its first bin instead of dividing by it (a1cfdd56). Letting that ROI through exposed NaN in QCOD, UNIFORMITY_PIU and COV, which are now guarded and tested in 2D and 3D (13460faf, ce35a99c, e6b24066).

PR #467, opened 2026-09-01
test(vetting): measure the three morphology contour cells, and find an out-of-core defect

darkclad:
Follow-up to #425 (now merged), which scoped its vetted claim to the in-RAM `SINGLEROI=false`
assertion and left the other production cells unmeasured. This measures them — and running it
found a defect.

SPEC 5.1 requires every cell of a family's config matrix to carry one of three dispositions
-- VALID, VALID-BUT-PRODUCTION-ONLY, INVALID. matrix/morphology.md left two reachable
production cells labelled NOT MEASURED, which is none of them. Both are now measured, per
SPEC 5.2's rule that verdicts are measured rather than hand-labelled, on one 64x64
single-disk fixture so the three cells are comparable numbers on one shape:

  feature                    segmented        out-of-core      whole-slide
  PERIMETER                  131.88225099     112.0            256.0
  MASS_DISPLACEMENT          2.75261401       2.75261401       3.34531182
  EDGE_MEAN_INTENSITY        257.0            257.0            397.0
  EDGE_STDDEV_INTENSITY      98.12659593      98.12659593      0.0
  EDGE_MAX / EDGE_MIN        397.0 / 117.0    397.0 / 117.0    397.0 / 397.0
  EDGE_INTEGRATED_INTENSITY  28784.0          28784.0          1588.0

The whole-slide cell is VALID-BUT-PRODUCTION-ONLY and degenerate exactly as the source
predicts: buildWholeSlideContour() pushes the four AABB corners each carrying aux_max, so
min == max == mean == the image maximum, stddev is 0 and integrated is 4*maximum. It is a
different quantity from the per-object edge statistics CellProfiler vets in #425, which is
why those rows scope themselves to the segmented in-RAM path.

The out-of-core cell agrees exactly for the five EDGE_* statistics and MASS_DISPLACEMENT --
and does NOT agree for PERIMETER. osized_calculate() sets fval_PERIMETER = K.size(), the
contour pixel count, where calculate() sums Euclidean step lengths: 112.0 against
131.88225099390849, ~15%, and the out-of-core value being an integer is what identifies the
definition rather than accumulation as the cause. CLAUDE.md requires the two paths to produce
identical values, so this is a defect and not a convention. It is characterized here, not
fixed: the fix is a src/nyx change and PERIMETER is vetted under a different recipe.

Why nothing caught it: test_2d_ooc_invariant.py has asserted *ALL_MORPHOLOGY* equality across
the two paths since it was written, and passes. Its fixture is a full-image RECTANGLE, and
around a rectangle every contour step is an axis-aligned unit step, so the pixel count and
the Euclidean sum are the same number. The invariant is sound; that shape cannot discriminate.
A disk separates them. The general form -- a path-equality fixture symmetric in the axis under
test proves nothing about that axis -- is recorded in not_covered.md section G, with a note to
audit the other _ooc_ fixtures on the same grounds.

Three assertions and seven registry rows back the cells: a discriminating-fixture invariant
for the six features that agree, test_2d_ooc_regression.py pinning the PERIMETER divergence so
a correct fix must break it, and test_2d_morphology_regression.py pinning the whole-slide cell
and asserting it differs from the segmented one so the CellProfiler evidence cannot be read
across. The coverage report is unchanged: these are regression rows for features whose rollup
is already vetted.

Still no src/nyx change.

## Rebased onto merged `main`

`audit/scan_morphology_coverage.py` arrived on `main` while this branch was out of date and
rejected it with 18 problems. Both causes are fixed:

- **The new assertions were invisible to it.** It credits a feature when the name appears on an
  assertion line, in an inline loop list, or in a `TABLE_OWNER` table; these tests looped over
  module-level dicts, so no feature name reached an assertion line. They now name their feature
  explicitly — the style the rest of the family uses, and a failure now says which feature moved.
  The three files are registered in `SOURCES`.
- **`current_test` is a per-feature rule**, not per-row: every row for a feature must name all
  files covering it. Fixing that surfaced a gap the registry never recorded —
  `test_2d_ooc_invariant.py` has been asserting OOC-vs-in-RAM equality for `ECCENTRICITY`,
  `ELONGATION`, `ORIENTATION`, `ROUNDNESS`, `MAJOR/MINOR_AXIS_LENGTH` and `EROSIONS_2_VANISH`
  since it was written, and no row said so. Those seven rows now name it. Pre-existing gap
  exposed, not new coverage.

`morphology_2d_coverage.csv` is regenerated, not edited. The `current_test` changes are spliced
field-by-field rather than round-tripped through `csv.DictWriter`, which drops quotes on any
field containing a `;` but no comma — that silently rewrote 34 unrelated 3D rows on the first
attempt. The registry diff is 21 added / 14 removed, all `2D,`, no churn elsewhere.

## Self-review against the series' own rules

Checked against `PR/open-review-questions.md`; two findings, both fixed.

**The seven new rows left `test_name` and `benchmark` empty** — the shape #437 review item 6
records, where a row passes `check_coverage.py` and is still an empty-column row.
`benchmarks.md` now defines **`bench_disk64_diagonal_boundary`** and all seven name it. The
fixture is worth a definition rather than prose: 1257 pixels, **112** of them on the
4-neighbour inner boundary — and 112 is exactly what the out-of-core path returns for
`PERIMETER`, so the entry carries the number that identifies the divergence as definitional.
Negative-controlled: renaming the heading makes `--check` reject all seven rows. `test_name`
stays empty because that column takes a gtest `SUITE.CASE` and these assertions are pytest;
the notes name the asserting function instead. **These are the registry's first pytest-only
rows**, so there is no precedent for that column — flagged rather than papered over.

**The matrix table presented its 21 cells as equally backed.** They are not: the segmented and
whole-slide columns are pinned literally; in the out-of-core column only `PERIMETER` is, and
the other six are asserted *equal to* the segmented column. All 21 were verified identical to
the pins, and the table now says which is which — including what the equality form would miss,
a change that moved both paths together.

Also confirmed clean: no `src/` change (question 1 — the defect is characterized, not fixed);
none of the contour `+1`-offset features (question 6) are claimed here; the generated CSVs are
regenerated by their writer, not edited; and the name checker was shown to actually key on the
new files by planting a bad name and watching it fire.

## Blocker round

Four blockers, all correct.

**`DIAMETER_EQUAL_PERIMETER` diverges too, and was missed.** Both paths compute it as
`fval_PERIMETER / M_PI` (`contour.cpp` 976, 1000), so it carries the same ratio exactly:
35.65070725258456 against 41.97942430353313. One defect with two public consequences, not two
defects — `test_2d_ooc_regression.py` asserts the derivation as an identity on both paths, so a
fix to `PERIMETER` alone must fix this one and a run where the ratios stop matching is a new
finding. Added to the matrix, its own rows, and §G.

**`MASS_DISPLACEMENT` had the wrong producer and an unmeasured verdict.** Its whole-slide note
was the shared contour explanation, which does not apply — it is
`BasicMorphologyFeatures::calculate()` and never reaches `buildWholeSlideContour()`. Running
CellProfiler with one full-frame object, as asked, **changed the verdict**: CP agrees at
3.345311793150965 against 3.3453118163885427, `rel=7e-9`, so that cell is **VALID**, not
production-only. Its `EDGE_*` siblings are not — an all-ones label image has an empty
`find_boundaries(mode="inner")` set, so CP returns exactly 0 for each while Nyxus reports four
synthesised corners. `gen_morphology_wholeslide_cellprofiler.py` asserts that zero rather than
describing it; promoting the row without a generator would have recreated the unbacked `vetted`
row #425 existed to remove.

**The SPEC dispositions are now literal.** `VALID` means oracle-backed and nothing else.
Path-equality cells are **`INVARIANT`** — SPEC §3 lists the status and `check_coverage.py` never
implemented it, so such a cell had to be written `vetted`, claiming a tool agreed when none was
involved. It now ranks with `regression` (cannot raise the headline), has its own report column,
and two self-tests. The out-of-core contour cells are **`impl-defect`**, not `INVALID`: that is
for degenerate configs dropped with a reason, and this is production code reached whenever an ROI
exceeds `ram_limit`.

**One row per cell.** 7 rows → **24**, distinguished by four new path-specific recipes
(`morphology.disk64_{segmented_inram,forced_ooc,wholeslide}`,
`morphology.cellprofiler_wholeslide_massdisp`). `current_test` stays the full covering set per
feature because that is what `scan_morphology_coverage.py` enforces; the cell is carried by
`config_recipe`. The disk fixture's three independent builders are now one,
`test_data.disk64_arrays()`.

## Verification

- `scan_morphology_coverage.py --check`: **123 rows, clean**; `coverage_report.md` and
  `morphology_2d_coverage.csv` both regenerate byte-identical
- `check_coverage.py --check`, `check_test_names.py --check`: clean
- `pytest tests/python/`: **98 passed / 1 skipped** (7 pre-existing Arrow failures of a
  tiff-only build)
- **ASan + UBSan** (gcc 13, `-fsanitize=address,undefined -fno-sanitize-recover=undefined`,
  Linux): gtest **883 passed / 1 skipped** of 884, exit 0; instrumented `pytest tests/python/`
  **97 passed / 1 skipped / 7 Arrow**, with `test_nonimq_wsi_scalability` deselected for the
  pre-existing `glszm.cpp:426` UBSan finding on `main`. **0 diagnostics** on both legs, and the
  gate log's md5 block matches the tree on every changed file.
- No `src/nyx` change; LF throughout, no whitespace churn


tests/python/test_2d_ooc_invariant.py
```diff
@@ -185,6 +185,69 @@ def test_2d_ooc_2d_morphology_matches_in_ram_invariant(tmp_path):
...
+    "EDGE_MAX_INTENSITY", "EDGE_MIN_INTENSITY", "EDGE_INTEGRATED_INTENSITY",
+]
+
+
+def test_2d_ooc_2d_contour_intensity_matches_in_ram_on_diagonal_boundary_invariant(tmp_path):
+    """The five EDGE_* statistics and MASS_DISPLACEMENT must agree out-of-core on a shape whose
+    boundary is diagonal, not only on the rectangle the other tests use.
+
+    This is the cell tests/vetting/matrix/morphology.md records for the out-of-core contour path.
+    PERIMETER is deliberately NOT in the list: it does NOT agree, and that divergence is pinned as a
```
vjaganat90: The diagonal fixture is a good catch. **Blocker:** the defect scope stops one feature too early. Both `ContourFeature::calculate()` and `osized_calculate()` immediately set `DIAMETER_EQUAL_PERIMETER = fval_PERIMETER / M_PI` from their differing perimeter values. On this fixture that is about `41.9794243` in RAM versus `35.6507073` out of core, and the rectangle hides this derived divergence for exactly the same reason. Please characterize both features in the regression test, matrix, registry, and §G instead of recording `PERIMETER` as the only defect. The six genuinely agreeing features can stay as they are.

tests/vetting/matrix/morphology.md
```diff
@@ -0,0 +1,91 @@
...
+reads a contour, and it has its own `BasicMorphologyFeatures::osized_calculate()`. It therefore gets
+its own axis below, which happens to land on the same dispositions — but by measurement, not by
+inheritance.
+
+### `ContourFeature` — the axis is which contour builder runs
+
+| cell | path | EDGE_* verdict | PERIMETER verdict |
+|---|---|---|---|
+| segmented, in-RAM | `calculate()` at `SINGLEROI=false` → `buildRegularContour()` | **VALID** — cellprofiler, recipe `morphology.cellprofiler_edge_intensity` | **VALID** — skimage, recipe `morphology.perimeter_circles` |
+| out-of-core | `osized_calculate()` → `buildRegularContour_nontriv()` | **VALID** — equals the in-RAM cell exactly (invariant, measured) | **INVALID — defect**, see below |
```
vjaganat90: This measurement is useful, but **blocker:** these two labels no longer mean the SPEC §5.1 dispositions. `VALID` is defined as an oracle-backed cell that gets a `vetted` row; the EDGE_* result here is an invariant against a separately vetted path. `INVALID` is defined as a degenerate/non-production config that gets `dropped_invalid`; this out-of-core cell is a real production path with an implementation defect and must not be dropped. Please record the first as invariant-backed, and the second as a reachable valid config blocked by `impl-defect` (or extend the vocabulary explicitly). Reusing `INVALID` for a bug makes a path that must be fixed read like a config that should be discarded.

tests/vetting/oracle_coverage.csv
```diff
@@ -134,6 +134,13 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
+2D,MASS_DISPLACEMENT,morphology,regression,,na,,rel=1e-9,test_2d_morphology_regression.py,test_2d_morphology_regression.py,,production-only-cell,in-tree,"VALID-BUT-PRODUCTION-ONLY cell, NOT vetted. SINGLEROI=true reaches ContourFeature::buildWholeSlideContour, which pushes the four AABB corners each carrying aux_max instead of tracing a boundary, so every edge statistic is degenerate by construction: min=max=mean=the image maximum, stddev=0, integrated=4*maximum. A different quantity from the per-object edge statistics the segmented cell computes, so no external tool reproduces it and the cellprofiler evidence does not carry across. Snapshotted in tests/python/test_2d_morphology_regression.py, which also asserts the two cells disagree. matrix/morphology.md.",,
```
vjaganat90: The matrix now separates the producer correctly, but **blocker:** this row reintroduces the attribution fixed in PR 425. `MASS_DISPLACEMENT` comes from `BasicMorphologyFeatures::calculate()` using geometric and intensity-weighted centroids; it never reaches `buildWholeSlideContour()`, so the contour/edge explanation does not support this pin. It is also too early to call this cell production-only: CellProfiler exposes the same `MeasureObjectIntensity` mass-displacement quantity, and a label covering the full 64×64 image is a direct mapping for this cell. Please run that candidate mapping before assigning the measured verdict, then give this row its own producer-specific note and recipe (or record the measured disagreement if one appears).

tests/vetting/oracle_coverage.csv
```diff
@@ -134,6 +134,13 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
+2D,PERIMETER,morphology,regression,,disagrees,,rel=1e-9,test_2d_ooc_regression.py,test_2d_ooc_regression.py,,impl-defect,in-tree,"NOT vetted -- known DEFECT, characterized. ContourFeature::osized_calculate sets fval_PERIMETER = K.size(), the contour pixel count, where calculate() sums Euclidean step lengths, so the out-of-core path returns 112.0 against the in-RAM 131.88225099390849 on a 64x64 disk -- 15%, and the OOC value is an integer, which identifies the definition rather than accumulation as the cause. CLAUDE.md requires the two paths to produce identical values, so this is a defect not a convention. Not caught before because test_2d_ooc_invariant.py's fixture is a full-image rectangle, where the pixel count and the Euclidean sum coincide. Pinned in tests/python/test_2d_ooc_regression.py, which a correct fix must break. matrix/morphology.md.",,
```
vjaganat90: **Blocker:** the seven added rows do not yet back all of the assertions this follow-up introduces. `test_2d_morphology_regression.py` pins whole-slide `PERIMETER`, but there is no row for that cell; the new diagonal-boundary invariant asserts six features, but none has an `invariant` row. The added rows also leave `config_recipe`, exact test name, and `benchmark` empty, so the registry cannot resolve the path, assertion, or fixture behind each verdict. Please define the disk/whole-slide recipes and benchmark and add one row per assertion as SPEC §3 requires. I would also make the disk builder/constants shared: right now the matrix’s “same shape” claim depends on three copied fixture implementations staying synchronized, and all three tests can remain green after one copy drifts. `check_coverage.py --check` currently validates pointers that are present; it does not detect these omissions.

tests/python/test_2d_ooc_invariant.py
```diff
@@ -185,6 +185,67 @@ def test_2d_ooc_2d_morphology_matches_in_ram_invariant(tmp_path):
...
+    assert abs(a - b) <= 1e-6 * max(abs(a), abs(b), 1.0) + 1e-9, (
+        "out-of-core %s diverges from in-RAM: %r vs %r" % (feature, a, b))
+
+
+def test_2d_ooc_2d_contour_intensity_matches_in_ram_on_diagonal_boundary_invariant(tmp_path):
+    """The five EDGE_* statistics and MASS_DISPLACEMENT must agree out-of-core on a shape whose
+    boundary is diagonal, not only on the rectangle the other tests use.
+
+    This is the cell tests/vetting/matrix/morphology.md records for the out-of-core contour path.
+    PERIMETER is deliberately NOT among them: it does NOT agree, and that divergence is pinned as a
```
vjaganat90, line 219: **Blocker — characterize the derived feature too.** This is not a `PERIMETER`-only defect. In `contour.cpp`, both paths assign `DIAMETER_EQUAL_PERIMETER = fval_PERIMETER / M_PI` immediately after computing their different perimeter values. Therefore this same disk returns about `41.9794243035` in RAM and `35.6507072526` out of core; the rectangle masked both divergences.

Please make the next pass do all of the following:
1. Add `DIAMETER_EQUAL_PERIMETER` to the known-defect regression and pin both values.
2. Add its `impl-defect` registry row and include it in `matrix/morphology.md` and `not_covered.md` §G.
3. Update the prose/PR description so nothing says `PERIMETER` is the only affected feature.

Acceptance condition: the diagonal fixture fails if either derived value moves, and both affected feature names resolve through the registry.

tests/vetting/matrix/morphology.md
```diff
@@ -1,77 +1,95 @@
...
```
vjaganat90: **Blocker — use the SPEC dispositions literally.** This row currently assigns two meanings that §5.1 does not define:

- `VALID` means oracle-backed and maps to an oracle test plus a `vetted` row. These EDGE_* measurements are backed by an in-RAM/out-of-core invariant, not by an oracle at the out-of-core cell.
- `INVALID` means a degenerate/non-production configuration that is dropped with a `dropped_invalid` row. The out-of-core path is reachable production code with an implementation defect; it must be fixed, not dropped.

Please label the agreeing cells explicitly as invariant-backed and give them `status=invariant` rows. Label the perimeter cell as a reachable production cell blocked by `impl-defect` (or add an explicit matrix term for that state). Acceptance condition: no invariant is presented as oracle-vetted, and no real production path is presented as a discardable invalid config.

tests/vetting/oracle_coverage.csv
```diff
@@ -134,6 +134,13 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
+2D,MASS_DISPLACEMENT,morphology,regression,,na,,rel=1e-9,test_2d_morphology_cellprofiler.h;test_2d_morphology_regression.py;test_2d_ooc_invariant.py,test_2d_morphology_regression.py,,production-only-cell,in-tree,"VALID-BUT-PRODUCTION-ONLY cell, NOT vetted. SINGLEROI=true reaches ContourFeature::buildWholeSlideContour, which pushes the four AABB corners each carrying aux_max instead of tracing a boundary, so every edge statistic is degenerate by construction: min=max=mean=the image maximum, stddev=0, integrated=4*maximum. A different quantity from the per-object edge statistics the segmented cell computes, so no external tool reproduces it and the cellprofiler evidence does not carry across. Snapshotted in tests/python/test_2d_morphology_regression.py, which also asserts the two cells disagree. matrix/morphology.md. Asserted by test_2d_morphology_whole_slide_edge_intensity_is_degenerate_regression / test_2d_morphology_whole_slide_differs_from_segmented_regression in test_2d_morphology_regression.py; test_name is empty because that column takes a gtest SUITE.CASE and this assertion is pytest.",,bench_disk64_diagonal_boundary
```
vjaganat90: **Blocker — this row still describes the wrong producer and assigns an unmeasured verdict.** `MASS_DISPLACEMENT` is computed by `BasicMorphologyFeatures::calculate()` from the geometric and intensity-weighted centroids. It never calls `ContourFeature::buildWholeSlideContour()`, so the four-corner/edge-statistic explanation cannot support this row.

There is also a direct accepted-oracle candidate: run CellProfiler `MeasureObjectIntensity` with one object covering the full 64×64 image and the same intensity image. That is the same centroid-distance quantity as this whole-image Nyxus cell.

Please:
1. Remove all contour-builder language from this row.
2. Run that full-image CellProfiler mapping before choosing the verdict.
3. If it agrees, add a producer-specific recipe and oracle assertion; if it disagrees, record the measured residual/definition and keep the regression disposition.

Acceptance condition: the row names `BasicMorphologyFeatures`, and `VALID-BUT-PRODUCTION-ONLY` is used only after the obvious CellProfiler mapping has been measured and ruled out.

tests/vetting/oracle_coverage.csv
```diff
@@ -134,6 +134,13 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
...rphology,regression,,disagrees,,rel=1e-9,test_2d_morphology_regression.h;test_2d_morphology_regression.py;test_2d_morphology_skimage.h;test_2d_ooc_invariant.py;test_2d_ooc_regression.py;test_nyxus.py,test_2d_ooc_regression.py,,impl-defect,in-tree,"NOT vetted -- known DEFECT, characterized. ContourFeature::osized_calculate sets fval_PERIMETER = K.size(), the contour pixel count, where calculate() sums Euclidean step lengths, so the out-of-core path returns 112.0 against the in-RAM 131.88225099390849 on a 64x64 disk -- 15%, and the OOC value is an integer, which identifies the definition rather than accumulation as the cause. CLAUDE.md requires the two paths to produce identical values, so this is a defect not a convention. Not caught before because test_2d_ooc_invariant.py's fixture is a full-image rectangle, where the pixel count and the Euclidean sum coincide. Pinned in tests/python/test_2d_ooc_regression.py, which a correct fix must break. matrix/morphology.md. Asserted by test_2d_ooc_perimeter_diverges_from_in_ram_regression in test_2d_ooc_regression.py; test_name is empty because that column takes a gtest SUITE.CASE and this assertion is pytest.",,bench_disk64_diagonal_boundary
```
vjaganat90: **Blocker — `current_test` expansion is not a substitute for one row per assertion/config cell.** The benchmark addition is good, but the registry is still missing:

1. A whole-slide `PERIMETER` regression row for the literal pin in `test_2d_morphology_regression.py`.
2. Six out-of-core `status=invariant` rows for `MASS_DISPLACEMENT` plus the five EDGE_* assertions in `test_2d_ooc_invariant.py`.
3. Path-specific config recipes that distinguish segmented in-RAM, forced out-of-core (`ram_limit=0`), and whole-slide (`single_roi=True`).
4. The `DIAMETER_EQUAL_PERIMETER` defect row requested above.

Please also put the disk fixture builder/constants in one shared helper (or add a check proving the three copies are identical). The matrix’s central claim is that all columns use one shape; three independent builders can drift while every test remains green.

Acceptance condition: each matrix cell points to its own registry row/config recipe, and changing one fixture definition cannot silently invalidate the cross-column comparison. The pytest `test_name` limitation can remain documented; that is not what blocks this.

tests/python/test_2d_ooc_invariant.py
darkclad: Fixed — DIAMETER_EQUAL_PERIMETER characterized alongside PERIMETER (35.65070725258456 vs 41.97942430353313), with the /pi derivation asserted as an identity on both paths; matrix, rows and §G updated.

tests/python/test_2d_ooc_invariant.py
darkclad: Fixed — DIAMETER_EQUAL_PERIMETER characterized alongside PERIMETER (35.65070725258456 vs 41.97942430353313), with the /pi derivation asserted as an identity on both paths; matrix, rows and §G updated.

tests/vetting/matrix/morphology.md
darkclad: Fixed — VALID is oracle-backed only; path-equality cells are status=invariant (implemented in check_coverage.py with its own rollup bucket and self-tests), and the out-of-core contour cells are flag=impl-defect rather than INVALID.

tests/vetting/matrix/morphology.md
darkclad: Fixed — VALID is oracle-backed only; path-equality cells are status=invariant (implemented in check_coverage.py with its own rollup bucket and self-tests), and the out-of-core contour cells are flag=impl-defect rather than INVALID.

tests/vetting/oracle_coverage.csv
darkclad: Fixed — producer corrected to BasicMorphologyFeatures::calculate(), and CellProfiler run with one full-frame object agrees (3.345311793150965 vs 3.3453118163885427, rel 7e-9), so the cell is VALID not production-only; new gen_morphology_wholeslide_cellprofiler.py backs it and asserts that CP returns 0 for the EDGE_* it cannot vet.

tests/vetting/oracle_coverage.csv
darkclad: Fixed — producer corrected to BasicMorphologyFeatures::calculate(), and CellProfiler run with one full-frame object agrees (3.345311793150965 vs 3.3453118163885427, rel 7e-9), so the cell is VALID not production-only; new gen_morphology_wholeslide_cellprofiler.py backs it and asserts that CP returns 0 for the EDGE_* it cannot vet.

tests/vetting/oracle_coverage.csv
darkclad: Fixed — 7 rows became 24, one per (feature x cell), under four new path-specific recipes (morphology.disk64_{segmented_inram,forced_ooc,wholeslide} and morphology.cellprofiler_wholeslide_massdisp); the disk fixture builders are consolidated into test_data.disk64_arrays().

tests/vetting/oracle_coverage.csv
darkclad: Fixed — 7 rows became 24, one per (feature x cell), under four new path-specific recipes (morphology.disk64_{segmented_inram,forced_ooc,wholeslide} and morphology.cellprofiler_wholeslide_massdisp); the disk fixture builders are consolidated into test_data.disk64_arrays().

vjaganat90:
LGTM

PR #379, opened 2026-07-17
fix(glcm): normalize HOM2/ENTROPY by sum_p (2D+3D) + fully oracle-vet…

darkclad:
… GLCM

f_GLCM_HOM2/f_entropy summed the unnormalized co-occurrence matrix (missing /sum_p that f_GLCM_JE/f_idm have), giving homogeneity>1 and negative entropy in both glcm.cpp and 3d_glcm.cpp. Add /sum_p; HOM2->IDM, ENTROPY->JE. Add IBSI gtests (2D) + a 3D config-sensitive equivalence gtest; update the stale regression goldens. Oracle-vet the remaining GLCM regression features vs MIRP/IBSI (2D) and by calc_ave / twin equivalence vs pyradiomics (3D): glcm 72->118 vetted (registry now 502/758). Full gtest suite 706/706.

PR #470, opened 2026-09-02
chore(vetting): one coverage-scanner library, and a feature report that checks the registry against the tree

darkclad:
# chore(vetting): one coverage-scanner library, and a feature report that checks the registry against the tree

34 files: 33 under `tests/vetting/` and one CI workflow. **No test source is touched** — not a
`.h`, not a `.cc`, not a `tests/python/` file; only the vetting tooling. **No `src/nyx` change**:
`git rev-parse HEAD:src` equals `git rev-parse upstream/main:src` (`4edf8045`). Nothing is deleted;
nine files are new and twenty-five change. Net **-326 lines**, because the twenty scanners drop from
3,839 lines to 601 and share one 788-line library instead of twenty copies of the same rule.

Eight commits, all `chore(vetting)`, so **nothing here reaches the CHANGELOG and nothing moves the
version** whether this is squashed or merge-committed. That is deliberate: `release-please-config.json`
sends `feat` to Features and bumps the patch pre-1.0, and `refactor` to Code Refactoring, and none of
this is observable to anyone using Nyxus.

---

## The problem

`oracle_coverage.csv` records what a row **claims** — a status, an oracle, a tolerance, a recipe.
The per-family scanners under `audit/` record what the **tree asserts** — which gtest case covers
which feature, under which oracle, with which drift guard behind it.

Neither could check the other. `coverage_report.md` is a sixteen-row tally of the registry against
itself; it cannot see a test at all. The two sides were only ever compared per family, by hand, when
somebody happened to run a scanner — and five families had no scanner to run.

That is the gap this closes, and it needed the three steps in order.

---

## 1. The one family with no coverage artifact

3D GLDZM was the only family in the registry with no coverage artifact at all: `scan_gldzm_coverage.py`
is 2D-only, so its 18 rows had nothing to be checked against, scanned or hand-written. Five other
family x dim combinations do have an artifact, but nothing re-derives it from the tree -- the report
names all five under *Families with no scanner*, and writing those scanners is separate work.

A `gldzm_3d_mirp_vetting_report.md` already existed, so the vetting had been done; only the artifact
that re-derives it from the tree was missing.

`scan_gldzm3d_coverage.py` + `gldzm_3d_coverage.csv` close it, modelled on the 3D NGLDM scanner —
the other family with no oracle test, for the same reason. `ORACLE_SUFFIX` is deliberately empty:
MIRP is config-matched and reproducible, but disagrees with Nyxus on all 16 features it computes by
up to **67.5×**, and an independent from-definition recomputation (26-connected zones, city-block
distance transform) reproduces MIRP to **rel 3.2e-16**. The definition is reachable and Nyxus is not
computing it, so nothing here may be promoted on the current implementation.

Its `--check` runs the assertion-identity tier, which is the tier that was missing when `3GLDZM_ZDM`
carried a golden of `222` — a factor of 14.5 off — with no test function and no `TEST()`
registration, so nothing could fail on it.

`gldzm_3d_golden_regen.md` records the regeneration path and one gap it does not close: this is the
only 3D family with no `*_dump_regression` test, so its pin table is transcribed by hand. That is a
C++ change and needs the ASan gate; it is not in this PR.

## 2. One library instead of twenty copies

`not_covered.md` already recorded that the coverage logic is copied across every `scan_*` file. The
cost was measurable: the twenty had drifted into four spellings of the function regex and two of the
comment stripper, and a fix applied to one copy stayed in one copy.

`audit/scanlib.py` holds it once — the coverage rule, the scan, the acceptance checks, the artifact
rendering and the run loop. **4,993 lines across 19 scanners become 1,755 across 20, plus a 788-line
library.** Thirteen families are now a docstring and one `Family(...)` call, the shortest 33 lines.
The library carries 238 lines more than the first version of this refactor, all of it the tightened
coverage rule the review rounds below asked for; `scanlib_selftest.py` (304) and `run_scanners.py`
(60) are new files with no equivalent before, and are counted in the diff rather than here.

**What is deliberately not unified**, because the difference is real and belongs where the family is
declared:

| family | what it does differently |
|---|---|
| 2D NGTDM, IMQ, 3D GLDM, 3D GLSZM | read `current_test` per test **kind** — a vetted row answers to the oracle files, a regression row to the snapshot one. Stricter than the shared per-feature reading, and catches a class it cannot |
| 2D moments | resolves coverage through golden tables; nothing in it names a feature on an assertion line |
| 2D radial | declares `test_2d_radial_mechanics.h` UNCREDITED — its pins characterize known defects, so crediting them would make the defects acceptance criteria |

Each keeps its own `collect` or checks and says why in its own docstring.

### Three behaviour changes, all in 2D morphology, all corrections

**a. `helper_features` over-credited coverage in 11 of 14 scanners.** A module-level pytest helper's
body was bounded at the next *helper* rather than the next top-level `def`, so the last helper in a
file swallowed every test function below it and each caller inherited every feature name in the
file. Three scanners carried the fix; eleven did not. scanlib takes the correct bound, which exposes
**eight `current_test` cells naming a file that covers nothing for them**:

- four chord rows (`MAXCHORDS_MAX/MIN`, `ALLCHORDS_MAX/MIN`) named `test_2d_morphology_invariant.py`,
  which reads those four in an `if` guard and asserts only the `_ANG` features;
- four `PERIMETER` rows named `test_2d_ooc_invariant.py` — **whose own docstring says so**:
  *"PERIMETER is deliberately NOT among them: it does NOT agree, and that divergence is pinned as a
  regression."* The rows and the file contradicted each other outright and nothing could see it.

Each of the eight keeps its other files; none is left empty.

**b. A pytest case can assert features that rule cannot see.** `test_2d_ooc_invariant.py` names the
six ellipse features in a local tuple and `EROSIONS_2_VANISH` in the comprehension that builds its
compared-column list, then loops that list while asserting — so no name reaches an assertion line.
Under the corrected bound those seven would have read as *uncovered*, which would have been wrong:
the case really does compare them. scanlib gains `py_loop_tables`, the Python twin of the
`loop_tables` concession the C++ equivalence tests already had, gated the same way — the function
must both assert and range-loop, so a plain lookup list is not coverage. Opt-in; 2D morphology is the
only family that takes it.

**c. `morphology_2d_coverage.csv` repeated eight features four times each** — 123 lines for 99
features, byte-identical repeats carrying no extra information. That scanner rendered one line per
*registry* row, and the registry has moved to one row per config cell. The artifact is a
feature→test rollup, so scanlib renders one line per feature and the 24 duplicates are gone. Any
family whose registry grows config-cell rows would have hit this.

## 3. The report, and the CI gate

`report_features.py` generates `features.csv` (923 rows) and `features.md`, one row per
**`(dim, feature, oracle, config_recipe)`** — the key the registry is already unique on, and the one
that keeps a feature's config cells apart, which matters because a feature vetted at `ibsi=true` can
diverge by two orders of magnitude at `ibsi=false`. `--check` enforces that uniqueness; it was true
but unchecked, which is the shape of #443's C3.

Columns are grouped `claim_*` / `scan_*` so a reader can see which side of the join a value came
from, and the derived **`verdict`** says whether they agree. That is what makes this a gate rather
than a summary.

**It reads the tree through the scanners themselves**, importing each `scan_*_coverage.py` and
running its `FAMILY` through scanlib, so the report and the per-family `--check` cannot disagree
about what a test covers. That is only possible because step 2 gave them a shared library; without
it, a tree-wide report means a twenty-first copy of the coverage rule — which is exactly how
`report_feature_tests.py` came to credit tests with features they never checked.

### What it found

**745 of 923 rows agree. None disagree. 178 (19%) come back `unscanned`:**

| dim | family | rows | features |
|---|---|---:|---:|
| 2D | firstorder | 36 | 36 |
| 2D | glcm | 59 | 59 |
| 2D | glrlm | 32 | 32 |
| 2D | gabor | 2 | 1 |
| 3D | firstorder | 49 | 36 |

These five have an `audit/<family>_<dim>_coverage.csv` with **no scanner behind it**. Those artifacts
record what someone believed rather than what the tree asserts, and no check anywhere could say so —
including all of first-order and all of 2D GLCM, two of the most-claimed families. They are declared
in `KNOWN_UNSCANNED` with a reason each, `features.md` gives the number a section of its own, and
`--check` fails if a sixth family appears undeclared. Writing those five scanners is the next branch;
2D GLCM and 2D GLRLM each have a working 3D twin to copy.

CI gains two steps beside the existing two: the twenty scanners in one loop, and
`report_features.py --check`. Both are stdlib-only and need no build.

### Six stale cells in the families no scanner covers

The five families with no scanner are reported as `unscanned` rather than green, which is honest but
is not the same as checked — and four of the five were vetted by the #42x–#46x series this PR
consolidates. Running a throwaway `scanlib.Family` declaration over them says how much that
distinction is worth:

- **2D GLCM — the registry itself was wrong, in two rows.** `test_2d_glcm_ibsi.h` holds
  `test_2d_glcm_hom2_ibsi()` and `test_2d_glcm_entropy_ibsi()`, added when #379 fixed the `/sum_p`
  normalisation for those two features. Neither the `GLCM_ENTROPY` nor the `GLCM_HOM2` row names
  that file in `current_test` — a `missing_current` violation that would fail CI in any of the
  twenty scanned families — and `glcm_2d_coverage.csv` recorded neither the `ibsi` token nor the
  two functions. Both rows and both artifact cells are corrected here.
- **2D first-order — four stale oracle cells.** `ENERGY` omitted `ibsi`; `INTERQUARTILE_RANGE`,
  `P10` and `P90` omitted `matlab` — in every case while the artifact's own `Test_Names` column
  already named the function that asserts it, so the two halves of one line disagreed.
- **2D GLRLM — clean**, artifact matching the tree feature for feature.

With those six cells fixed, a scan of 2D GLCM and 2D GLRLM comes back with nothing to report:
artifact identical to the tree, every claimed oracle asserted, no covering file omitted. That is
what their scanners will confirm rather than discover.

**Nothing else moves.** `features.csv`, `features.md` and `coverage_report.md` regenerate byte for
byte — `current_test` is not a report column and no status changed — and all five gates stay
clean. The three families whose dry run is not yet trustworthy (3D first-order's golden table is
written `{ "3COV"` with a space the shared table pattern does not accept; gabor asserts inside a
`.cc`) are written up in `PR/todo.md` item 59 with what each scanner will need, because a
declaration that silently reads nothing produces a confident wrong answer — the exact failure this
series exists to remove.

## 4. What review caught

Four rounds of review — six findings on the first, then eight posted as blockers across the next
three — and every one of them reproduced before it was fixed. They are kept here rather than folded
into the sections above because most are the very defect this PR exists to prevent: a check that
goes quiet, or a generated artifact that says something untrue. Two are regressions this branch
introduced, and are named as such. Where a round changed a generated artifact, the round says which
cells; rounds 3 and 4 changed none.

### Round 1 — a dropped check, and three pointer columns

A fourth commit, after review of the three above. Two of the four are the class of defect this PR
exists to prevent — a check that goes quiet, and a generated artifact that says something untrue —
so they are worth naming rather than folding into the others.

**A dropped acceptance check, in two families.** Three pre-refactor scanners ran
`no_oracle_claimed`, the reverse of `oracle_mismatch`: an oracle-suffixed test asserts a feature
while the row claims no oracle at all. Moving to declarations carried it forward for 2D radial and
silently dropped it for **2D neighbor and 2D zernike** — the other two families holding rows that
claim no oracle, which is the only place it can fire. Restored in both declarations. The
regression is reproducible: blank the `oracle` cell on those nine rows and the pre-refactor
scanners report 8 and 1 problems where the post-refactor ones reported `clean`; they now report 8
and 1 again. Nothing else covered it — `report_features.py`'s verdict for a `vetted` row with an
empty oracle cell is `agree`, because the empty cell short-circuits the mismatch branch.
`scanlib.py` had also recorded the loss as if it were the inherited state ("Only 2D radial runs it
today"); it now names all three.

**94 rows pointed at the wrong dimension's config matrix.** `pointers()` fell back from
`matrix/<family>3d.md` to `matrix/<family>.md` for a 3D family, and three 3D families have no 3D
matrix — so 3D GLCM (59 rows), 3D GLDZM (18) and 3D morphology (17) each linked a file documenting
the **2D** calculator, `matrix/morphology.md` being titled "2D morphology config matrix" outright.
The fallback is gone; an empty cell is the honest reading of *no 3D matrix exists yet*, and it is
what every other missing pointer already renders.

**235 rows could not reach their own vetting report.** The `vetting_report` pointer was keyed on the
row's *claimed* oracle, so a family whose oracle disagrees with Nyxus — which claims none on purpose
— could never link the report that recorded the disagreement. 3D GLDZM was the clearest case: this
PR adds its scanner, whose `ORACLE_SUFFIX={}` is justified entirely by
`gldzm_3d_mirp_vetting_report.md`, and all 18 of its rows rendered an empty cell. The pointer now
takes the claimed oracle first, then the oracles the tree asserts against, then — only when neither
side names one and the family x dim has exactly one report — that report. Rows with a
`vetting_report` go from 637 to 872.

**And four smaller ones.** The CI scanner loop exited on the first failing family, so a change
breaking the shared library reported one family instead of twenty; it now runs all twenty and fails
once. The family x oracle matrix in `features.md` keyed on family alone, merging 2D and 3D for the
nine family names that exist in both — 2D morphology's `matlab 33` and 3D morphology's `mirp 8` sat
in one row; it is keyed by dim now, and every column total is unchanged, which is the invariant that
says the old table mislabelled rather than miscounted. `scan_gldm3d`'s write line had lost its
feature count. The README described `audit/features.csv`, a git-ignored byproduct of the retired
pipeline, two sections below the table linking the new tracked `features.csv`; the two share a
basename and nothing else, and the README now says so.

No `audit/*_coverage.csv` changed, and outside the three pointer columns no `features.csv` value
changed — row identity and order included.

---

### Round 2 — coverage credited to things nothing asserted

Three blockers on the review of the four commits above, all reproduced before they were fixed. Two
of them are the defect this PR exists to prevent -- a rule that credits coverage nothing asserted --
and the third is the report claiming a guarantee it did not make.

**`py_loop_tables` credited every local literal whenever the function looped at all.** The gate
checked that *some* `for x in name:` existed, then read features out of *every* local list or tuple
literal in the block. A case holding `unused = ("PERIMETER",)` beside a real loop over something
else reported `PERIMETER` as covered. The literals are now resolved by name: only the one the loop
iterates is credited, plus whatever a reached literal itself names -- `cols = [c for c in df if c
in ellipse]` reaches `ellipse` -- which is the actual shape in `test_2d_ooc_invariant.py`. The seven
features that concession exists for are credited exactly as before; the artifact does not change.

**A multi-value helper credited each caller with the values it throws away.** `_fd()` in
`test_2d_morphology_fraclac.py` returns both fractal dimensions; the box-count case takes
`bc, _ = _fd(...)` and the two perimeter cases take `_, pf = _fd(...)`. All three were credited with
both features, so the artifact said the box-count case vets the perimeter dimension and the
perimeter cases vet the box-count one. Helper attribution now maps each returned position to the
features that reach it -- through the local the column lookup named -- and credits only the
positions the caller both binds and asserts on. A caller that does not unpack keeps the whole
helper, as does any feature the return positions do not account for. Two cells of
`morphology_2d_coverage.csv` change and nothing else does; both features keep the one case that
really asserts both.

**The verdict was feature-scoped while the report's key was config-scoped.** `verdict_of()` read
only the feature-wide oracle and regression buckets: it never consulted `config_recipe`,
`test_name`, or the row's own file, so a vetted row's recipe could be swapped for another of its
family's and the verdict stayed `agree`. Both halves are addressed rather than one:

- **the evidence is row-scoped where the row identifies an assertion.** `test_name` resolves through
  `test_all.cc` to the function it runs, and the verdict then asks whether *that* function asserts
  *this* feature at *this* row's kind. 307 rows carry one; all 307 resolve and agree, so this is a
  strengthening the tree already satisfies. `scan_row_tests` shows the functions it resolved to.
- **the configuration is checked where the family can say how.** `Family.recipe_reader` maps a
  recipe to the pattern its asserting function's name must match -- 3D gldm already had one for its
  own use, and 3D glszm and 3D ngtdm get one, the three families whose recipes are told apart by the
  function name rather than by file or kind. 151 rows are now config-checked, and the reviewer's
  reproduction -- `3GLSZM_GLN`'s recipe swapped for `glszm3d.regression_constant_roi` -- comes back
  `recipe-mismatch` instead of `agree`.
- **and the rest says so.** A new `verdict_scope` column reads `row+config` (151), `row` (156),
  `feature` (438) or `none` (178), per row, and `features.md` gives it a table. A recipe cannot be
  verified against a tree that records no configuration; where that is the situation, the report now
  says which rows it is true of instead of letting the key imply otherwise.

**And two smaller ones, found while fixing those.** `case_to_file` read a `TEST()` body to the first
newline-then-`}`, which runs straight past the 182 registrations written on one line into the next
multi-line case -- the case then resolves to a function it never calls. Only multi-line families read
it today, so nothing was mis-resolved; the body is brace-matched now, and the row-scoped verdict
depends on it being right. And `Family.fn_prefix` defaulted to `test_2d_<family>` for any dim that
was not `3D`, so IMQ's prefix matched nothing in the tree; a prefix that matches nothing makes the
registration sweep pass by finding no registrations at all.

#### The controls are committed this time

`audit/scanlib_selftest.py` (stdlib, no build, runs in CI beside the scanners) puts eighteen
negative controls on fixtures where the answer is known: a literal beside a real loop, a literal the
loop reaches through another, a discarded return value, a bound value that reaches no assertion, a
one-line `TEST()` registration, and five faults injected into a real registry row -- a swapped
recipe, an undeclared recipe, a `test_name` pointing at another feature's case, one naming no case
at all, and one blanked. A per-family `--check` cannot catch any of these: it compares an artifact
to the tree through the same rule that produced it.

### Round 3 — evidence that answered a weaker question

Three more blockers, all on the round-2 commit, all reproduced before the fix. Each is the same
shape as the two the round-2 commit itself closed: a rule that answers a weaker question than the
one it is written to answer, and passes.

**A row's evidence was checked for the right feature but not for the right KIND or the right TOOL.**
For a non-vetted row the covering set unioned the regression, invariant and oracle buckets, so a
regression row could be satisfied by an oracle case — the drift guard reads as present when only the
oracle test exists. Pointing 2D `NGTDM_COARSENESS`'s regression row at
`TEST_2D_NGTDM_COARSENESS_IBSI` returned `agree`. The vetted half was the same mistake through a
different door: the oracle check above it is feature-wide, so the same feature's `oracle=mirp` row
agreed while naming only the IBSI case. `row_covering()` now returns the row's own bucket — the
oracle set filtered to the tool the row claims for `vetted`, the regression set for `regression`,
the invariant set otherwise — and the two failures are told apart: `row-test-wrong-kind` and
`row-test-wrong-oracle`.

Reading the claimed oracle off a function needs `scanlib.oracle_token`, which is the name suffix per
SPEC 2 plus a family-declared exception: 3D GLDM asserts against PyRadiomics through the Python API
in `test_3d_gldm_compatibility`, whose suffix names no oracle. That map already existed inside the
family; it is declared on the `Family` now, so the report reads the same one rather than a copy —
and it is the **only** asserted function in the tree whose oracle its name cannot say, which was
measured rather than assumed.

**The feature and the recipe could be proven by different functions.** They were two existential
tests over the same list, so a row naming two individually wrong cases passed: on `3GLSZM_GLN`'s
`pyradiomics_bincount20` row, `TEST_3D_GLSZM_IBSI_GAPPED_PYRADIOMICS` asserts GLN at the wrong
recipe and `TEST_3D_GLSZM_SAE_PYRADIOMICS` runs at the right recipe asserting SAE — verdict
`agree`, `row+config`. The recipe test now ranges only over the functions already established as
this row's evidence, so one function has to answer the whole row.

**`py_loop_tables` tied a literal to a loop but not the loop to an assertion.** `scan()` establishes
only that the function asserts *somewhere*, so `checked = ("AREA",)` followed by
`for c in checked: print(c)` and an unrelated `assert 1 == 1` credited AREA. The loop body must now
either assert, or accumulate into a name the function later asserts on — which is the shape the tree
actually uses, since `test_2d_ooc_invariant.py` collects into `bad` inside the loop and asserts
`not bad` outside it. Requiring the assertion *inside* the body would have dropped the seven
features this concession exists for; requiring the body to feed one keeps them and refuses the
logging loop.

**Nothing in the report or the twenty artifacts moves.** All three are strengthenings the tree
already satisfies: 745 rows still agree, the scope split is unchanged at 151 `row+config` / 156
`row` / 438 `feature` / 178 `none`, and no `audit/*_coverage.csv` changes. The self-test grows from
18 controls to 26 — the split-evidence case, the wrong-kind and wrong-oracle cases, the logging
loop, and the accumulate-then-assert positive that guards against over-tightening it.

### Round 4 — the same two questions, one level down

Two more, both the same shape as everything above: a check that answers a weaker question than the
one it is written to answer, and passes.

**A feature-scoped verdict still accepted the wrong kind of evidence.** The row-scoped path had been
taught to read a row at its own kind in round 3, but the fallback below it — the one every row
without a `test_name` takes, 438 of them — still unioned the oracle, regression and invariant
buckets before returning `agree`. So an oracle case could stand in for a drift guard that does not
exist: the reviewer's reproduction demotes 3D morphology's MIRP `3MESH_VOLUME` row to
`status=regression` with its oracle cleared, and because the MATLAB row keeps the feature vetted and
some oracle case asserts it, the verdict stayed `agree`. A regression row is a claim that a snapshot
guards this feature; nothing checked that a snapshot existed. The fallback now reads the row's own
bucket — `regression` for a regression row, `other` for an invariant one — and returns the new
`wrong-kind` verdict when the tree covers the feature but not at the kind claimed. **Zero rows in
the registry change**: every non-vetted row in a scanned family already has its matching bucket.

**`py_loop_tables` tied the loop to an assertion but not to what the loop iterates.** Round 3
required the loop body to assert or to feed an assertion; it did not require either to mention the
loop variable. So `checked = ("AREA",)` with `for c in checked: assert frame["PERIMETER"] == ...`
credited AREA — an assertion sitting inside the loop, about something the loop never touches. The
loop variable must now reach the assertion: named on an assertion line in the body, or carried into
the accumulation the function asserts on afterwards, which is the `bad.append(..., c, ...)` under
`assert not bad` shape the real case uses. The seven features the concession exists for are
credited exactly as before and no artifact moves.

The self-test grows to 28 controls. The two new ones are written so they cannot pass vacuously: the
unrelated-assertion case asserts that `PERIMETER` — the name genuinely on the assertion line — is
still credited while the iterated literal is not, and the `3MESH_VOLUME` case pins the untouched row
as `agree` beside the demoted one as `wrong-kind`.

### And the CI step this PR added

The round-1 commit changed the scanner step from `python3 "$s" --check || exit 1` to a shell
accumulator — `rc=0`, `|| rc=1`, `exit $rc` — so that a change breaking the shared library would
report all twenty families rather than stopping at the first. That step then failed on both pushes
that carried it, with **twenty `clean` lines logged under it, nothing on stderr, and no scanner
exiting non-zero**. The run before it, with the same scanners and the `|| exit 1` form, was green.

The construct does not reproduce: the byte-identical script, run under `/usr/bin/bash -l` on Ubuntu
24.04 against this tree, exits 0, and each of the twenty scanners exits 0 individually there. So the
runner-side mechanism is not named here — what is fixed is the reliance on it.

`audit/run_scanners.py` takes the accumulation out of the shell and puts it in the process that
already knows the answer: it runs the same twenty declarations through the same `scanlib.run`, and
its exit status is an ordinary Python one, which is the shape every other check step in that
workflow already uses. It keeps the property the shell loop was written for and adds two the loop
did not have — it **names** the families that failed, and a family that raises is counted rather
than allowed to abort the other nineteen. It is also one command a developer can run locally, which
the loop was not.

Verified in both directions: 20 of 20 clean and exit 0 on the tree as it stands; with one registry
row's `current_test` pointed at a file that covers nothing, exit 1, `19 of 20 families clean`,
`FAILED: scan_zernike_coverage.py`, and the other nineteen still reported.

## New artifacts

| Path | What |
|---|---|
| `tests/vetting/audit/scanlib.py` | The shared scanner library |
| `tests/vetting/audit/scan_gldzm3d_coverage.py`, `gldzm_3d_coverage.csv` | The missing family |
| `tests/vetting/audit/gldzm_3d_golden_regen.md` | Its regeneration guide |
| `tests/vetting/report_features.py` | The report generator and its `--check` |
| `tests/vetting/features.csv`, `features.md` | The report |
| `tests/vetting/audit/scanlib_selftest.py` | The coverage rule's negative controls |
| `tests/vetting/audit/run_scanners.py` | The twenty scanners' CI entry point |

## How it was verified

Not "the tests pass" — the scanners' **behaviour** was pinned, and every claim below was measured
rather than read off the code.

- **A behavioural baseline, re-captured on this base.** Each scanner's artifact *and* its `--check`
  output were recorded under a four-mutation battery applied to the registry: blank a `test_name`,
  force a row to `vetted`, point `current_test` at a missing file, corrupt an oracle token. Byte-
  identical CSVs alone would not catch a weakened acceptance check. After the rebase onto `5aca80b9`
  the baseline was re-captured from the pristine upstream scanners on the new base rather than
  carried over. All twenty match except 2D morphology's three intended differences.
- **Every finding was reproduced before its fix and controlled after it.** Each was run against the
  pre-fix code and produced the wrong answer first;
  `audit/scanlib_selftest.py` now holds **28 negative controls** on fixtures where the answer is
  known, and runs in CI beside the scanners. Several are two-sided on purpose — the loop rule is
  pinned by a case that must stay credited as well as one that must not, so it fails on
  over-tightening as well as under-tightening.
- **The dropped check was pinned by its own reproduction.** Blanking the `oracle` cell on the nine
  2D neighbor and 2D zernike rows makes the restored check report 8 and 1 problems — the counts the
  pre-refactor scanners produce on the same injection, and the counts the post-refactor scanners
  reported as `clean` before the fix.
- **`features.csv` was diffed column by column at every step.** Row identity and order never change.
  Across the whole branch the only shared columns that moved are `scan_oracle_tests` and
  `scan_n_oracle` on the two fraclac rows; `scan_row_tests` and `verdict_scope` are new; **no
  verdict changed**, and every tightening landed on a tree that already satisfied it. Of the twenty
  coverage artifacts, only `morphology_2d_coverage.csv` moved (two cells), plus the six stale cells
  corrected in `glcm_2d_coverage.csv` and `firstorder_2d_coverage.csv`.
- **The pointer fixes were diffed row by row.** The only cells that went from set to empty are the
  94 wrong-dimension matrix links; 235 gained a `vetting_report`; no non-pointer column changed.
  Each of the six family x dim groups the single-report fallback fires on was checked against the
  report it links — including the two whose registry spelling differs from the report's
  (`gldzm_3d_mirp_vetting_report.md` tabulates all 16 as `LDHGLE`, `LDE`, … without the `3GLDZM_`
  prefix), which an exact-name search reads as a miss.
- **All five gates clean, on Windows and on Linux**: `check_coverage.py --check` clean and `--write`
  reproduces `coverage_report.md` unchanged; `check_test_names.py --check` 0 violations;
  `scanlib_selftest.py` 28/28; `run_scanners.py` 20/20 families; `report_features.py --check` clean
  at 923 rows.
- `pytest tests/python/` — **98 passed, 1 skipped, 7 failed.** The 7 are pre-existing HU and
  signed-int16 failures that reproduce identically on an untouched worktree with the same extension;
  no C++ and no python test is touched by this branch.
- **ASan is not applicable** and was not run: `src/` is byte-identical to `upstream/main`.

## Left open, deliberately

- **`scan_intensity_histogram_coverage.py` counts a bare readout as coverage** (`fvals[(int)...]`
  lines as well as assertion lines). Every other scanner refuses to, for the reason the shared
  docstring gives. Preserved so this refactor changes no behaviour, and isolated in that family's
  `collect_override` where it is visible. Removing it will drop coverage the family currently claims,
  so the affected rows need re-examining one by one — its own pass.
- **The registry itself.** A follow-up moved all 923 claims out of `oracle_coverage.csv` and into
  claim blocks in the 55 test files that assert them, deleting the registry and the 25 per-family
  artifacts. It was pushed onto this branch and then withdrawn: it is a second, larger change that
  wants its own review, it collides with #471/#472/#473 on files they edit, and this PR is complete
  without it. The gap it addresses is real and is stated in §3 — a row and the assertion it describes
  live in different files, which is how the six stale cells survived.
- **The five unscanned families**, as above.
- **3D GLDZM's missing dump test**, as above.


tests/vetting/report_features.py
```diff
@@ -0,0 +1,431 @@
...
+    Judged against the ROW's own claim, not the feature's. A feature normally carries both a vetted
+    row and a drift-guard row, and the guard is not at fault for the oracle test the other row
+    records -- they are two assertions about the same feature. Only the FEATURE-level question
+    "an oracle asserts this and no row anywhere says vetted" is a real gap.
+    """
+    key = (row["dim"], row["family"])
+    if key not in cov:
+        return "unscanned"
+    c = cov[key]
+    f, status, oracle = row["feature"], row["status"].strip(), row["oracle"].strip()
```
vjaganat90, line 262: **[Blocker] The verdict is not actually config- or assertion-specific.** The report declares `(dim, feature, oracle, config_recipe)` as its key because configurations can disagree by orders of magnitude, but `verdict_of()` only consults feature-wide oracle/regression buckets. It never uses `config_recipe`, `claim_test_name`, or the row's claimed file to select the corresponding assertion. I reproduced this with the vetted `3GLSZM_GLN` row: changing its recipe from `glszm3d.pyradiomics_bincount20` to either `glszm3d.regression_constant_roi` or `glszm3d.regression_ut_phantom_nobinning` still returns `agree`. The generated rows likewise receive the same complete list of oracle and regression tests regardless of which configuration they claim. Please make the scanned evidence row-scoped—ideally by resolving a general assertion identifier to its function and configuration—or narrow the report's contract to feature-level coverage. As written, the central config-aware guarantee is not enforced.

tests/vetting/audit/scanlib.py
```diff
@@ -0,0 +1,550 @@
...
+        if fam.featureset_loop and fam.featureset_loop.search(block):
+            hits.setdefault(fn, set()).update(all_names)
+        # A table the function builds and then loops over while asserting -- the equivalence tests
+        # put their feature pairs in `std::vector<Pair> pairs = { ... };` and assert inside the
+        # loop, so the names never appear on an assertion line. Credited only when the function
+        # both asserts and range-loops, so a plain lookup table is not mistaken for coverage.
+        if fam.loop_tables and re.search(r"for\s*\([^)]*:\s*\w+\s*\)", block):
+            for m in re.finditer(r"=\s*\{(.*?)\};", block, re.S):
+                hits.setdefault(fn, set()).update(feat_re.findall(m.group(1)))
+        for name, feats in helpers.items():
```
vjaganat90: **[Blocker] Helper propagation still converts unasserted values into coverage.** Every feature read by a helper is credited to every test that calls that helper, even when the caller discards part of the result. There is a current example in `test_2d_morphology_fraclac.py`: `_fd()` returns both `FRACT_DIM_BOXCOUNT` and `FRACT_DIM_PERIMETER`; the box-count test does `bc, _ = _fd(...)`, while the perimeter tests do `_, pf = _fd(...)`. Nevertheless, the generated artifact credits all three tests with both features. That contradicts the stated rule that coverage requires an assertion. Please make helper attribution aware of which returned value reaches the assertion, or provide an explicit family-level mapping for multi-value helpers. A negative test for “helper returns two features, caller asserts only one” would pin the intended behavior.

tests/vetting/audit/scanlib.py
```diff
@@ -0,0 +1,550 @@
...
+def py_literal_features(block, feat_re):
+    """Feature names in the local sequence literals of a case that loops one of them while asserting.
+
+    A pytest case can name every feature it compares in a list it builds and then iterates -- the
+    ellipse tuple in test_2d_ooc_invariant.py is the case -- so no name ever reaches an assertion
+    line and the shared rule sees nothing. This is the same concession loop_tables makes for the C++
+    equivalence tests, and it is gated the same way: the function has to both assert and range-loop,
+    so a plain lookup list is not mistaken for coverage.
+    """
+    if not PY_LOOP_NAME.search(block):
```
vjaganat90: **[Blocker] `py_loop_tables` credits every local literal when any loop exists.** This check only establishes that the function contains some `for … in <name>` loop; the code then extracts features from every local list or tuple literal without checking whether that literal feeds the loop or its assertion. For example, a function containing `unused = ("PERIMETER",)`, `checked = ("AREA",)`, and a loop over `checked` reports both features as covered—I reproduced that directly through `py_literal_features()`. This reintroduces the false-positive coverage this refactor is meant to remove. Please associate each assignment's left-hand name with the collection that is actually iterated—following the current `ellipse → cols` flow where needed—and add an unused-literal-beside-a-real-loop negative control.

vjaganat90 (commented):
Overall looks good but these need a tweak.

tests/vetting/report_features.py
```diff
@@ -0,0 +1,558 @@
...
+            return "no-assertion-at-all", scope
+
+    if scope == "feature":
+        return "agree", scope
+
+    if unresolved:
+        return "test-name-unresolved", scope
+    # the row's own assertion, at the row's own kind: a vetted row answers to an oracle-suffixed
+    # function, and a feature-wide `asserted` set does not say that THIS case is one of them
+    covering = c.asserted.get(f, set()) if status == "vetted" else (
```
vjaganat90: **[Blocker] The row check still accepts the wrong evidence kind and oracle.** For every non-vetted row, `covering` unions regression, invariant/other, and oracle functions, so a regression claim can be satisfied by an oracle case. I reproduced this by changing the `NGTDM_COARSENESS` regression row's `test_name` to `TEST_NYXUS.TEST_2D_NGTDM_COARSENESS_IBSI`; it still returns `('agree', 'row')`. The vetted analogue also passes: the same feature's `oracle=mirp` row still agrees when `test_name` names only the IBSI case, because the oracle check above is feature-wide and is not tied to the resolved function. Once a row supplies `test_name`, please require its resolved evidence to match the exact claim—claimed oracle for `vetted`, regression bucket for `regression`, and the corresponding other bucket for invariant—and add both mutations as negative controls.

tests/vetting/report_features.py
```diff
@@ -0,0 +1,558 @@
...
+        return "row-test-lacks-feature", scope
+
+    # the configuration, where the family declares how to read one off a function name
+    readers = s.fam.recipe_reader
+    if not readers:
+        return "agree", scope
+    recipe = row["config_recipe"].strip()
+    if recipe not in readers:
+        return "recipe-unreadable", "row+config"
+    if not any(readers[recipe].match(fn) for fn in fns):
```
vjaganat90: **[Blocker] Feature and recipe can be proven by different functions.** The feature check and recipe check are separate existential tests over `fns`, so two individually wrong cases can combine into `row+config`. On the `3GLSZM_GLN` `pyradiomics_bincount20` row, I set `test_name` to `TEST_NYXUS.TEST_3D_GLSZM_IBSI_GAPPED_PYRADIOMICS;TEST_NYXUS.TEST_3D_GLSZM_SAE_PYRADIOMICS`: the first asserts GLN at the wrong recipe, the second matches the requested recipe but asserts SAE, yet the verdict is still `('agree', 'row+config')`. Please require one resolved function to be both in the row's exact covering set and matched by the recipe reader, and pin this split-evidence case in the self-test.

tests/vetting/audit/scanlib.py
```diff
@@ -0,0 +1,713 @@
...
+    but never iterates is not coverage, and reading every local literal because the function happens
+    to loop somewhere would credit it anyway -- which is the false positive this rule exists to
+    refuse. The reach is transitive because the case builds one list from another
+    (`cols = [c for c in df if c in ellipse]`, then `for c in cols:`), so a literal named inside a
+    reached literal is reached too.
+    """
+    literals = {}
+    for m in PY_ASSIGN_LITERAL.finditer(block):
+        literals.setdefault(m.group(1), []).append(_bracketed(block, m.end() - 1))
+    reached = {n for n in PY_LOOP_NAME.findall(block) if n in literals}
```
vjaganat90: **[Blocker] `py_loop_tables` now ties a literal to a loop, but still does not tie that loop to an assertion.** `scan()` only establishes that the function contains some assertion somewhere. For example, `checked = ("AREA",)` followed by `for c in checked: print(c)` and then an unrelated `assert 1 == 1` still reports AREA as covered. That conflicts with the documented rule that the function loops the table *while asserting* and still lets a lookup/logging loop become coverage. Please require the relevant loop body to contain the assertion (and ideally use its iterator), then add a loop-plus-unrelated-assert negative control.

tests/vetting/report_features.py
darkclad: **Fixed** — the verdict is assertion-specific where the row identifies an assertion, config-specific where the family can say what a config looks like, and the rest now says so per row instead of implying otherwise.

- `test_name` resolves through `test_all.cc` to the function it runs, and the verdict asks whether *that* function is the row's evidence. 307 rows carry one; all 307 resolve and agree, so this is a strengthening the tree already satisfied.
- `Family.recipe_reader` maps a recipe to the pattern its asserting function's name must match. 3D GLDM had one; 3D GLSZM and 3D NGTDM gain one — the three families whose recipes are distinguishable by function name rather than by file or kind. 151 rows are config-checked, and your reproduction (`3GLSZM_GLN`'s recipe swapped for `glszm3d.regression_constant_roi`) now returns `recipe-mismatch`.
- A new `verdict_scope` column reads `row+config` (151) / `row` (156) / `feature` (438) / `none` (178). The 438 carry no `test_name` at all, so nothing identifies the assertion they describe — the honest reading is feature-level, and the column says that rather than letting the key imply more. `PR/todo.md` item 58 tracks closing them per family.
- `scan_row_tests` shows the functions the row's own `test_name` resolved to.

Controls are in `audit/scanlib_selftest.py` (runs in CI): swapped recipe, undeclared recipe, `test_name` pointing at another feature's case, `test_name` naming no case, and a blanked `test_name` falling back to `feature`.

tests/vetting/audit/scanlib.py
darkclad: **Fixed** — helper attribution is now per returned position, and a caller is credited only with the positions it both binds and asserts on.

`helper_positions()` maps each element of a helper's tuple return to the features that reach it, following the local the column lookup named (`bc = [c for c in row.index if c.endswith("FRACT_DIM_BOXCOUNT")][0]`, then `return float(row[bc]), float(row[pf])`). `helper_credit()` reads the caller's unpacking against that: `_` is dropped, and so is a bound name that never reaches an assertion line. A caller that does not unpack keeps the whole helper, as does any feature the return positions do not account for.

Your example is the measured result — `morphology_2d_coverage.csv` changes in exactly two cells:

- `FRACT_DIM_BOXCOUNT` loses `test_2d_morphology_perimeter_disk_analytic` and `..._koch_snowflake_analytic`
- `FRACT_DIM_PERIMETER` loses `test_2d_morphology_boxcount_known_dimension_analytic`

Both keep the gtest case `test_2d_morphology_fractal_dimension_blob512_fraclac`, which asserts both on its own assertion lines and does not go through `_fd`.

The negative test you asked for is in `audit/scanlib_selftest.py`: a helper returning two features with one caller per half, a caller that does not unpack (keeps both), a single-value helper (unchanged), and a caller that *binds* the second value but never asserts on it (still not credited).

tests/vetting/audit/scanlib.py
darkclad: **Fixed** — literals are resolved by name against the loop, so a literal the loop never reaches is not credited.

`PY_ASSIGN_LITERAL` now captures the assigned name and `PY_LOOP_NAME` the iterated one; only the iterated literal is credited, plus any literal a reached literal names. That second hop is the `ellipse → cols` flow you pointed at: `cols = [c for c in df_ram.columns if ... or c in ellipse]`, then `for c in cols:`.

Your exact case is a control in `audit/scanlib_selftest.py` — `unused = ("PERIMETER",)` beside `checked = ("AREA",)` and a loop over `checked` credits AREA only. The seven features the concession exists for in `test_2d_ooc_invariant.py` are credited exactly as before, so no artifact moves.

Your follow-up (loop not tied to an assertion) is answered on the round-2 comment below and is fixed too.

tests/vetting/report_features.py
darkclad: **Fixed** — a row's evidence is now read at its own kind and its own oracle, and the two ways of missing are told apart.

`row_covering()` returns the row's own bucket rather than a union: the oracle set filtered to the tool the row claims for `vetted`, the regression set for `regression`, the invariant set otherwise. Both of your mutations fail now:

- 2D `NGTDM_COARSENESS` regression row → `TEST_2D_NGTDM_COARSENESS_IBSI` gives `row-test-wrong-kind`
- the same feature's `oracle=mirp` row naming only the IBSI case gives `row-test-wrong-oracle` (naming the MIRP case still agrees)

Filtering by the claimed oracle needs a function → oracle map, which is `scanlib.oracle_token`: the SPEC 2 name suffix, plus a `Family.fn_oracle` exception for a function whose suffix names no oracle. I measured that rather than assuming it — `test_3d_gldm_compatibility` (3D GLDM's PyRadiomics case through the Python API) is the only asserted function in the tree in that position, and the map it already used privately is declared on the `Family` now so the report reads the same one.

Both mutations are pinned in `audit/scanlib_selftest.py`, each beside the unmutated row. No row in the registry fails the stricter rule: 745 still agree and no artifact changes.

tests/vetting/report_features.py
darkclad: **Fixed** — one resolved function has to answer the whole row.

The two tests are no longer independent: the recipe reader now ranges only over `hits`, the functions already established as this row's evidence by `row_covering()`, so a case that matches the recipe but asserts a different feature can no longer supply the half the other case is missing.

Your split-evidence case is the control, in `audit/scanlib_selftest.py`: `3GLSZM_GLN`'s `pyradiomics_bincount20` row with `test_name` set to `TEST_3D_GLSZM_IBSI_GAPPED_PYRADIOMICS;TEST_3D_GLSZM_SAE_PYRADIOMICS` now returns `('recipe-mismatch', 'row+config')` instead of `agree`.

tests/vetting/audit/scanlib.py
darkclad: **Fixed** — the loop body has to do the comparing, not merely exist.

A `for x in NAME:` credits NAME only if its body either asserts, or accumulates into a name the function later asserts on. Your example — `for c in checked: print(c)` beside an unrelated `assert 1 == 1` — credits nothing now.

One note on why it is not "the assertion must be inside the loop body": the case this concession exists for does not do that. `test_2d_ooc_invariant.py` collects into `bad` inside the loop and asserts `not bad` **outside** it, which is the ordinary accumulate-then-assert shape. Requiring the assertion inside the body would have dropped the seven features the rule was written to credit, so the rule follows the value instead: the body has to feed something the function asserts.

Both directions are controls in `audit/scanlib_selftest.py` — the logging loop credits nothing, and the accumulate-then-assert loop credits exactly the two features its list names — so the rule cannot be tightened into dropping the real case or loosened back into accepting the logging one.

tests/vetting/report_features.py
```diff
@@ -0,0 +1,598 @@
...
+    if status == "vetted":
+        if not c.asserted.get(f):
+            return "claim-without-assertion", scope
+        if oracle and oracle not in c.oracles.get(f, set()):
+            return "oracle-mismatch", scope
+    else:
+        # a regression or invariant row claims a drift guard, so that is what it is checked against
+        if c.oracles.get(f) and (row["dim"], f) not in vetted_features:
+            return "assertion-without-claim", scope
+        if not (c.regression.get(f) or c.other.get(f) or c.asserted.get(f)):
```
vjaganat90, line 274: **[Blocker] Feature-scope rows still accept the wrong evidence kind.** This union lets an oracle assertion satisfy a regression or invariant claim before line 278 returns `agree`. Concrete reproduction: change the MIRP `3MESH_VOLUME` row to `status=regression` and clear its oracle; the remaining MATLAB row keeps the feature vetted, no regression assertion exists, yet the result is still `agree` at `feature` scope. Even without `test_name`, require the matching feature-wide bucket (`c.regression` for regression, `c.other` for invariant), and pin this mutation as a negative control.

tests/vetting/audit/scanlib.py
```diff
@@ -0,0 +1,766 @@
...
+    `bad.append(...)` / `assert not bad` shape these cases actually use, where the assertion is
+    outside the loop by construction.
+    """
+    literals = {}
+    for m in PY_ASSIGN_LITERAL.finditer(block):
+        literals.setdefault(m.group(1), []).append(_bracketed(block, m.end() - 1))
+    asserted = asserted_names(block)
+    reached = {name for name, body in loop_bodies(block)
+               if name in literals
+               and (ASSERTION.search(body) or set(PY_ACCUMULATE.findall(body)) & asserted)}
```
vjaganat90: **[Blocker] Loop-table evidence is still not tied to the loop iterator.** Any assertion inside the loop body currently credits the iterated literal, even when that assertion never uses the iterator. For example, `checked = ("AREA",); for c in checked: assert frame["PERIMETER"] == other["PERIMETER"]` credits both `AREA` and `PERIMETER`. Please require the iterator to reach the assertion - or the accumulation later asserted - and add this unrelated-in-loop assertion as a negative control.

vjaganat90 (changes requested):
The six earlier reproductions are addressed - thank you. Two false-positive coverage paths remain and should be fixed before merge. Please also refresh the PR body to match the current diff: it now has 32 files, 7 commits, and a net -380 lines, while the opening still says 30 files, 4 commits, and -1,133 lines; update the related scanner-size/count narrative as appropriate.

vjaganat90 (approved):
After addressing the two remaining issues. This is good to go.

tests/vetting/report_features.py
darkclad: **Fixed** — the fallback now reads the row's own bucket, not the union.

Your reproduction is exact: with `3MESH_VOLUME`'s MIRP row demoted to `status=regression` and its oracle cleared, the MATLAB row keeps the feature vetted so `assertion-without-claim` stays quiet, and the union kept `no-assertion-at-all` quiet too. A regression row is a claim that a snapshot guards the feature, and nothing checked that one existed.

`verdict_of` now requires the matching feature-wide bucket after the "nothing at all" test — `c.regression` for a regression row, `c.other` for an invariant one — and returns a new `wrong-kind` verdict when the tree covers the feature but not at the kind claimed. That is the feature-scoped twin of the `row-test-wrong-kind` case added last round, so both paths now read a row at its own kind.

**Zero rows change**: every non-vetted row in a scanned family already has its matching bucket, so this is a strengthening the tree already satisfies. Your mutation is pinned in `audit/scanlib_selftest.py`, alongside the untouched row asserted as `agree` — so the control cannot pass by the check having gone quiet.

tests/vetting/audit/scanlib.py
darkclad: **Fixed** — the loop variable now has to reach the assertion.

You are right that the previous round stopped one step short: it required the body to assert or to feed an assertion, but neither had to mention what the loop iterates, so an assertion about a different column sitting inside the loop credited the list.

`iterator_reaches_assertion()` accepts exactly two shapes: the loop variable named on an assertion line inside the body, or carried into an accumulation whose target the function asserts on afterwards — `bad.append((lab, c, p, q))` under `assert not bad`, which is what `test_2d_ooc_invariant.py` actually does, with the assertion outside the loop by construction. Your `for c in checked: assert frame["PERIMETER"] == other["PERIMETER"]` credits nothing from the literal.

The control is in `audit/scanlib_selftest.py` and is deliberately not a bare "credits nothing": it asserts that `PERIMETER` — the name genuinely on the assertion line — is still credited by the ordinary line rule while `AREA` and the rest of the iterated literal are not, so it fails if the fix over-tightens as well as if it under-tightens. The seven features this concession exists for are credited exactly as before and no artifact moves.

PR #471, opened 2026-09-08
fix(morphology): report ROI radii as distances, and vet the max and median against scikit-image

darkclad:
## What the features compute

`RoiRadiusFeature` takes every ROI pixel, measures its distance to the nearest pixel of the ROI's own
contour, and reports the mean, maximum and median of those distances — statistics of the ROI's
inradius map. That distance is taken exactly, by one of two engines: an exact separable Euclidean
distance transform over the ROI's bounding box, or a full linear scan of the contour per pixel
(`Pixel2::exact_min_sqdist()`), rather than the approximate hill-descent `min_sqdist()` beside it.
The two engines return the same integers and the choice between them is a cost estimate; the
out-of-core path always takes the scan. The median is taken over the distances directly rather than
through `TrivialHistogram`, whose item type is `unsigned int`.

**Three feature values change**, since these are now distances rather than squared distances. On the
8×8 `shape2d` fixture:

| feature               | value                                   |
| --------------------- | --------------------------------------- |
| `ROI_RADIUS_MEAN`   | 1.07692307692308 → 0.62471694950458789 |
| `ROI_RADIUS_MAX`    | 4 → 1.4142135623730951                 |
| `ROI_RADIUS_MEDIAN` | 1 → 1                                  |

Anyone comparing against stored Nyxus output sees the first two move; the `fix(` prefix puts that in
the changelog.

## The approximate minimum, measured against the exact one

`min_sqdist()` is a coarse-to-fine hill descent that assumes a locally-unimodal ordered contour; its
own header says it "can settle in the wrong basin on a closed contour, OVERESTIMATING the true
minimum". How much it overestimates was never measured. Per ROI pixel, `sqrt(min_sqdist(K))` against
`sqrt(exact_min_sqdist(K))` on the same contour:

| fixture     |  pixels | contour | pixels overestimated | worst absolute | worst relative | on-contour pixels given a non-zero distance |
| ----------- | ------: | ------: | -------------------: | -------------: | -------------: | ------------------------------------------: |
| `shape2d` |      26 |      18 |            5 (19.2%) |          1.414 |         +41.4% |                                     4 of 11 |
| disk R=10   |     317 |      56 |             7 (2.2%) |          1.414 |        +123.6% |                                     1 of 33 |
| disk R=20   |   1 257 |     112 |          146 (11.6%) |         10.440 |         +1040% |                                    10 of 67 |
| disk R=40   |   5 025 |     224 |          748 (14.9%) |          9.220 |          +849% |                                    9 of 135 |
| disk R=100  |  31 417 |     564 |        6 271 (20.0%) |         15.133 |         +1403% |                                   15 of 340 |
| disk R=200  | 125 629 |   1 128 |       29 012 (23.1%) |          4.325 |          +300% |                                    3 of 681 |

The share of wrong pixels **grows with the ROI**, from 2% to 23%, and a single pixel can be off by
14×. The last column is the sharpest form of it: a pixel that lies *on* the contour has distance 0 by
definition, and the descent hands back a non-zero number for some of them.

Carried into the published statistics:

| fixture     | `ROI_RADIUS_MEAN` exact → approx     | `ROI_RADIUS_MAX`                      | `ROI_RADIUS_MEDIAN`  |
| ----------- | --------------------------------------- | --------------------------------------- | ---------------------- |
| `shape2d` | 0.624717 → 0.864819 (**+38.4%**) | 1.414214 → 2.000000 (**+41.4%**) | 1.0 → 1.0             |
| disk R=10   | 2.833411 → 2.849871 (+0.58%)           | unchanged                               | unchanged              |
| disk R=20   | 6.087110 → 6.397909 (**+5.1%**)  | unchanged                               | 5.0 → 5.830952        |
| disk R=40   | 12.705792 → 12.795812 (+0.71%)         | unchanged                               | 11.045361 → 11.180340 |
| disk R=100  | 32.631372 → 32.709646 (+0.24%)         | unchanged                               | 28.600699 → 28.635642 |
| disk R=200  | 65.914354 → 65.954533 (+0.06%)         | 199.002513 → 199.077874 (+0.038%)      | 57.801384 → 57.939624 |

Worth noting which assertion catches which: `MAX` on a disk is insensitive to the approximation
until R = 200, so the closed-form oracle alone would not see it — the skimage `MEDIAN` pins at
R = 20 and R = 40 are what fail (5.0 against 5.830952). The two oracles are not redundant.

## The cost of the exact scan, and the engine that removes it

`exact_min_sqdist()` is O(contour) per pixel against `min_sqdist()`'s O(log contour). Timed on filled
disks over the same pixels and the same contour:

|  pixels | contour |     exact | approximate | ratio | error removed (`MEAN`) |
| ------: | ------: | --------: | ----------: | ----: | -----------------------: |
|   7 845 |     280 |   1.41 ms |     1.25 ms | 1.1× |                       — |
|  31 417 |     564 |  12.11 ms |     5.24 ms | 2.3× |                    0.24% |
| 125 629 |   1 128 |  96.90 ms |    20.55 ms | 4.7× |                    0.06% |
| 502 625 |   2 260 | 762.40 ms |    92.90 ms | 8.2× |                       — |

The ratio grows with the contour, so the large single-ROI case pays. That cost is what the second
engine removes.

The quantity is a **distance transform seeded at the contour**, and the exact separable form
(Felzenszwalb & Huttenlocher, two passes) costs O(bounding box) and is independent of the contour's
length. Measured like-for-like against the scan in the same optimised build, on the same pixels and
the same contour:

| ROI pixels | contour | bbox cells |       scan | transform |         speed-up |
| ---------: | ------: | ---------: | ---------: | --------: | ---------------: |
|        317 |      56 |        441 |   0.012 ms |  0.003 ms |            3.7× |
|      1 257 |     112 |      1 681 |   0.094 ms |  0.011 ms |            8.3× |
|      5 025 |     224 |      6 561 |   0.722 ms |  0.057 ms |           12.7× |
|     31 417 |     564 |     40 401 |  11.511 ms |  0.358 ms |           32.2× |
|    125 629 |   1 128 |    160 801 |  93.422 ms |  2.317 ms |           40.3× |
|    502 625 |   2 260 |    641 601 | 760.304 ms |  7.901 ms | **96.2×** |

So the exact answer on the largest fixture now costs **7.9 ms against the 92.9 ms the approximation
cost** — cheaper than the method it replaced, which removes the accuracy-for-speed trade rather than
re-balancing it.

**It is exact, and no value moves.** Sites lie on the integer lattice and both engines work in
integer squared distances, so a cell holds the same integer the scan returns and the two agree bit
for bit — verified pixel by pixel over 21 shapes (below). Nothing in this PR's oracle or regression
expectations is re-pinned by it. The transform keeps the lower envelope's breakpoint comparisons in
integer arithmetic, cross-multiplying rationals rather than dividing, so no rounding can place a
breakpoint on the wrong side; `int64_t` covers any raster up to 2^20 cells per side.

**The engine is chosen, not assumed.** The transform loses on sparse ROIs: a one-pixel-wide diagonal
carries a bounding box of P² cells for P pixels and a contour of only P, and is **7.9× slower**
there. Fill ratio alone does not separate that case — a thin ring is nearly as sparse (1.2% against
0.14%) and the transform still wins 4.9×, because its contour is long and that is what the scan pays
for. `RoiRadiusFeature::edt_is_cheaper` therefore estimates both costs from three counts known
before any distance is computed — bounding-box area, pixel count, contour length — and a 16 MB
ceiling bounds the raster a worker thread may allocate. Since both engines are exact, the estimate
decides only how long a ROI takes, never what it reports.

The raster spans the union of the ROI's pixels and its contour, because a contour traced on a padded
image does not lie inside the ROI's own bounding box (see the `MEAN` disposition below). That union
is measured from the pixels rather than read from `LR::aabb`, which is accumulated during scanning
and would silently disagree with a cloud filled by another route.

## Why the out-of-core path keeps the scan

`osized_calculate()` is unchanged and now records the reason rather than asserting an equality.

A ROI reaches that path by its bounding box raster in `Pixel2` exceeding the RAM limit, so a
transform's raster of that same box is on the order of the budget that sent it there — one third of
it on Linux, two thirds on Windows, growing linearly past the threshold. That was the expected
objection; measurement gave a stronger one.

On a 8000×8000 ROI forced out-of-core on a 4 GB Linux box, with the page cache evicted before each
run so the reads reach the device:

| build                                 |   wall | peak RSS | values    |
| ------------------------------------- | -----: | -------: | --------- |
| without the transform                 | 5:22.5 |  1038 MB | identical |
| transform in-RAM only (this PR)       | 5:15.7 |  1038 MB | identical |
| transform out-of-core, ceiling lifted | 5:19.2 |  1038 MB | identical |

Within noise, and instrumentation showed the transform declining to run even with the ceiling
removed. The run is 144 s user and 168 s system with **zero major page faults**: it is bound by
per-pixel stdio against the disk-backed pixel cloud, not by distance arithmetic. Against that, the
distance work is under 1% of the runtime — so there is nothing there to win, and the raster question
never arises.

## What is now vetted

`ROI_RADIUS_MAX` and `ROI_RADIUS_MEDIAN` carry oracle rows on a new recipe `morphology.radius_disks`
and benchmark `bench_radius_disks` (filled disks at R = 10, 20, 40). **Morphology 76 → 78 features
vetted; the tree 605 → 607.**

The reference is `skimage.segmentation.find_boundaries(connectivity=1, mode='inner')` — the
foreground pixels 4-adjacent to background — plus a minimum over the pixels it returns, agreeing to
double precision on all three disks. Three points, since the registry recorded these as "not cleanly
external-vettable":

- **The boundary is the only convention in the quantity**, and it is skimage's, not a
  re-implementation of the Nyxus tracer. It is CellProfiler's own edge definition, which is why
  `gen_morphology_{,wholeslide_}cellprofiler.py` already reproduce the `EDGE_*` statistics through
  it; on the r=20 disk it returns 112 boundary pixels, the count `benchmarks.md` records for
  `bench_disk64_diagonal_boundary`.
- **The distance step carries no convention**, and the generator computes it twice — an exhaustive
  minimum and `scipy.ndimage.distance_transform_edt` over the complement — requiring the two to agree.
  That equivalence is also what licenses the in-tree transform: the generator has been asserting for
  the whole of this PR that a distance transform and an exhaustive minimum give the same radii.
- **Disks and not the 8×8 raster**, for the reason `morphology.perimeter_circles` exists: on 26 pixels
  with a hole the two boundary conventions have nothing to converge to.

Three radii rather than one because a single disk cannot distinguish a distance from its square —
either is just a number on one fixture, while across a doubling of R a distance grows 2.1× and a
square 4.2×.

### A second oracle, for what the pins cannot see

The pins say Nyxus agrees with a reference implementation of the definition; they cannot say the
value obeys the geometry, since a reference sharing a mistake would agree with it. So the closed form
is a separate assertion with a separate oracle (`test_2d_morphology_analytic.h`), and it is **exact**.
On a disk the centre attains the maximum, so the value is the centre's distance to the *nearest*
boundary pixel — not the axial `(R, 0)` but the one at offset `(1, R−1)`: inside (`1+(R−1)² ≤ R²`),
boundary (its neighbour `(1, R)` is not inside), and nearer than `(R, 0)` (`R²−2R+2 < R²`). So

> **`ROI_RADIUS_MAX = √((R−1)² + 1)`**, matched to 1e-12 relative.

| R  | `ROI_RADIUS_MAX` | √((R−1)²+1)     | MAX/(R−1) |
| -- | ------------------ | ------------------ | ---------- |
| 10 | 9.0553851381374173 | 9.055385138137417  | 1.006154   |
| 20 | 19.026297590440446 | 19.026297590440446 | 1.001384   |
| 40 | 39.012818406262319 | 39.01281840626232  | 1.000329   |

Confirmed nearest at R = 5, 10, 20, 40, 80 and 160, so the form is not fitted to the three radii the
test uses. It also derives the convergence rather than fitting it: `MAX/(R−1) = √(1+1/(R−1)²) → 1`,
i.e. linear growth in R. This gives `ROI_RADIUS_MAX` the ≥2-oracle redundancy SPEC §3.1 tracks, and
it lives outside the `_skimage` function because the coverage scanner credits oracle tokens by
function suffix — an `analytic` claim asserted there would credit scikit-image with a claim it is not
making.

**What the assertions discriminate.** Against an implementation returning the squared distance, both
tests fail: `ROI_RADIUS_MAX` reads 82 against a pinned 9.055, `MEDIAN` 5 against 2.236, and
`MAX/(R−1)` reads 9.1, 19.1, 39.0 — growing where the pinned ratios converge.

## The two engines are pinned to the same values

Twelve assertions, in the invariant and mechanics tiers, over
`tests/test_2d_morphology_invariant.h` and `tests/test_2d_morphology_mechanics.h`.

The **invariant** tier states that the two engines are the same measurement, which is the property
that lets one be chosen on a cost estimate without the choice being observable. Over a battery of 21
shapes — disks R = 1…40, a diagonal, two rings, a square with a hole, an L and eight multi-lobe
blobs — every ROI pixel's squared distance is the same integer from both, the three reported
statistics are bit-identical, and the in-RAM and out-of-core paths agree. Exact equality, not a
tolerance: a transform agreeing only to within a tolerance would not be exact and the claim would be
false.

Contours are supplied by the tests rather than traced, so a defect in contour construction moves both
sides equally and cannot mask a disagreement between the engines. End-to-end path equality with a
*traced* contour is a separate and stronger claim, blocked by the out-of-core contour divergence
described below; the header comment records that, so the two cannot drift apart silently.

The **mechanics** tier covers degenerate and out-of-range inputs, each asserting a defined outcome
rather than the absence of a crash — a routine returning a plausible number for a ROI with no contour
would pass a no-crash test and still be wrong. It covers no sites at all, sites lying entirely
outside the raster, zero-width and zero-height rasters, a single cell, a single site checked against
arithmetic rather than against the scan, sites on all four corners, an empty ROI, a ROI whose contour
was never built, a one-pixel ROI, and a contour lying outside the ROI's bounding box.

**Verified to fail when they should.** A deliberate off-by-one in the transform's final write fails
five of the twelve, including the two that check against arithmetic rather than against the scan — so
a defect moving both engines together is still caught. `..._paths_agree_invariant` additionally
counts how many shapes route through the transform and fails if none do, so neither a drifting
battery nor a change to the estimate's constants can quietly reduce it to comparing the scan with
itself.

## The docs now describe this quantity

`docs/source/Math/f_morphology.rst` defined these features over "`OC_i` — segment connecting centroid
to an edge pixel", and `featurelist.rst` repeated it as "centroid to edge distance". That text
describes a **different feature, which already exists**:
`calculate_inscribing_circumscribing_circle` (`circle.cpp:220`) takes exactly that distance and
publishes twice its minimum and maximum as the two circle diameters. On `shape2d`:

|                 | from the docs' formula | pinned in the tree                                  | difference |
| --------------- | ---------------------- | --------------------------------------------------- | ---------- |
| 2 × max ‖OC‖ | 12.331707339908755     | `DIAMETER_CIRCUMSCRIBING_CIRCLE` 12.3317073399088 | 0          |
| 2 × min ‖OC‖ | 0.8284868934052972     | `DIAMETER_INSCRIBING_CIRCLE` 0.828486893405308    | 0          |

Defining `ROI_RADIUS_MAX` that way would have made it `DIAMETER_CIRCUMSCRIBING_CIRCLE / 2`. Both
pages now define the inradius statistics the code measures and point at the circle diameters for the
centroid-to-edge extremes. **No feature value moves from this**; it is the documentation catching up
with what the oracle rows are scoped to.

## The three cells that are not vetted, and why

`matrix/morphology.md` now carries a `RoiRadiusFeature` section — the file went from two producers to
three — enumerating four cells. Three are dispositions rather than claims:

**`ROI_RADIUS_MEAN` on any fixture.** `ContourFeature::buildRegularContour` traces in an image padded
by one pixel on every side and restores only the bounding-box origin, so every contour pixel is
reported one pixel right and one down of where it is. Shifting the *reference* boundary by (+1, +1)
reproduces every Nyxus value exactly, which identifies that offset as the whole of the difference; on
the 8×8 mask the traced contour **is** the skimage inner boundary shifted by (+1, +1), all 18 pixels,
compared as sets. On a disk `MAX` and `MEDIAN` are unmoved by the shift because the pixel attaining
them moves with it, so those two promote and `MEAN` does not.

| fixture              | reference          | reference shifted (+1,+1) = Nyxus | difference |
| -------------------- | ------------------ | --------------------------------- | ---------- |
| disk R=10`MEAN`    | 2.7517095857010947 | 2.8334114624727413                | +2.97%     |
| disk R=20`MEAN`    | 6.047026418093844  | 6.087109772358641                 | +0.66%     |
| `shape2d` `MEAN` | 0.3076923076923077 | 0.6247169495045879                | +103%      |

That defect is in `contour.cpp` and reaches `radial_distribution.cpp`, `circle.cpp`, the 62 weighted
moments and the GPU rebasing too, so it is not this PR's to fix — and the generator prints the
comparison on every run, so `MEAN` promotes with a registry edit once it is. It is also why the
transform's raster spans the union of pixels and contour rather than the pixels alone.

**Out-of-core — `impl-defect`.** Both paths take the exact minimum over the same ROI pixels, so the
only differing input is the contour, and it differs: on `bench_disk64_diagonal_boundary` both
builders return **112** pixels and not the same 112.

|                       | in-RAM             | out-of-core        | difference |
| --------------------- | ------------------ | ------------------ | ---------- |
| `ROI_RADIUS_MEAN`   | 6.087109772358636  | 7.169174091182726  | +17.8%     |
| `ROI_RADIUS_MEDIAN` | 5.0                | 6.708203932499369  | +34.2%     |
| `ROI_RADIUS_MAX`    | 19.026297590440446 | 19.026297590440446 | 0          |

No translation of the inner boundary in ±4 reproduces it, so it is pinned as a characterization
rather than explained as an offset, and guarded by a test asserting both sides, the inequality and
`MAX`'s equality. It was invisible because `test_2d_ooc_invariant.py` exercises these features only on
a full-image rectangle, where the two builders coincide — the same blind spot that hid the
`PERIMETER` divergence.

**Whole-slide — `VALID-BUT-PRODUCTION-ONLY`.** `reduce_trivial_2d_wholeslide()` calls
`ContourFeature::extract()` whenever ROI radius is requested and then `RoiRadiusFeature::extract()`,
so this is production. At `SINGLEROI=true` the contour is `buildWholeSlideContour()`'s four AABB
corners, and the inradius map becomes a distance-to-nearest-corner map over the whole frame:

|                       | whole-slide (4096 px) | segmented (1257 px) | ratio |
| --------------------- | --------------------- | ------------------- | ----- |
| `ROI_RADIUS_MEAN`   | 24.490797504013532    | 6.087109772358636   | 4.0× |
| `ROI_RADIUS_MAX`    | 45.254833995939045    | 19.026297590440446  | 2.4× |
| `ROI_RADIUS_MEDIAN` | 25.495097567963924    | 5.0                 | 5.1× |

All three follow in closed form from corners on an AABB spanning 0..64 — `MAX` is `32√2`, `MEDIAN` is
`√650` — and recomputing from corners at 0..63 reproduces none of them, which fixes the geometry
rather than assuming it. No tool computes distance-to-bounding-box-corners as a shape descriptor, so
it is kept as a snapshot, the same disposition the `EDGE_*` and `PERIMETER` whole-slide cells carry.

## Known follow-up, agreed as separate

`current_test` is feature-wide across the registry — 458 of 926 rows in 19 of 25 families, with 20
scanners each enforcing it — where SPEC §3 reads row-level. Left as its own cleanup rather than
half-applied here; `test_name` on these rows is already exact and row-level.

Benchmarking the out-of-core path on large images surfaced three further defects, none of which this
PR touches and all of which predate it — they reproduce on a build of the base commit with this
branch's changes removed. They are recorded rather than fixed here: the out-of-core contour is
truncated above the image's tile size, which is the same divergence as the `impl-defect` cell above
but far larger on a large ROI; a strip-layout TIFF over 1024 px in either dimension faults during
prescan; and a tiled TIFF that omits the optional `SampleFormat` tag is accepted by the whole-image
path and rejected by the segmented one. The first is why the end-to-end path-equality test is not in
this PR.

## Verification

- gtest: **902 cases, 901 passed, 1 skipped** (the GPU Gabor case), Windows/MSVC.
- pytest `tests/python/` (whole directory): **100 passed, 1 skipped, 7 failed** — the 7 are the
  pre-existing `@pytest.mark.arrow` failures of an `ALLEXTRAS=OFF` build, unrelated to this change.
- `check_coverage.py --check`, `check_test_names.py --check`, `report_features.py --check`: all
  clean (932 registry rows), with the generated reports regenerated.
- `gen_morphology_radius_skimage.py`: 6 verified, 0 failed, 0 unproducible.
- AddressSanitizer + UndefinedBehaviorSanitizer on Linux/gcc (`-fsanitize=address,undefined`
  `-fno-sanitize-recover=undefined`): **897 passed, exit 0, 0 diagnostics**, plus the pytest suite
  under the same sanitizers (99 passed, the same 7 Arrow failures). The sanitizer run covers the
  transform's raster indexing and the out-of-core path's disk-backed pixel cloud, both exercised by
  the new tests.


tests/vetting/config_recipes.md
```diff
@@ -255,6 +255,40 @@ oracle for the Nyxus-original features); it is not built in this tree, and the g
   agree to 3.8e-15. Note `nnz(bwperim(...))` counts perimeter *pixels* (846 here) and MATLAB's
   `regionprops('Perimeter')` returns 952.848 - neither is this quantity.
 
+## morphology.radius_disks
```
vjaganat90, line 258: [Blocker] This new oracle/config cell is absent from `tests/vetting/matrix/morphology.md`, whose scope still explicitly excludes this producer. SPEC §5.1 requires the family matrix to enumerate each real point and its disposition. Please add an ROI-radius section covering the in-RAM oracle point and the out-of-core path/disposition rather than leaving the recipe visible only here.

tests/vetting/oracle_coverage.csv
```diff
@@ -131,9 +131,12 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
+2D,ROI_RADIUS_MAX,morphology,regression,,na,morphology.shape2d_native,rel=1e-3,test_2d_morphology_analytic.h;test_2d_morphology_regression.h;test_2d_morphology_skimage.h,,,,in-tree,"RoiRadiusFeature measures every ROI pixel's distance to the ROI's own contour and reports the mean, maximum and median of those distances, on both the in-RAM and out-of-core paths. This row is the shape2d drift guard and claims nothing: the skimage reference disagrees on shape2d for the contour-offset reason recorded on the ROI_RADIUS_MEAN row. The vetting claim for this feature is its morphology.radius_disks row.",TEST_NYXUS.TEST_2D_MORPHOLOGY_RADIUS_REGRESSION,bench_shape8_concave_holed
```
vjaganat90: [Blocker] SPEC §3 makes this one row per assertion, but `current_test` is feature-wide here: this regression row names the analytic and skimage files even though `test_name` identifies only the regression case; the vetted rows below have the inverse problem. Please make every row name only its own assertion file/case and update `scan_morphology_coverage.py` to validate row-level ownership.

tests/vetting/audit/morphology_2d_coverage.csv
```diff
@@ -96,8 +96,11 @@ Dim,Family,FeatureName,List_of_Oracles,Test_Names,Regression,Reg_Test_Name,Notes
 2D,morphology,GEODETIC_LENGTH,imea,test_2d_morphology_geodetic_length_thickness_imea,N,,
 2D,morphology,THICKNESS,imea,test_2d_morphology_geodetic_length_thickness_imea,N,,
 2D,morphology,ROI_RADIUS_MEAN,,,Y,test_2d_morphology_radius_regression,
-2D,morphology,ROI_RADIUS_MAX,,,Y,test_2d_morphology_radius_regression,
-2D,morphology,ROI_RADIUS_MEDIAN,,,Y,test_2d_morphology_radius_regression,
+2D,morphology,ROI_RADIUS_MAX,analytic;skimage,test_2d_morphology_roi_radius_disk_closed_form_analytic;test_2d_morphology_roi_radius_disks_skimage,Y,test_2d_morphology_radius_regression,
+2D,morphology,ROI_RADIUS_MEDIAN,skimage,test_2d_morphology_roi_radius_disks_skimage,Y,test_2d_morphology_radius_regression,
+2D,morphology,ROI_RADIUS_MAX,analytic;skimage,test_2d_morphology_roi_radius_disk_closed_form_analytic;test_2d_morphology_roi_radius_disks_skimage,Y,test_2d_morphology_radius_regression,
```
vjaganat90: This feature rollup now emits `ROI_RADIUS_MAX` three times and `ROI_RADIUS_MEDIAN` twice because it renders once per registry assertion. This CSV is a feature-to-tests summary, so collapse registry rows by feature before rendering and aggregate the oracle/regression test names into one feature row.

tests/vetting/audit/morphology_2d_radius_skimage_vetting_report.md
```diff
@@ -0,0 +1,160 @@
...
+| ROI_RADIUS_MEDIAN | `vetted`/`skimage` on `morphology.radius_disks`; `regression` drift guard on `morphology.shape2d_native` |
+
+## Outside this round
+
+`docs/source/Math/f_morphology.rst` defines these three features as statistics of the distance from
+the **centroid** to each **edge pixel**, and `docs/source/featurelist.rst` repeats it as "centroid to
+edge distance". The implementation measures a different quantity — for each ROI pixel, the distance
+to the nearest contour pixel. On the disks above the documented quantity gives MEAN 9.57 / 19.54 /
+39.53 where the implementation gives 2.75 / 6.05 / 12.69. Which of the two the feature is meant to be
+is a question for the maintainers, not something this round settles.
```
vjaganat90, line 196: [Blocker] This cannot remain an unanswered maintainer question while the PR changes all three public values and promotes two as vetted. The public docs define centroid-to-edge statistics; this implementation and its oracles measure each ROI pixel to its nearest contour pixel. Please resolve the intended contract, then either implement the documented quantity or update both public docs and scope the oracle to the chosen definition.

tests/vetting/benchmarks.md
```diff
@@ -320,6 +320,31 @@ have to add.
...
+| | |
+|---|---|
+| Data | built by `calculate_disk_radius_values(R)` in `tests/test_2d_morphology_common.h`: mask `(x-c)² + (y-c)² <= R²`, `c = R + 3`, on a frame padded by 3 on every side |
+| ROI | one per disk — **317**, **1257** and **5025** pixels; **56**, **112** and **224** of them boundary pixels under the 4-neighbour inner boundary |
+| Shape | filled digital disks, intensity 1 everywhere (the radius features read the mask alone) |
+| Why it exists | the `ROI_RADIUS_*` values are distances, and one disk cannot tell a distance from its square |
+
+Recipes: `morphology.radius_disks`.
+
+Tests reaching it today: `test_2d_morphology_skimage.h`
```
vjaganat90, line 334: Please include `test_2d_morphology_analytic.h` / `TEST_2D_MORPHOLOGY_ROI_RADIUS_DISK_CLOSED_FORM_ANALYTIC` here as well; that test calls the same disk helper and claims `bench_radius_disks` in the registry.

vjaganat90 (changes requested):
The production correction is coherent: taking the exact square root on both paths and keeping the median in floating point fixes two real defects, and the three-radius fixture is a useful discriminator. I am requesting changes for the unresolved public contract and the SPEC bookkeeping: settle what ROI_RADIUS_* means before calling it vetted, add the new cell to the morphology matrix, and restore assertion-level registry ownership plus a one-row-per-feature coverage rollup.

darkclad:
Comments 1, 3, 4 and 5 are addressed in the commit just pushed — 4 resolved on evidence rather than as a decision: `calculate_inscribing_circumscribing_circle` (`circle.cpp:220`) already computes the documented centroid-to-edge quantity and publishes twice its min and max as `DIAMETER_INSCRIBING_CIRCLE` / `DIAMETER_CIRCUMSCRIBING_CIRCLE` (12.331707339908755 vs the pinned 12.3317073399088 on shape2d), so the docs were describing a different feature and are now corrected to the inradius statistics the code measures, with no feature value moving; writing the matrix section you asked for in 1 also turned up an out-of-core defect — both contour builders return 112 pixels on `bench_disk64_diagonal_boundary` and not the same 112, moving `ROI_RADIUS_MEAN` +17.8% and `MEDIAN` +34% while `MAX` stays bit-identical — now `flag=impl-defect` with a guard that pins both sides, the inequality and `MAX`'s equality.

On 2 I have replied on that thread rather than changing it: `scan_morphology_coverage.py` currently *enforces* the feature-wide `current_test` (it rejected the row-level form while these rows were being written), and the column is registry-wide — 458 of 926 rows across 19 of 25 families, with 20 scanners each carrying its own copy of the check — so I would rather do it as its own change than leave morphology meaning something different from the other eighteen; say the word and I will do the morphology half here.

tests/vetting/config_recipes.md
darkclad: **Fixed.** `matrix/morphology.md` now carries a `RoiRadiusFeature` section with four cells — in-RAM oracle on disks, in-RAM production-only on shape2d, out-of-core, and whole-slide — taking the file from two producers to three, and the scope paragraph no longer excludes this producer.

Writing it turned up a defect: both contour builders return **112** pixels on `bench_disk64_diagonal_boundary` and not the same 112, moving `ROI_RADIUS_MEAN` +17.8% (6.087109772358636 → 7.169174091182726) and `MEDIAN` +34% while `MAX` stays bit-identical, so that cell is `flag=impl-defect` on `morphology.disk64_forced_ooc` with three registry rows and a guard pinning both sides, the inequality and `MAX`'s equality.

tests/vetting/oracle_coverage.csv
darkclad: **Not fixed — I would like your call first.** `scan_morphology_coverage.py::disagreements()` currently *enforces* the feature-wide form (it rejected the row-level `current_test` while these rows were being written), and the column is registry-wide: **458 of 926 rows across 19 of 25 families**, with 20 scanners each carrying its own copy of the check; morphology alone is 85 of 126 rows.

Changing morphology alone makes `current_test` mean one thing here and another for the other eighteen families, so I would rather invert the check in all 20 scanners and re-cut all 458 rows as its own change — note `test_name` on these rows is already exact and row-level — but if you want the morphology half in this PR, say so and I will do it and file the rest.

tests/vetting/audit/morphology_2d_coverage.csv
darkclad: **Fixed** in `render()`: the rollup is now one row per feature, 126 → 99 rows, 0 duplicates, and the diff is 27 deletions with no other row changed.

Worth flagging that it was not specific to the new rows — morphology is the only family of twenty whose rollup duplicated, with `PERIMETER`, `MASS_DISPLACEMENT`, the five `EDGE_*` and `DIAMETER_EQUAL_PERIMETER` already emitted 4× each; `ROI_RADIUS_*` only made it visible.

tests/vetting/audit/morphology_2d_radius_skimage_vetting_report.md
darkclad: **Fixed**, and it resolved on evidence rather than as a decision: `calculate_inscribing_circumscribing_circle` (`circle.cpp:220`) already computes the documented centroid-to-edge quantity and publishes twice its min and max as `DIAMETER_INSCRIBING_CIRCLE` / `DIAMETER_CIRCUMSCRIBING_CIRCLE` — on shape2d `2*max||OC||` = 12.331707339908755 against the pinned 12.3317073399088, and `2*min||OC||` = 0.8284868934052972 against 0.828486893405308 — so the docs text was describing those features, and defining `ROI_RADIUS_MAX` that way would make it `DIAMETER_CIRCUMSCRIBING_CIRCLE/2`.

Taking your second option: `docs/source/Math/f_morphology.rst` and `docs/source/featurelist.rst` now define `ROI_RADIUS_*` as statistics of the ROI's inradius map — each ROI pixel's distance to the nearest contour pixel — and point at the two circle diameters for the centroid-to-edge extremes, with the oracle scoped to that definition and no feature value moving.

tests/vetting/benchmarks.md
darkclad: **Fixed.** `bench_radius_disks` now lists `test_2d_morphology_analytic.h` / `TEST_2D_MORPHOLOGY_ROI_RADIUS_DISK_CLOSED_FORM_ANALYTIC` beside the skimage test.

Added a line noting that both call the same `calculate_disk_radius_values()` helper and both claim this benchmark in the registry.

tests/vetting/matrix/morphology.md
```diff
@@ -78,6 +85,59 @@ are asserted **equal to the segmented column** by
...
+map: every ROI pixel's distance to the nearest contour pixel. The feature reads no setting of its
+own, so its only axis is the run mode — which decides which contour builder fills `K` — and, for the
+oracle cell, which fixture it is measured on.
+
+| run mode | fixture | verdict | oracle / reason |
+|---|---|---|---|
+| in-RAM (`calculate`) | disks R = 10, 20, 40 | **VALID** | `skimage` on `morphology.radius_disks`, and `analytic` for `MAX` on the same recipe |
+| in-RAM (`calculate`) | 8×8 `shape2d` | **VALID-BUT-PRODUCTION-ONLY** | no tool reproduces it *on this fixture*; recipe `morphology.shape2d_native`, drift guard only |
+| out-of-core (`osized_calculate`) | disk64 | **impl-defect** | the two paths disagree; recipe `morphology.disk64_forced_ooc` |
+| whole-slide (`SINGLEROI=true`) | — | not reached | `buildWholeSlideContour()` pushes four AABB corners, so the inradius map is a corner artefact; no row claimed and none asserted |
```
vjaganat90: [Blocker] This whole-slide path is reached: `workflow_2d_whole.cpp` calls `reduce_trivial_wholeslide()`, whose 2-D branch enters `reduce_trivial_2d_wholeslide()`; that function explicitly calls both `ContourFeature::extract()` (when ROI radius is requested) and `RoiRadiusFeature::extract()`. These features therefore consume the four AABB corners in production. Please measure and pin this cell, give it a SPEC disposition, and add one registry row per `ROI_RADIUS_*` assertion.

vjaganat90 (changes requested):
The follow-up resolves the disk oracle, public definition, benchmark, rollup, and out-of-core characterization. One substantive blocker remains at the whole-slide cell. Please also update the PR body's “Not addressed here” section; it now contradicts the documentation fixed on this head. I’m treating the repository-wide legacy  semantics as separate cleanup, not as a reason to hold this PR.

tests/vetting/matrix/morphology.md
darkclad: **Fixed.** You are right that it is production — `reduce_trivial_2d_wholeslide()` (`reduce_trivial_rois.cpp:415`) calls `ContourFeature::extract()` whenever `RoiRadiusFeature::required()` holds and then `RoiRadiusFeature::extract()` at line 524 — so the "not reached" verdict was wrong; the cell is now **VALID-BUT-PRODUCTION-ONLY** on `morphology.disk64_wholeslide`, with three registry rows carrying `flag=production-only-cell` and a matrix subsection giving the measurement.

Measured on the 64×64 frame: `ROI_RADIUS_MEAN` 24.490797504013532, `MAX` 45.254833995939045, `MEDIAN` 25.495097567963924 over 4096 pixels, against 6.087109772358636 / 19.026297590440446 / 5.0 segmented — the four AABB corners turn the inradius map into a distance-to-nearest-corner map, and recomputing from corners at 0..64 reproduces all three exactly while corners at 0..63 do not, which fixes the geometry rather than assuming it (`MAX` is `32*sqrt(2)`, `MEDIAN` is `sqrt(650)`); pinned with those closed forms beside the literals in `test_2d_morphology_whole_slide_roi_radius_is_corner_distance_regression`.

tests/vetting/oracle_coverage.csv
vjaganat90: It should be its own change. Doesn't make sense to only isolate morphology

vjaganat90 (approved):
LGTM

PR #472, opened 2026-09-08
fix(ngldm): compute the 3D NGLDM over the ROI, over 26 neighbours, at the IBSI dependence count

darkclad:
# 3D NGLDM: over the ROI, over 26 neighbours, at the IBSI dependence count

**18 of the 19 `3NGLDM_*` features change value** -- every one except `3NGLDM_DCP`, which Nyxus
hard-codes to 1 and which reads 1.0 before and after. The family now reproduces **MIRP 2.6.0 to a worst relative
difference of 8.7e-16** across the 17 features MIRP computes, whenever both tools are given the same
grey levels.

Full record, with both comparison tables and the reproduction:
`tests/vetting/audit/ngldm_3d_mirp_vetting_report.md`.

## What was wrong

**The matrix was built over the bounding box, not the ROI.** `calc_ngld_matrix` iterated the box and
explicitly declined to skip background:

```cpp
// Do not skip off-ROI pixels
//	if (cpi == 0)
//		continue;
```

On the segmented phantom that is **511,360** voxels against the ROI's **274,432** — the box interior,
its one-voxel shell dropped by the `1 .. dim-1` loop bounds. Background voxels all bin to one grey
level and each sees ~24 identical neighbours, so they piled into a single row at maximum dependence.

**The neighbourhood had 24 members, not 26.** `(0,0,+1)` and `(0,0,-1)` — the voxels directly above
and below the centre — were missing from `shifts`, and `maxNr = nsh + 1` carried a comment describing
the 2D count of 8.

**`calc_features` differed from the MIRP-vetted 2D twin in four more ways**, each changing values:

| | 3D was | 2D twin, and now 3D |
|---|---|---|
| emphasis sums | `j` from 1, dropping column 0 while `Ns` counted it | `j` from 0 |
| dependence count | `j` in `LDE`/`HDE`/`LDLGLE`, `j+1` in the rest, in one loop body | `j+1` throughout — column 0 is IBSI dependence 1, the centre voxel |
| `GLNU` / `DCNU` | re-derived inline over `j` in `[1,Nr)` | the `Sg`/`Sr` marginals `calc_rowwise_and_columnwise_totals()` already fills |
| `GLV` grey level | the row index `i+1` | `U[i]`, the binned intensity |

The last of those was the single furthest feature from MIRP once the matrix was right, at 2.67x.

## How it was checked

`gen_ngldm3d_mirp.py` runs MIRP three times on the same phantom and verifies **every column of all
three published tables**, not just the half it produced: the MIRP column against a fresh run, the Nyxus
column against the goldens pinned in the C++ headers, and the derived column recomputed from the two
— the fbn64 ratio as a value, the samelevels residual as an upper bound. It then checks the 16 oracle
goldens of both oracle tables against their own runs, and fails on any feature MIRP can vet that a
header does not pin. Last run: **83 verified, 0 failed**. Negative control: perturbing `3NGLDM_GLNU`'s Nyxus column or
`3NGLDM_HDE`'s ratio in the report makes the run fail, naming the offending column.

- **`ngldm3d.mirp_fbn64`** — MIRP discretises the ROI itself, `fixed_bin_number` n=64. Not
  config-matched; this is the measurement of the discretisation gap.
- **`ngldm3d.mirp_samelevels`** — MIRP is handed the grey levels Nyxus bins to, with
  `base_discretisation_method="none"`, so both compute the NGLDM over identical levels.
- **`ngldm3d.mirp_ibsi_rawlevels`** — the `IBSI=true` point, where Nyxus does not bin at all and the
  raw intensity is the grey level. Worst residual 7.54e-15 over 2001 grey rows.

The second is where the 8.7e-16 comes from:

| feature | Nyxus | MIRP | rel |
|---|---|---|---|
| `3NGLDM_LDE` | 0.15365 | 0.15365 | 7.2e-16 |
| `3NGLDM_HDE` | 40.6394 | 40.6394 | 0 |
| `3NGLDM_LGLCE` | 0.000783908 | 0.000783908 | 0 |
| `3NGLDM_HGLCE` | 1873.25 | 1873.25 | 0 |
| `3NGLDM_GLNU` | 6480.48 | 6480.48 | 0 |
| `3NGLDM_DCNU` | 32085.4 | 32085.4 | 0 |
| `3NGLDM_GLV` | 153.096 | 153.096 | 1.9e-16 |
| `3NGLDM_DCV` | 14.6164 | 14.6164 | 8.5e-16 |
| `3NGLDM_DCENT` | 8.40569 | 8.40569 | 4.2e-16 |
| `3NGLDM_DCENE` | 0.00347501 | 0.00347501 | 1.2e-16 |

(Ten of seventeen shown; the report carries all of them and the fbn64 table beside it.)

**`Ns` is the cheap independent check.** `GLNU/GLNUN` and `DCNU/DCNUN` are each exactly `Ns` by
construction, and both now read **274,432** — the ROI's voxel count, so every ROI voxel contributes
one matrix entry and nothing outside the ROI does.

## Where the vetting lands

**16 of the 19 features are `vetted` against MIRP at BOTH of the family's config points** — 32 oracle
rows in `test_3d_ngldm_mirp.h`, `rel=1e-3` (SPEC §7's same-definition tier):

| recipe | Nyxus config | how MIRP is set | worst residual | scope |
|---|---|---|---|---|
| `ngldm3d.mirp_samelevels` | `GREYDEPTH=64`, `IBSI=false` | `none`, over the levels Nyxus bins to | **8.7e-16** | narrowed per SPEC §4 — covers the NGLDM, not the discretisation, because the levels are reproduced from Nyxus' binning in numpy |
| `ngldm3d.mirp_ibsi_rawlevels` | `IBSI=true` | `none`, over the same raw values | **7.54e-15** | not narrowed — `IBSI` is `disable_binning`, so the raw intensity is the grey level on both sides and there is no binning step to reproduce |

The tree goes 607 → 623 features vetted (80% → 82%); the `ngldm` family 17 → 33.

The `IBSI=true` point is the weaker discriminator of the two, and that is measured rather than
assumed: at raw resolution **83.46% of ROI voxels have no matching neighbour** and the maximum
dependence reached is 17 of a possible 26. It is asserted anyway — SPEC §5.1 maps a VALID cell to an
oracle assertion, and it exercises the 26-neighbourhood and the ROI masking over 2001 grey rows where
`samelevels` uses 44.

**The regression table is three features, not nineteen.** `3NGLDM_DCP` (hard-coded to 1 on the Nyxus
side and 1 on MIRP's at both config points — an agreement that cannot fail), `3NGLDM_GLM` and
`3NGLDM_DCM` (MIRP's NGLDM emits no `gl_mean` / `dc_mean` column). The other sixteen keep no
snapshot: it would pin the same Nyxus run to a second literal and add no path or config coverage,
which is the shape morphology's five PCA axis features already have.

## `ngldm3d.mirp_fbn64` is a non-equivalent mapping, not a config cell

At MIRP's own `fixed_bin_number` the two tools do not share a grey-level ladder:

| | levels the ROI reaches |
|---|---|
| MIRP, `fixed_bin_number` n=64 | 1-64 |
| Nyxus, `GREYDEPTH=64` | 21-64, 44 distinct |

Nyxus bins with `to_grayscale(i, 0, ROI max, 64)` over a volume the loader has already shifted by its
minimum (−1024 on this CT phantom). Two independent causes: the lower bin edge is 0 rather than the
ROI minimum, and the volume-wide shift. Neither is an NGLDM question, so the recipe carries no rows
and exists as the measurement of that gap — which the residual ratios confirm, the
grey-level-weighted features being the ones far from 1 (`LGLCE` 0.024x, `HDLGLE` 0.0065x) and the
purely dependence-side ones near it (`DCNU` 0.79x).

`matrix/ngldm3d.md` records this as a mapping of the one `GREYDEPTH=64, IBSI=false` config point
rather than as a cell of its own. Both actual config points are now VALID and asserted; what the two
of them still do not cover is Nyxus' `GREYDEPTH` binning itself, which the matrix says explicitly.

## The matrix is built from the ROI's voxels

`prepare_NGLDM_matrix_kit` walks the ROI's voxel cloud once, setting the mask and collecting the binned
grey levels in the same loop, and `calc_ngld_matrix` takes its centres straight from that cloud rather
than scanning the bounding box and skipping background. The centre loop's cost follows the ROI's voxel
count, not its bounding box's volume, which is what matters for thin, diagonal and irregular ROIs.
Neighbours are still looked up in the mask. The NGLDM is a table of counts, so the order the centres
are visited in does not change it: all 19 `3NGLDM_*` values at both config points are bit-identical
to the bounding-box scan.

## The ROI mask costs a byte per bounding-box voxel

`prepare_NGLDM_matrix_kit` allocates a `SimpleCube<unsigned char>` over the bounding box — 25% on top
of the `PixIntens` cube already there, and the cheapest correct element type. Masking by "intensity
is zero" instead would allocate nothing, and is what the NGTDM family does, but it is not sound here:
`aux_image_cube` fills background with 0 and an ROI voxel whose intensity *is* 0 is then
indistinguishable from it. The MIRP-vetted 2D twin builds an explicit mask for the same reason, out
of a 4-byte `ImageMatrix`.

## 2D was checked and is untouched

`ngldm.cpp` already builds a mask `ImageMatrix` from `r.raw_pixels`, skips non-mask centres and
non-mask neighbours, and its 8-shift table is the correct 2D Chebyshev-1 neighbourhood. It is the
implementation 3D was brought into line with.

## Values change for users

18 of the 19 `3NGLDM_*` features move; `3NGLDM_DCP` is hard-coded to 1 and is the one that does not.
On the segmented phantom, for instance `3NGLDM_GLNU` 115443.18172715895 → 6480.4799440298511 and
`3NGLDM_LDE` 0.10159976999534079 → 0.15365019670462546. Anyone comparing against stored Nyxus output
sees eighteen of them move, and the `fix(` prefix puts that in the changelog.

## Gates

Rebased onto `upstream/main` `c79c8224`, after PR #470's `scanlib` refactor, PR #471's ROI-radius
work and the intensity-domain series. The gtest and pytest counts include the cases those brought.
The family's two prescans pass the environment's `fpimageOptions` to `scan_slide_props`, which the
intensity-domain series added; every 3D NGLDM pin and both MIRP golden tables hold unchanged on the
new base.

| gate | result |
|---|---|
| gtest (Windows/MSVC) | 948 cases, **947 passed, 1 skipped** (GPU Gabor) |
| `pytest tests/python/` (whole directory) | **116 passed, 1 skipped**, 7 pre-existing Arrow failures of an `ALLEXTRAS=OFF` build |
| `check_coverage.py` / `report_features.py` `--check` | clean; the registry moves 607 → 623 vetted (80% → 82%) |
| `run_scanners.py --check` | 20 of 20 families clean, 35 rows for this one |
| `gen_ngldm3d_mirp.py` | **83 verified**, 0 failed, 0 unproducible, 0 unquoted |
| ASan + UBSan (Linux/gcc) | 941 cases, 940 passed, 1 skipped, **0 diagnostics** (`NOEXTRAS=ON` compiles out the seven DICOM cases) |

## What this leaves open

- **The discretisation gap** is filed as its own item. It is the same `to_grayscale(i, 0, ...)` call
  in every texture family, so any family whose oracle recipe is `fixed_bin_number` on a CT-like
  fixture is measuring it rather than itself. Closing it makes `ngldm3d.mirp_fbn64` assertable, and
  its rows would sit beside the samelevels ones rather than replacing them.
- **The tightest guard on the sixteen vetted features is now the oracle's `rel=1e-3`**, since their
  `rel=1e-9` snapshots were removed as duplicates. Drift between those two bands would pass. If that
  is not wanted, tightening the oracle to `rel=1e-9` restores it in one character — the reason it
  sits at the SPEC §7 tier is that an oracle band has to hold on arm64, where this repo has been
  bitten before with no local detector.
- **A level-0 row is dropped from three emphasis numerators while `Ns` keeps it** (`ngldm.cpp:290`,
  `3d_ngldm.cpp:287`). Untouched here deliberately: the guard is identical in the MIRP-vetted 2D twin,
  1-basing the level would break this PR's own oracle agreement (`LGLCE` rel 0.0 → 6.1e-2), and no
  fixture has an ROI reaching level 0. It closes with the discretisation item.
- **`3NGLDM_GLM` and `3NGLDM_DCM`** have no counterpart in any tool — MIRP's NGLDM emits no
  `gl_mean` / `dc_mean` column — so they stay snapshots at every recipe.


tests/vetting/config_recipes.md
```diff
@@ -583,21 +583,29 @@ oracle for the Nyxus-original features); it is not built in this tree, and the g
...
+  levels 1-64; Nyxus bins it over `[0, ROI max]` of a volume shifted to a zero minimum and lands on
+  levels 21-64. The comparison therefore measures the level ladder. `3NGLDM_GLM` and `3NGLDM_DCM`
+  have no MIRP counterpart at all.
+
+## ngldm3d.mirp_samelevels
+- The same phantom and tool with `base_discretisation_method="none"`, MIRP handed the grey levels
+  Nyxus' `GREYDEPTH=64` binning produces (`by_slice=False`, distance 1, alpha 0 as above). Both sides
+  then compute the NGLDM over identical levels. Generator: `oracles/gen_ngldm3d_mirp.py`, which
+  reproduces the Nyxus binning in `nyxus_grey_levels()`.
+- **Config-matched, agreeing to 8.7e-16 over the seventeen comparable features, not asserted
```
vjaganat90: [Blocker] This is a valid scoped MIRP oracle, not a reason to leave the family regression-only. SPEC §4 explicitly permits an oracle to cover one stage when the notes narrow that scope. Move the 16 meaningful MIRP-comparable assertions to `test_3d_ngldm_mirp.h` at the same-definition `rel=1e-3` tier and mark them `vetted` on this recipe, stating “NGLDM matrix/formulas given these grey levels; discretisation not vetted.” Keep `3NGLDM_DCP`, `3NGLDM_GLM`, and `3NGLDM_DCM` regression-only; DCP is non-discriminating and MIRP lacks the other two.

tests/vetting/oracles/gen_ngldm3d_mirp.py
```diff
@@ -104,48 +112,108 @@ def read_nifti(path):
...
 
-def parse_report(txt):
-    """-> {feature: MIRP value} from the report's comparison table.
+def parse_report(txt, heading):
+    """-> {feature: MIRP value} from the comparison table under `heading`.
 
-    Rows look like | `3NGLDM_GLNU` | 115443 | 4350.27 | **26.5x** | -- Nyxus, MIRP, ratio. Only
-    the MIRP column is this generator's to verify; the Nyxus column is the program's own output and
-    the ratio is derived from the two.
+    Rows look like | `3NGLDM_GLNU` | 5636.02 | 4350.27 | 1.30x | -- Nyxus, MIRP, ratio. Only the
```
vjaganat90: [Blocker] The script says it verifies both comparison tables, but this parser discards the Nyxus and derived columns. Every Nyxus value and every claimed ratio/relative error could be changed while the generator still prints success. Parse the checked-in C++ oracle pins, verify both report columns against their sources, and recompute/check the final column in both directions; this should replace the report-only check once the MIRP header is added.

tests/vetting/matrix/ngldm3d.md
```diff
@@ -16,38 +16,49 @@ small.
...
+  voxels of Chebyshev distance 1;
 - the **coarseness parameter** *alpha* is fixed at 0 (an exact grey-level match is required).
 
+Both therefore match MIRP's `d1_a0.0` by construction.
+
 ## Config points
 
 | GREYDEPTH | IBSI | verdict | recipe / oracle |
 |---|---|---|---|
 | 64 | false | **VALID-prod-only** → regression | `ngldm3d.regression_ut_phantom`; all 19 features pinned in `test_3d_ngldm_regression.h` on `bench_ut57_3d` at `rel=1e-9` |
```
vjaganat90: [Blocker] Rows 29–31 are not three config points: they repeat the same Nyxus `GREYDEPTH=64, IBSI=false` point and classify candidate MIRP mappings as cells. SPEC §5 matrices enumerate each Nyxus config point once with `VALID`, `VALID-BUT-PRODUCTION-ONLY`, or `INVALID`. Collapse this to the actual point, record fbn64 as a non-equivalent mapping and samelevels as its scoped valid MIRP mapping, and give the real `IBSI=true` point a permitted measured disposition rather than “not exercised.”

tests/vetting/audit/ngldm_3d_mirp_vetting_report.md
```diff
@@ -1,128 +1,122 @@
...
+
+**What this agreement covers.** The NGLD matrix — which voxels are centres, which neighbours count,
+how dependence maps to a matrix column — and all seventeen feature formulas over that matrix. It does
+not cover the discretisation, because the levels are an input to the comparison rather than a result
+of it: the generator reproduces Nyxus' binning in Python to feed MIRP. That reproduction is not taken
+on trust. Its agreement to 8.7e-16 across seventeen features with different sensitivities is what
+confirms it, and the generator additionally checks the measured level span (21-64, 44 distinct)
+against this report.
+
+**What it discriminates.** Every one of these seventeen moves by orders of magnitude under an
```
vjaganat90: This overstates the negative control. `3NGLDM_DCP` is hard-coded to 1 and its old and new pins are both 1.0; lines 113–119 correctly admit it cannot discriminate these defects. Please state which assertions catch which failure rather than saying all 17 move by orders of magnitude.

vjaganat90 (changes requested):
The production correction itself is well supported: ROI masking, the full 26-neighbourhood, one-based dependence counts, the existing marginals, and actual LUT values all line up with the MIRP result. I am requesting changes because the evidence is being left as an offline claim instead of a SPEC oracle assertion, the matrix models oracle runs as duplicate config cells, and the generator does not verify the Nyxus half of its comparison. Please also correct the PR body: 18 of 19 values move, not all 19; `3NGLDM_DCP` is 1.0 before and after.

darkclad:
All five points are addressed in the follow-up commit: 16 features now carry `vetted` rows against MIRP at `ngldm3d.mirp_samelevels` (`test_3d_ngldm_mirp.h`, `rel=1e-3`, scope stated on the assertion, the registry row, the recipe and the matrix), the generator verifies all three columns of both published tables against their own sources (34 → 50 checks, with a negative control), the matrix enumerates one row per Nyxus config point with a measured disposition for `IBSI=true` (Ng=2001 raw levels, 83.46% of ROI voxels with no matching neighbour), the negative-control paragraph is now a per-failure-mode table noting `3NGLDM_DCP` catches none of them, and the 18-of-19 count is corrected everywhere. Gates on the round: gtest 904/903, ASan+UBSan 900/899 with 0 diagnostics, `pytest` 98 passed, generator 50/0 — and two calls are yours: I kept your `rel=1e-3` rather than tightening to the measured 8.7e-16 (arm64 float divergence has cost us three families with no local detector), and I filed the `IBSI=true` assertions as owed rather than adding them here.

tests/vetting/config_recipes.md
darkclad: Fixed. `tests/test_3d_ngldm_mirp.h` now asserts 16 features against MIRP at `ngldm3d.mirp_samelevels`, `rel=1e-3`, with the rows `status=vetted, oracle=mirp, flag=scoped-oracle` and the scope sentence repeated on the assertion, the registry row, this recipe and the matrix; `3NGLDM_DCP`, `3NGLDM_GLM` and `3NGLDM_DCM` stay regression-only for the reasons you gave. Tree goes 605 → 621 features vetted (80% → 82%); one deviation to flag for your call — I kept your `rel=1e-3` rather than tightening to the measured 8.7e-16, because an oracle band has to hold on arm64 and these are sums over 274,432 voxels.

tests/vetting/oracles/gen_ngldm3d_mirp.py
darkclad: Fixed, and it was worse than you said: the derived column was also unchecked, because the samelevels residuals are ~1e-16 and the comparison floored at 1e-12, so that column passed anything. The generator now parses the goldens out of the C++ headers and checks all three columns against their own sources — MIRP against a fresh run, Nyxus against the pins, the derived column recomputed (the fbn64 ratio as a value at six significant figures, the samelevels residual as an upper bound) — plus the 16 oracle goldens; 34 checks became 50, and perturbing a Nyxus value or a ratio now fails the run naming that column.

tests/vetting/matrix/ngldm3d.md
darkclad: Fixed. The matrix now has one row per Nyxus config point with a SPEC §5.1 verdict, and the two MIRP mappings of the `IBSI=false` point moved into their own table marked equivalent (`samelevels`) and non-equivalent (`fbn64`). `IBSI=true` gets a measured disposition instead of "not exercised": `IBSI` reaches `to_grayscale` as `disable_binning`, so MIRP matches it by construction, but on `bench_ut57_3d` it reaches Ng=2001 raw levels with 83.46% of ROI voxels having no matching neighbour — VALID and weakly discriminating, recorded as owed an oracle round rather than claimed as one.

tests/vetting/audit/ngldm_3d_mirp_vetting_report.md
darkclad: Fixed — you are right, and the report contradicted itself two sections apart. The paragraph is now a table of which assertions catch which of the five failure modes, stating that `3NGLDM_DCP` catches none of them and is excluded from the sixteen; two of the five are caught by fewer than half the set, which is the actual argument for asserting all sixteen rather than a representative few.

tests/test_3d_ngldm_regression.h
```diff
@@ -2,126 +2,58 @@
...
+// 17 features MIRP computes) once both sides are given the same grey levels. They are not asserted
+// against MIRP here because the two discretisations differ: at `fixed_bin_number` n=64 MIRP spreads
+// this ROI over levels 1-64, while Nyxus bins it over [0, ROI max] of a volume whose minimum has
+// been shifted to 0, which places the ROI on levels 21-64. Measurements, both level sets, and the
+// reproduction: tests/vetting/audit/ngldm_3d_mirp_vetting_report.md.
 //
 // 3NGLDM_GLM and 3NGLDM_DCM have no counterpart in any tool -- MIRP's NGLDM emits no gl_mean /
-// dc_mean column -- so they cannot be vetted even once the implementation is corrected.
+// dc_mean column.
 static const ref_vals_map<double> ngldm_3d_regression_ref_vals{
```
vjaganat90, line 22: [Blocker] Sixteen entries here now duplicate `test_3d_ngldm_mirp.h` on the exact same Nyxus fixture and config; only the literal source and tolerance differ. That adds no path or config coverage. Please keep regression pins/tests/registry rows only for `3NGLDM_DCP`, `3NGLDM_GLM`, and `3NGLDM_DCM`, and let the MIRP file own the other sixteen.

tests/vetting/matrix/ngldm3d.md
```diff
@@ -9,45 +9,62 @@ prepare_NGLDM_matrix_kit (NGLDM, greyLevelsLUT, Ng, Nr, r, STNGS_NGREYS(s), STNG
...
-than new plumbing.
+| 64 | false | **VALID** | `mirp` at recipe `ngldm3d.mirp_samelevels` — 16 features `vetted` in `test_3d_ngldm_mirp.h` at `rel=1e-3`; `3NGLDM_DCP`, `3NGLDM_GLM` and `3NGLDM_DCM` are regression-only for the reasons below. The same point also carries the family's 19 drift guards on `bench_ut57_3d` at `rel=1e-9` (`ngldm3d.regression_ut_phantom`) |
+| any other | false | **VALID** — same cell | `GREYDEPTH` sets the ladder's height, not its kind. Every depth reaches MIRP the same way `ngldm3d.mirp_samelevels` does, so it is one cell exercised at the depth the tests pin (64), not a family of cells |
+| any | true | **VALID, not yet asserted** | binning is *off*, so the raw intensity is the grey level and MIRP at `base_discretisation_method="none"` matches by construction — the same mapping as `mirp_samelevels`, with the identity in place of the 64-level ladder. Measured on `bench_ut57_3d`: Ng = 2001 distinct raw levels over 274,432 ROI voxels, **83.46% of them with zero matching neighbours** and a maximum dependence of 17 of a possible 26. A valid but weakly discriminating point, owed an oracle round rather than claimed as one |
```
vjaganat90: [Blocker] `VALID, not yet asserted` is not a completed SPEC §5 disposition: `VALID` explicitly maps to an oracle assertion, and `IBSI=true` is a distinct reachable feature×config point. Since MIRP with discretisation disabled is already identified as matching, please add the scoped oracle assertions and registry rows for this cell, or use another permitted measured disposition if an actual incompatibility is demonstrated.

vjaganat90 (changes requested):
The scoped MIRP oracle and generator repair are strong. The two remaining coverage issues are inline. Please also update the PR body: it still says all 19 features move although `3NGLDM_DCP` remains 1.0, and its “Why the rows are still status=regression” section no longer describes the 16 vetted rows. Once those are corrected, this should be LGTM.

tests/test_3d_ngldm_regression.h
darkclad: Fixed. `test_3d_ngldm_regression.h` now holds three features -- `3NGLDM_DCP`, `3NGLDM_GLM` and `3NGLDM_DCM`, the only ones no tool can judge -- and the 16 duplicate regression rows are out of the registry; the MIRP file owns those features outright. You were right that the snapshot could not catch anything the oracle missed: both read the same Nyxus run through the same fixture helper, so only the literal and the band differed, and the same precedent already exists for morphology's five PCA axis features. One consequence worth naming: the family's tightest guard on those sixteen is now the oracle's `rel=1e-3` rather than a `rel=1e-9` snapshot, so drift between those bands would pass -- if you would rather have the tighter band, tightening the oracle to `rel=1e-9` is the one-character change that restores it.

tests/vetting/matrix/ngldm3d.md
darkclad: Fixed, and it turned out to be assertable rather than incompatible. `IBSI` reaches `to_grayscale` as `disable_binning`, so that point does not bin at all -- the raw loader-shifted intensity is the grey level, 2001 distinct here -- and MIRP at `base_discretisation_method="none"` over the same raw values is config-matched by construction; measured worst relative difference **7.54e-15** across the same sixteen features. So the cell is now **VALID** and asserted: recipe `ngldm3d.mirp_ibsi_rawlevels`, 16 more vetted rows, a second golden table in `test_3d_ngldm_mirp.h`, and the fixture takes the IBSI flag so both points come from one helper. Note its scope is *wider* than `mirp_samelevels`', not narrower -- there is no binning step for the generator to reproduce -- though it is the weaker discriminator of the two, which the matrix records with the measurement (83.46% of ROI voxels have no matching neighbour, maximum dependence 17 of 26).

vjaganat90:
LGTM

vjaganat90 (approved):

sameeul:
Please fix in this PR:

1. `src/nyx/features/3d_ngldm.cpp` (~line 213, `prepare_NGLDM_matrix_kit()`) — this does two separate full passes over `r.raw_pixels_3D` (one in gather_unique_intensities, one to build roi_mask). Could these be merged into a single pass? For large ROIs this doubles the linear-time setup cost unnecessarily.
2. `src/nyx/features/3d_ngldm.cpp` (~line 112, `calc_ngld_matrix()`) — this scans the entire bounding-box cube (W*H*D) and skips non-ROI voxels via roi_mask, rather than iterating `r.raw_pixels_3D` directly (coordinates are already known). For thin/diagonal/irregular ROIs where the bounding box is much larger than the ROI itself, this wastes most of the loop on background cells that just continue.

Suggested follow-up (separate PR):

3. `src/nyx/features/ngldm.cpp` (2D NGLDM) has the identical defects this PR fixes for 3D:
   - `calc_features()` loops for `(int j=1; j<Nr; ++j)` (~line 270), excluding the zero-dependency column (`j=0` → IBSI dependence-count 1) from f_LDE, f_HDE, f_LGLCE, f_HGLCE, f_GLNU/f_GLNUN, and f_DCNU/f_DCNUN, even though Ns is computed over the full `j=0..Nr-1` range — the same Ns/loop-range mismatch diagnosed here for 3D.
   - f_GLV (~line 318) uses `i_1base = i + 1` (row index) instead of the actual grey level U[i], same bug as the 3D version had.

   Since the 2D and 3D implementations share the same structure, applying this PR's fix there too would resolve it. Given it's a separate file/feature, happy to see it as a quick follow-up PR rather than scope creep here.

darkclad:
**Fixed** `prepare_NGLDM_matrix_kit()` now makes one pass over `r.raw_pixels_3D`, setting the ROI mask and collecting the binned grey levels in the same loop, and `gather_unique_intensities()` is gone. All 19 `3NGLDM_*` values at `GREYDEPTH=64` and at `IBSI=true` are bit-identical to before. Done in 1e6e855b.

darkclad:
**Fixed** `calc_ngld_matrix()` now takes its centres straight from the voxel cloud, so the loop costs the ROI's voxel count rather than the bounding box's volume, and neighbours are still tested against the mask; the matrix is a table of counts, so the values are bit-identical. The per-centre `std::find` over the LUT and the 26 neighbour `to_grayscale` calls now dominate, and caching a level index per voxel would remove both at the cost of a second bounding-box cube, so that is left out. Done in 1e6e855b.

darkclad:
**Answered** 2D `ngldm.cpp` does not have these defects: its emphasis and non-uniformity loops run `for (int j = 0; j < Nr; ++j)` (lines 282 and 323) and `GLV` uses `U[i]` (line 322), as they have since 2023, and this PR does not touch the file. The `j = 1` loops and `i_1base = i + 1` at roughly lines 277, 329 and 338 are the pre-fix 3D file on `main`, which is what this PR replaces; the MIRP-vetted 2D twin is the reference the 3D fix was aligned to, so no follow-up PR is needed.

PR #473, opened 2026-09-08
fix(3d gldzm): compute the GLDZM over the ROI, at 26-connectivity, at the IBSI distance

darkclad:
# 3D GLDZM: over the ROI, at 26-connectivity, at the IBSI distance

All 18 `3GLDZM_*` features change value, and **16 of them become `status=vetted`** against MIRP
2.6.0, to a worst **absolute residual of 5.7e-15**. The family had no vetted row before this.

Full record, with every comparison table, the negative controls and the reproduction:
`tests/vetting/audit/gldzm_3d_mirp_vetting_report.md`.

## What was wrong

Four defects, all in `prepare_GLDZM_matrix_kit` or in what it called. Two of them share a root cause
worth stating first: **the family's MATLAB-style binning sends intensity 0 to level 1**
(`bin_array_matlab`: `if (i == 0) return 1;`). Everything downstream that used "intensity is zero"
to mean "outside the ROI" was therefore testing something that, after binning, is true of no voxel at
all.

**1. Zones were grown over the bounding box, and the background joined them.** The zone search seeded
at every voxel of the box and skipped level 0 only under `IBSI=true`:

```cpp
// Skip 0-intensity pixels (usually out of mask pixels)
if (ibsi_grey_binning(greyInfo))
    if (inten == 0)
        continue;
```

In the default mode there is no zero left to skip, so the background filling the rest of the box was
indistinguishable from a genuine level-1 ROI voxel. On the segmented phantom the box is **551,040**
voxels against the ROI's **274,432**.

**2. `dist2border` measured the distance to the bounding box, not to the ROI.** Its rays stopped at
`I.zyx(z, y, x) == 0` *or at the box margin*, and since no voxel is 0 after binning, **no ray ever
stopped at the ROI**. The function reduced to `min(x+1, W-x, y+1, H-y)` — a property of the box's
geometry with the ROI nowhere in it. That is the whole of `3GLDZM_LDE` **314.01** against MIRP's
**11.23**, and of `3GLDZM_ZDV` 79.72 against 3.25.

**3. `dist2border` never scanned the z axis.** Left, right, up and down, all within one z-slice, and
the minimum of four; `z` was a parameter used only to index the slice.

**4. Zones were 6-connected**, against the 26 IBSI defines in 3D — the connectivity MIRP implements
and the one Nyxus' own 3D GLSZM already uses for the same notion of a zone.

## What replaced them

| | before | after |
|---|---|---|
| which voxels are the ROI's | inferred from the binned intensity | marked from `r.raw_pixels_3D`, in the same cube that goes on to hold the distance — as 2D NGLDM's `ngldm.cpp` does, and as 2D GLDZM still does not |
| distance to the ROI border | four in-plane rays over the cube the search was destructively marking | a city-block distance transform, breadth-first from the ROI's surface inwards, computed once |
| zone growth | a parent-stack walk over 6 axis neighbours | a 26-connected component walk over an explicit voxel stack |

**Six rays would not have been enough.** Fixing defect 3 by adding the two z rays leaves an
approximation: the shortest of six rays is an upper bound on the city-block transform, and putting
that metric back into the fixed code **fails 9 of the 16 oracle assertions**, worst `LDHGLE` 77.12
against 60.19. On the segmented phantom the two metrics give mean distances of 5.5418 and 5.5198.

## Why the vetted cell is a new phantom and not the default configuration

Because MIRP's `fixed_bin_number` and Nyxus' `GREYDEPTH` do not put a CT-like ROI on the same grey
levels, and that is a property of the binning rather than of the GLDZM:

| | levels the ROI reaches |
|---|---|
| MIRP, `fixed_bin_number` n=64 | 1-64 |
| Nyxus, `GREYDEPTH=64` | 22-64, 43 distinct |

`bench_compat_gldzm_3d` removes the question instead of arguing about it: **its voxel values are the
grey levels**. MIRP runs at `base_discretisation_method="none"` and Nyxus at `IBSI=true`, so neither
side discretises and the comparison is of the GLDZM alone. 13 of the 16 values come back
bit-identical to MIRP's doubles; the worst of the other three is 5.7e-15 absolute, on `3GLDZM_ZDE`,
the only one whose formula takes a logarithm.

## The three config points

The family's axis is the binning scheme, not two independent settings: `IBSI=true` overwrites the
grey depth with 0 and every branch keys off `greyInfo` alone. So there are three points, and each has
a measured disposition and its own assertions (`matrix/gldzm3d.md`):

| binning scheme | verdict | rows | backed by |
|---|---|---|---|
| **none** (`IBSI=true`, or `GREYDEPTH=0`) | VALID(mirp) | 16 vetted + 2 regression | the 16 features pinned against MIRP, `GLM`/`ZDM` as drift guards, plus a mechanics case holding the two spellings to one set of values |
| **MATLAB** (`GREYDEPTH > 0`) | VALID-BUT-PRODUCTION-ONLY | 18 regression | the 18 pins on `bench_ut57_3d`; no tool reproduces this scheme, below |
| **radiomics** (`GREYDEPTH < 0`) | VALID(mirp) at a bin count the scheme maps identically | 16 vetted + 2 regression | the 16 asserted against MIRP **at this cell's own settings**, `GREYDEPTH=-8` on the compatibility phantom, where `to_grayscale_radiomix` maps levels 1..8 back onto themselves so MIRP's goldens apply unchanged |

Every point carries coverage for all 18 features. `3GLDZM_GLM` and `3GLDZM_ZDM` can be *vetted*
nowhere — MIRP emits no counterpart column and IBSI defines neither — but they are Nyxus outputs at
every cell, so every cell carries a drift guard for them rather than leaving itself uncovered.

`gldzm3d.mirp_fbn64` is **a config-mismatched reference mapping, not an invalid Nyxus cell**: the
MATLAB point is a perfectly valid configuration, and what fails is the attempt to map MIRP onto it.
The residual has exactly the shape a discretisation gap has — the grey-level-weighted features are
the ones far from 1 (`LDLGLE` 0.13x, `LGLZE` 0.29x) and `GLNU`, which weights nothing by grey level,
is 0.94x. `matrix/gldzm3d.md` files it under reference mappings with that measurement, and the
drift-guard rows carry `flag=discretisation-mismatch`.

A third run, `gldzm3d.mirp_samelevels`, hands MIRP the levels Nyxus bins to on the **274,432-voxel**
segmented ROI and agrees to **1.0e-14** — which is what says the agreement on the small phantom is
not an artifact of its size.

## The phantom, and the controls that say it is doing work

A 12×12×12 cube of 2×2×2 bricks with a 6×6×6 corner cut out of it, inside a two-voxel background
margin. A brick's grey level is `1 + 4*(bz%2) + 2*(by%2) + (bx%2)`, except the brick at `(3,3,3)`,
which carries 1 instead of the 8 that rule gives it. Nothing about it is arbitrary — each choice
separates one defect, and each control was run by patching `3d_gldzm.cpp` and rebuilding:

| choice | separates | control: put the defect back |
|---|---|---|
| the background margin | the mask, not the binned intensity, says what the ROI is | on the segmented phantom, zones over the box give 127,933 against 40,769 |
| the corner cut | the ROI is non-convex, so near the cut the shortest way out is diagonal | six-ray distance: **9 of 16 fail** |
| the `(3,3,3)` implant | it touches eight level-1 bricks at a corner and nowhere else | 6-connected: **all 16 fail**, `ZP` 0.125 against 0.12037. 18-connected: **all 16 fail**, the same 0.125 |

**Its zone count is derivable without any tool**: 189 bricks less the seven the implant absorbs is
182 zones over 1512 voxels, which is exactly the pinned `3GLDZM_ZP`.

The phantom is built by the generator (`--write-phantom`), and every other run of the generator
rebuilds it in memory and fails if the checked-in NIfTI pair has drifted from the rule.

## Values change for users

All 18 move. On the segmented phantom, for instance `3GLDZM_LDHGLE` 734618.35720259824 →
14203.744217420099 and `3GLDZM_ZP` 0.46617376982276121 → 0.14855774836753732. Anyone comparing
against stored Nyxus output sees every one of them move, and the `fix(` prefix puts that in the
changelog.

## Also in here

- **`test_3d_gldzm_dump_regression()`** — 3D GLDZM was the only 3D family whose pin table was
  transcribed by hand, which is the condition that once let `3GLDZM_ZDM` sit at 222 with no test.
- **A shared `test_3d_gldzm_common.h`**, so the oracle file and the drift-guard file cannot drift
  apart in the mock 3D workflow.
- **The 16 newly vetted features reach `externally_vetted_3d_feature_names()`** in
  `test_3d_coverage_common.h`, so the SPEC 1 completeness guard stops counting them as un-oracled;
  its counts move 126 -> 142 and 87 -> 71, on top of the 16 NGLDM features #472 added to the same
  index.
- **`bench_gldzm_zerolevel_3d` and `test_3d_gldzm_mechanics.h`** — the family had no fixture that
  could reach a ROI voxel whose grey level is 0, which is what `IBSI=true` and a negative
  `GREYDEPTH` both produce. Those voxels used to be dropped; they are lifted onto a valid level and
  kept. 64 voxels, so every value the fixture produces is derived by hand rather than pinned. The
  fixture runs at both points that can produce a level 0 — `IBSI=true` and `GREYDEPTH=-5` — because
  the two gather the ROI's levels differently (a 1..max ladder against a set) and each lifts them on
  its own code path.
- **The ROI's mark, zero-level check and level collection share one pass over the voxel cloud**,
  with a second pass only when a lift is needed, and the distance cube is not refilled after
  `allocate()` has zeroed it.
- `matrix/gldzm3d.md` and `audit/gldzm_3d_golden_regen.md`, neither of which existed.

## Gates

Run on the branch as rebased onto `main` at `685da533`.

| gate | result |
|---|---|
| gtest (Windows/MSVC) | **989 cases, 988 passed, 1 skipped** (GPU Gabor) — 59 of them this family's |
| `pytest tests/python/` (whole directory) | **116 passed, 1 skipped**, 7 pre-existing Arrow failures of an `ALLEXTRAS=OFF` build |
| `check_coverage.py --check` | clean; `coverage_report.md` regenerated — 623 -> **639** features vetted repo-wide |
| `scan_gldzm_coverage.py --check` | clean, 18 rows (the 2D family; 3D GLDZM has no scanner in `main` yet) |
| `audit/run_scanners.py` | 20 of 20 families clean |
| `report_features.py --check` | clean, 984 rows |
| `gen_gldzm3d_mirp.py` | **64 verified**, 0 failed, 0 unproducible, 0 unquoted — all three columns of every published table |
| ASan + UBSan (Linux/gcc) | **982 cases, 981 passed, 1 skipped, 0 sanitizer diagnostics**, all twelve source checksums identical before the build and after the run |

## What this leaves open

- **2D GLDZM carries defects 1 and 2, and is vetted anyway because its recipe hides them.**
  `gldzm.cpp` has the same `intensity == 0` test in `dist2border` and the same `IBSI`-guarded
  zero-skip. `gldzm.ibsi_phantom_2d` runs at `IBSI=true`, where the binning is off, background stays
  0 and both readings are correct — so the vetted cell is sound. At a positive `GREYDEPTH` the 2D
  family is wrong the same two ways, and no cell of `matrix/gldzm.md` is vetted there. Fixing it
  moves 2D goldens this PR does not own.
- **The admission gate does not model per-feature scratch**, for this family or any other:
  `get_ram_footprint_estimate_3D` budgets a ROI's cubes, and the zone list and the two index buffers
  sit on top of that. The report carries the full accounting. Making the estimate model scratch is a
  repo-wide change — every family declaring a worst case, and the trivial/out-of-core decision moving
  for all of them — so it is filed rather than taken here.
- **The discretisation gap** is filed as its own item. The same `floor(n * i / ROI max) + 1` over a
  zero lower edge is in every Nyxus texture family, so any family whose oracle recipe is
  `fixed_bin_number` on a CT-like fixture is measuring it rather than itself.
- **`3GLDZM_GLM` and `3GLDZM_ZDM`** have no counterpart in any tool — MIRP's GLDZM emits no
  `dzm_gl_mean` / `dzm_zd_mean` column and IBSI defines neither — so they stay drift guards at every
  recipe.
- **The radiomics scheme at a bin count that actually re-bins.** The `-8` identity path *is*
  exercised and vetted; what remains uncovered is a bin count that changes the levels, which puts
  the cell back into the same lower-bin-edge question the MATLAB point has. It needs a fixture
  that does not exist yet, not new plumbing. This is the only thing about the family still
  unexercised.


src/nyx/features/3d_gldzm.cpp
```diff
@@ -54,6 +57,75 @@ void D3_GLDZM_feature::clear_buffers()
...
+		d = dist.depth();
+
+	for (auto& v : dist)
+		if (v)
+			v = DIST_UNSETTLED;
+
+	// the 6 city-block moves
+	static const int mv[6][3] = { {-1,0,0}, {+1,0,0}, {0,-1,0}, {0,+1,0}, {0,0,-1}, {0,0,+1} };
+
+	std::vector<std::array<int, 3>> frontier, next_frontier;	// x,y,z of the voxels settled last
```
vjaganat90: [Blocker] These queues are not included in the claimed RAM bound. A thin ROI can put nearly every voxel into `frontier` (12 bytes per entry), while `Z` and the later zone stack also scale with ROI size; the binned cube plus distance cube are therefore not the full 8-byte scratch cost. Since `get_ram_footprint_estimate_3D()` is the admission gate, please either account for these buffers there or use storage whose peak is actually covered by the estimate, then correct the memory claim.

src/nyx/features/3d_gldzm.cpp
```diff
@@ -75,220 +147,113 @@ void D3_GLDZM_feature::prepare_GLDZM_matrix_kit (SimpleMatrix<unsigned int>& GLD
...
-					// Update zone size
-					zoneSize++;
-
-					// Make the new neighborhood pixel current parent
-					z = _z;
+				// Grey level 0 is the one level the LUT never carries -- the IBSI branch fills it
+				// with 1..max and the other erases 0 -- so a zone of it would have no row in the
+				// matrix. It reaches here only from the binning schemes that leave a zero at zero,
+				// which the MATLAB one does not.
+				if (inten == 0)
```
vjaganat90: [Blocker] The explicit mask still is not authoritative here. With IBSI/no binning or negative binning, a legitimate ROI voxel can retain intensity 0; it reaches this line with a nonzero `dist` mark and is nevertheless discarded, while the LUT construction above omits 0 too. The 1..8 compatibility phantom hides this. Please preserve zero-valued ROI voxels (mapping them to a valid grey level as required) and add a discriminating fixture containing both zero and nonzero values inside the ROI.

tests/vetting/matrix/gldzm3d.md
```diff
@@ -0,0 +1,67 @@
...
+- the **zone connectivity** is fixed at 26, which is what IBSI defines in 3D;
+- the **distance metric** is fixed at the city-block distance to the ROI border, likewise.
+
+## Config points
+
+| GREYDEPTH | IBSI | verdict | recipe / oracle |
+|---|---|---|---|
+| any | true | **VALID(mirp)** → vetted | `gldzm3d.mirp_compat_phantom`; 16 of the 18 features pinned in `test_3d_gldzm_mirp.h` on `bench_compat_gldzm_3d` at an absolute 1e-9, measured residual 5.7e-15 |
+| 64 | false | **VALID-prod-only** → regression | `gldzm3d.regression_ut_phantom`; all 18 features pinned in `test_3d_gldzm_regression.h` on `bench_ut57_3d` at `rel=1e-9` |
+| 64 | false | **INVALID** as a vetting cell | `gldzm3d.mirp_fbn64` — MIRP at `fixed_bin_number` n=64 is NOT config-matched to the row above, because the two tools put this ROI on different grey levels (1-64 against 22-64, 43 distinct). The residual measures the binning, not the GLDZM |
```
vjaganat90: [Blocker] This is not yet a SPEC §5 matrix. The next row repeats the same Nyxus `64,false` point and classifies an incompatible MIRP mapping as an INVALID config; the negative branch is left with the non-permitted “not exercised” verdict; and `0,false` is omitted even though it reaches the same no-binning implementation as `IBSI=true`. Enumerate each effective Nyxus point once, move `mirp_fbn64` to a reference-mapping/evidence section, and give the positive, negative, and no-binning points permitted measured dispositions with the required assertions.

tests/vetting/oracles/gen_gldzm3d_mirp.py
```diff
@@ -68,184 +92,317 @@
...
+    first and the search stops at the next heading of the same level.
+    """
+    m = re.search("^" + re.escape(heading) + r"\s*$", txt, re.M)
+    if not m:
+        raise RuntimeError("no %r section in %s" % (heading, os.path.basename(REPORT)))
+    rest = txt[m.end():]
+    nxt = re.search(r"^##\s", rest, re.M)
+    section = rest[: nxt.start()] if nxt else rest
+    rows = re.findall(
+        r"^\|\s*`(3GLDZM_[A-Z0-9_]+)`\s*\|\s*[-0-9.eE+]+\s*\|\s*([-0-9.eE+]+)\s*\|",
```
vjaganat90: [Blocker] This parser discards the Nyxus and derived columns, so the generator still succeeds if every published Nyxus value and ratio/relative error in these two tables is corrupted. Parse the regression pins, verify both value columns against their sources, and recompute/check the final column; the generator should make every numeric claim it says it re-verifies fail closed.

tests/vetting/oracle_coverage.csv
```diff
@@ -749,24 +749,40 @@ dim,feature,family,status,oracle,agreement,config_recipe,tolerance,current_test,
...
+3D,3GLDZM_SDE,gldzm,vetted,mirp,agreed,gldzm3d.mirp_compat_phantom,abs=1e-9,test_3d_gldzm_mirp.h,,,,audit,"vetted vs mirp 2.6.0 on bench_compat_gldzm_3d, recipe gldzm3d.mirp_compat_phantom: MIRP by_slice=False with base_discretisation_method=none, Nyxus IBSI=true, so both read the phantom's voxel values as grey levels 1..8 and neither discretises. SPEC 7 exact tier, an absolute 1e-9 band; 13 of the 16 values are bit-identical to MIRP's doubles and the worst residual over the other three is an absolute 5.7e-15, on 3GLDZM_ZDE, the only one whose formula takes a logarithm. The phantom separates each of the family's three definition choices and each is negative-controlled: 6- or 18-connected zones fail all 16 assertions, the shortest-of-six-axis-rays distance fails 9 of 16. Generator tests/vetting/oracles/gen_gldzm3d_mirp.py, report tests/vetting/audit/gldzm_3d_mirp_vetting_report.md",TEST_NYXUS.TEST_3D_GLDZM_SDE_MIRP,bench_compat_gldzm_3d
```
vjaganat90: These rows now identify runnable checked-in assertions, so `source=audit` is no longer correct under SPEC §3; the source is `in-tree`. Please update all 34 new vetted/regression rows. The audit report remains supporting evidence in `notes`, not the assertion source.

vjaganat90 (changes requested):
The core GLDZM correction and the scoped MIRP agreement are valuable. I am requesting changes for two production risks and two substantive SPEC/provenance gaps: zero-valued ROI voxels are still discarded, the new per-voxel work queues are outside the claimed RAM bound, the config matrix is incomplete/misclassified, and the generator verifies only one side of its published comparison tables. The registry source value is a straightforward cleanup. Please update the PR body after these are fixed, especially its memory and matrix claims.

src/nyx/features/3d_gldzm.cpp
darkclad: the queues are not in the claimed RAM bound

**Correct, and the claim was wrong rather than merely incomplete.** I counted the cubes and wrote
"peak is back to 8 bytes per bounding-box voxel, so the estimate still bounds the family". The
frontier, the zone stack and `Z` are not in that number, and a thin ROI does put nearly every voxel
in the frontier at once.

Both halves of your alternative are now done, and the answer to the second is not the one I would
have liked to give:

**Storage.** The frontier and the zone stack held `std::array<int,3>` — 12 bytes an entry. They now
hold packed linear indices, 4 bytes. A voxel enters exactly one of `frontier` / `next_frontier` over
the whole walk, so the two together never exceed the ROI's voxel count; the zone stack never exceeds
the largest zone.

**The claim.** Corrected in `audit/gldzm_3d_mirp_vetting_report.md`, which now carries the whole
accounting as two tables — the three cubes at 12 bytes per bounding-box voxel against the estimate's
12, and then the three ROI-scaled buffers that are *not* in the budget, each with its worst case.
It says plainly: **the estimate does not bound this family's peak, and it never did.** `Z` and the
zone-growing stack predate this branch and are in the 2D twin too, and the estimate models the cubes
a ROI needs rather than the per-feature scratch on top of them — for every texture family, not just
this one.

I did not take the "account for these buffers there" option, and I want to be explicit about why
rather than quietly not doing it: `get_ram_footprint_estimate_3D` is per-ROI and shared, so making it
model scratch means every family declaring a worst case and the trivial/out-of-core decision moving
repo-wide. That is a change I do not think belongs in a GLDZM PR. It is filed, and the report says
so in "What is left open". If you would rather it happen here, say so and I will do it.

src/nyx/features/3d_gldzm.cpp
darkclad: zero-valued ROI voxels are still discarded

**Correct, and the compatibility phantom did hide it exactly as you say.** Under `IBSI=true` (which
bins nothing) or a negative `GREYDEPTH` (radiomics, which maps 0 to 0), a genuine ROI voxel arrives
at that line with grey level 0, a nonzero `dist` mark, and no row in the LUT — and was skipped.

Fixed by preserving them, as you asked: `prepare_GLDZM_matrix_kit` now **lifts the ROI's levels by
one** whenever any ROI voxel bins to 0, before the LUT is built. A lift by one rather than a re-map
onto a dense ladder, so the levels become valid and nothing else about their spacing changes — the
grey-level-weighted features are weighted by that spacing. The MATLAB scheme already sends 0 to
level 1, so it is unaffected, and neither existing fixture moves by a digit.

**The discriminating fixture is `bench_gldzm_zerolevel_3d`**, new: an 8x8x8 volume whose 4x4x4 ROI
carries raw intensity 0 where `x+y+z` is even and 5 elsewhere. It is built by the generator like the
compatibility phantom and checked against its rule on every run. It is deliberately small enough
that **every number it produces is derivable by hand** — each parity class is one 26-connected zone,
both reach the surface so both sit at distance 1, and the levels after the lift are 1 and 6 — so
`test_3d_gldzm_mechanics.h` asserts a derivation rather than a snapshot, and the derivation is
written out above the table. MIRP cannot arbitrate this one: its GLDZM has the same problem with a
level 0 in its input.

**What the negative control turned up is worth your attention.** With the lift disabled, the run does
not merely drop those voxels — a zone of level 0 indexes one row past the matrix, and
`SimpleMatrix`'s bounds-checked accessor throws `std::out_of_range` out of a feature calculation.
That is the pre-existing radiomics-branch hazard the old `inten == 0` skip was standing in front of.
The lift removes the condition rather than the symptom, and the invariant is now written at the
matrix write in `calc_gldzm_matrix`.

tests/vetting/matrix/gldzm3d.md
darkclad: not yet a SPEC §5 matrix

**Correct on all four counts.** Rewritten.

The axis was wrong to begin with: `GREYDEPTH` and `IBSI` are not independent, because `IBSI=true`
overwrites the grey depth with 0 and every branch keys off `greyInfo` alone. So the axis is the
**binning scheme**, it has exactly three values, and the matrix has three rows — one each for none,
MATLAB and radiomics. `IBSI=true` and `GREYDEPTH=0, IBSI=false` are two spellings of one point, which
is the omission you flagged; they are now one row, and
`test_3d_gldzm_no_binning_spellings_agree_mechanics` holds them to one set of values so the claim is
asserted rather than read off the source.

`mirp_fbn64` is out of the config-point table and into a **reference-mappings** section with the
other two MIRP configurations, each saying which point it answers and what the answer was. It answers
none: it is not config-matched to the MATLAB point and cannot be made so.

Every point now has a permitted disposition, and "not exercised" is gone:

- **none** → VALID(mirp), 16 pinned features plus the two mechanics assertions
- **MATLAB** → VALID-BUT-PRODUCTION-ONLY, 18 regression pins, with the reason no tool reproduces it
- **radiomics** → VALID(mirp), and **measured rather than asserted**, per SPEC §5.2 step 4: at
  `GREYDEPTH=-8` on the compatibility phantom the scheme is the identity on the fixture's levels
  1..8, so the point returns the no-binning point's values and MIRP's goldens apply unchanged.
  `test_3d_gldzm_radiomics_binning_is_identity_here_mechanics` asserts that equality.

The matrix states the limit on that last one plainly rather than leaving it to a reader of the test
name: it does not generalise to a bin count that actually re-bins, and the fixture to measure that
does not exist yet.

tests/vetting/oracles/gen_gldzm3d_mirp.py
darkclad: the parser discards two of the three columns

**Correct.** The generator now parses all three and checks each against its own source:

- the MIRP column against this run of MIRP;
- the **Nyxus column** against the regression pins for the two segmented-phantom tables, and against
  the oracle pins within the band the assertions use for the vetted cell — parsing the pins, as you
  asked;
- the **third column** recomputed from the row's own two values and compared as the string the
  report writes.

Checking it caught something first: the published `rel` column was not derivable from the published
columns, because the MIRP column was truncated to twelve significant figures while `rel` had been
computed from full precision. The tables are re-rendered at full precision so every row is now
self-consistent, and the recomputation is exact.

**Negative-controlled in both directions.** Corrupting a Nyxus value in the samelevels table and a
ratio in the fbn64 table makes the run fail naming the column and the row:

```
FAIL 3GLDZM_ZP: third column '0.77', recomputed from this row's own values as '0.55'
FAIL 3GLDZM_GLNU: nyxus column 1349.4969192278445 against 1349.3969192278446 (rel 7.41e-05);
                  third column '0.0e+00', recomputed from this row's own values as '7.4e-05'
```

The run went from 48 verified to **64**.

tests/vetting/oracle_coverage.csv
darkclad: `source=audit` should be `in-tree`

**Correct.** All 34 rows now read `source=in-tree`. The audit report stays where it was, in `notes`,
as supporting evidence.

## What was re-run

| gate | result |
|---|---|
| gtest (Windows/MSVC) | **908 cases, 907 passed, 1 skipped** (GPU Gabor) — +3 mechanics cases |
| `pytest tests/python/` | 98 passed, 1 skipped, 7 pre-existing Arrow failures of an `ALLEXTRAS=OFF` build |
| `check_coverage.py --check`, `check_test_names.py --check` | clean |
| `audit/run_scanners.py` | 20 of 20 families clean |
| `report_features.py --check` | clean, 939 rows |
| `gen_gldzm3d_mirp.py` | **64 verified**, 0 failed, 0 unproducible, 0 unquoted |
| ASan + UBSan (Linux/gcc, `ubtest1`) | **904 cases, 903 passed, 1 skipped, 0 sanitizer diagnostics**, all twelve source checksums identical before the build and after the run |

The ASan gate script gained a check per finding, so a regression on any of them fails the gate rather
than a reviewer: the zero-level lift must be present, no coordinate-triple buffer may come back, and
all three mechanics cases must be registered.

tests/vetting/matrix/gldzm3d.md
```diff
@@ -0,0 +1,79 @@
...
+| 2 | **MATLAB** — `floor(n·i / ROI max) + 1`, clipped into `[1, n]` | `GREYDEPTH > 0, IBSI=false` | **VALID-BUT-PRODUCTION-ONLY** | recipe `gldzm3d.regression_ut_phantom`: all 18 features pinned in `test_3d_gldzm_regression.h` at `rel=1e-9` on `bench_ut57_3d`. No tool reproduces this scheme — see "Why point 2 has no oracle" |
+| 3 | **radiomics** — `(i − ROI min) / binW + 1`, `binW = (max − min) / n` | `GREYDEPTH < 0, IBSI=false` | **VALID(mirp)**, on a fixture the scheme maps identically | `test_3d_gldzm_radiomics_binning_is_identity_here_mechanics`: at `GREYDEPTH=-8` on `bench_compat_gldzm_3d` the scheme is the identity on the fixture's levels 1..8, so the point returns point 1's values and MIRP's goldens for point 1 apply unchanged. Measured, not argued — see the limit below |
```
vjaganat90, line 39: [Blocker] This is very close, but the radiomics point still has a config-level `VALID(mirp)` verdict without feature × config oracle coverage. The mechanics test proves only Nyxus(`GREYDEPTH=-8`) == Nyxus(no-binning); it never asserts the radiomics output directly against MIRP, and the registry has no radiomics rows. Per SPEC §5.2, please add direct `-8`-vs-MIRP assertions plus `vetted` rows for the 16 mapped features, with regression rows for `3GLDZM_GLM` and `3GLDZM_ZDM`. Those two features also still need config-specific regression coverage at the no-binning point; they are Nyxus outputs, not “outside” those cells.

vjaganat90:
Thanks for the thorough update—the production fix, zero-level fixture, generator verification, and registry-source corrections address the prior findings. One config-coverage blocker remains inline.

Please also align the PR body with the revised matrix:

- `mirp_fbn64` is a config-mismatched reference mapping, not an `INVALID` Nyxus cell.
- The radiomics `-8` identity path is exercised; only a configuration that actually re-bins remains unexercised.

The RAM-estimator limitation is now honestly scoped, and I do not consider it a blocker for this PR.

tests/vetting/oracle_coverage.csv
darkclad: **Fixed** All 34 `3GLDZM_` rows now read `source=in-tree`, because each names a runnable checked-in assertion. The vetting report remains in `notes` as supporting evidence only. Done in 2a7a88a2.

tests/vetting/matrix/gldzm3d.md
darkclad: **Fixed** The radiomics cell now has its own 16 `vetted` rows at `gldzm3d.mirp_compat_phantom_radiomics`, each from a direct assertion of Nyxus at `GREYDEPTH=-8` against MIRP, and the identity test stays only as the reason the goldens transfer. `3GLDZM_GLM` and `3GLDZM_ZDM` now have regression pins at both the no-binning and radiomics cells, so the registry grows from 34 to 54 rows, and the PR body and matrix now call `mirp_fbn64` a config-mismatched reference mapping. Done in 8433f905.

sameeul:
A few optional cleanups, none blocking:

1. **`src/nyx/features/3d_gldzm.cpp` (~line 164)** — `dist.fill(0)` right after `dist.allocate()` is dead work: `allocate()` calls `resize()` on a freshly-constructed `SimpleCube`, which already value-initializes every new `int` to 0. For a large ROI bounding box this is an extra full `O(W*H*D)` pass for no observable effect — safe to drop.

2. **`src/nyx/features/3d_gldzm.cpp` (~line 169, `prepare_GLDZM_matrix_kit`)** — this makes up to four separate full passes over `r.raw_pixels_3D` (marking `dist`, checking `roi_has_zero_level`, shifting `D`, then building `I`/`n_ibsi_levels`). For a large 3D ROI (millions of voxels) this quadruples memory traffic versus fusing into one or two passes that mark `dist`, track the zero-level flag, and collect the level set/max together.

3. **`src/nyx/features/3d_gldzm.cpp` (~line 226)** — the new 26-connected zone flood-fill duplicates the same connectivity-walk concept already hand-rolled, differently, in `D3_GLSZM_feature::gather_size_zones` (`src/nyx/features/3d_glszm.cpp`) — no shared helper exists for either. Not a bug, but the two nearly-identical 3D zone-finding algorithms (GLDZM vs GLSZM) now live as separate implementations with different code paths; a future connectivity fix to one (like the one this PR just made) won't automatically propagate to the other. Might be worth extracting a shared 26-connectivity flood-fill helper at some point.

darkclad:
**Diverted** Items 1 and 2 are done in 9027d200: the redundant `dist.fill(0)` is gone, and one pass over the voxel cloud now marks the ROI, detects a level 0 and collects the levels, with a second pass only when a lift is needed; no feature value moves. Breaking that equivalence showed nothing covered the radiomics set path's lift, so the zero-level fixture now also runs at `GREYDEPTH=-5` and fails if that lift is removed. Item 3, a shared 26-connected zone walk for 3D GLDZM and GLSZM, is deferred to its own change because it has to move GLSZM without moving its outputs.

vjaganat90 (approved):
LGTM

PR #474, opened 2026-09-16
test(vetting): scan every family and fold coverage CSVs into report_output.csv

darkclad:
Five family x dim combinations had an audit/<family>_<dim>_coverage.csv and no scanner behind it, so nothing re-derived them from the tree: 2D first-order, 3D first-order, 2D gabor, 2D GLCM and 2D GLRLM. Each now has a scan_*_coverage.py, so all 25 are scanned and KNOWN_UNSCANNED is empty. Every new scan was compared against the hand-written file it replaces: GLRLM, GLCM and 3D first-order match it feature for feature, and 2D first-order and gabor find cases the hand-written file missed.

Scanners no longer write a CSV of their own. scanlib.run only runs the acceptance checks, report_features.py already joined every scan into one table, and it now carries each family's note as scan_notes, so the 26 per-family audit/*_coverage.csv files are deleted. features.csv and features.md are renamed report_output.csv and report_output.md.

What the new scanners needed:
- 3D first-order: its 17 PyRadiomics TEST() bodies called the assert helper directly, so no test_name could resolve to a function. They now call test_3d_firstorder_<feature>_pyradiomics(), and the three regression calls sit on one line where the feature is on the assertion line. The family runs the identity checks.
- 2D first-order: one registry row per oracle, and the pyradiomics and matlab cases each loop a golden table keyed by feature name, so the scanner reads both tables and the family runs the identity checks. The MEDIAN_ABSOLUTE_DEVIATION regression call sits on one line for the same reason as the 3D ones.
- The matlab table is written { "3COV", ... }; the cpp table-key pattern now allows the space. No cell moves in any family scanned before.
- gabor: both oracle cases assert through a file-local helper; the scanner credits a case with what the helper it calls asserts, and reads each row's config off the case name.
- 2D GLCM: the pytest _AVE case loops goldens.items(); py_loop_tables reads a dict looped that way, with two new selftest controls.

Registry current_test was wrong in 45 rows. The base 2D GLRLM and 2D GLCM rows named the 3D test files; test_nyxus.py was named for features it only folds into a checksum or a positional mean; test_2d_glcm_pyradiomics.py for base GLCM rows when it asserts their _AVE twins. Every row now gets the verdict agree, so the report's allowlist is empty.


tests/vetting/audit/scan_imq_coverage.py
```diff
@@ -247,14 +247,13 @@ def disagreements(fam, cov):
...
 
 FAMILY = scanlib.Family(
-    dim="IMQ", family="imq", out="imq_coverage.csv",
+    dim="IMQ", family="imq",
     sources=SOURCES,
     oracle_suffix=ORACLE_SUFFIX,
     notes=NOTE,
     collect_override=collect,
     # the mechanics guards are neither oracle nor regression, and this family names them in a
     # column of their own rather than folding them into the notes
```
darkclad: This comment still says IMQ names its mechanics guards "in a column of their own", but this PR removed that column. The guards now appear only in the `scan_other_tests` column of `test_output.csv`, so the comment should say that.

